PR #5, opened 2018-04-17
Upgrade func dependencies and results

d-g-h:
- [x] Upgrade func dependencies
- [x] set Chrome screenshot
- [x] temp exclude failing tests


components/button/functional/button.functional-spec.js
```diff
@@ -56,14 +56,14 @@ describe('Button example-with-icons tests', () => {
...
   });
 
-  if (browser.browserName.toLowerCase() === 'chrome') {
+  if (browser.browserName === 'chrome') {
     it('Should not visual regress', async () => {
       expect(await browser.protractorImageComparison.checkScreen('buttonPage')).toEqual(0);
     });
   }
 
   // Disable IE11: Async timeout errors
```
clepore: Just curious, what does this comment mean and is it still valid (only because it doesn't look like it disables anything below)?

d-g-h:
Nice questions, this information should probably go in TESTING.md for the time being as well. I will add it there.

> How come we do so much browser exclusion logic?

Each browser has a different Selenium driver with different capabilities. We plan highlight this difference for manual testing. As browser capabilities get updated, we should revisit tests that don't work. As for the Chrome exclusions, we are only testing visual regression on Chrome. Chrome is the default local functional test browser, and will be responsible for aiding the creation of the baseline images for visual regression testing.

> Also rather than copy and paste the axeOptions into each file, can we extract that into an import?

They all are not the same for each test. I agree, rule commonalities should be extracted into a file at some point, or a method should be created to configure axe. This a bit complex as the light theme is not completely WCAG AA... and per component in various states (open/close) may not be WCAG 2AA as well. TLDR; Yes, probably should reduce the number of lines this configuration is taking up at some point. Have reviewed all of these rules with QA.

We can cover this in greater detail on a Webex, this is by no means a final solution

components/button/functional/button.functional-spec.js
d-g-h: The next line is ` if (browser.browserName !== 'ie') {` so this battery of  accessibility tests don't run on ie

clepore: Okay, I was looking more for a disabling options, versus an exclusion, but ok.

d-g-h: I can change the comment to `Exclude`

nickwynja:
>This a bit complex as the light theme is not completely WCAG AA... and per component in various states (open/close) may not be WCAG 2AA as well. TLDR; Yes, probably should reduce the number of lines this configuration is taking up at some point. Have reviewed all of these rules with QA.

Does the WCAG AA issues just throw a warning or an error? Is there a way to configure that? It might be nice to have all of the accessibility issues reported but not break the build, instead of supressing them.

d-g-h:
> Does the WCAG AA issues just throw a warning or an error? Is there a way to configure that? It might be nice to have all of the accessibility issues reported but not break the build, instead of supressing them.

Definitely should throw an error, and reports it as an "violation" as the page is not accessible for the rule in question. We will never pass some of these rules without drastic changes to light-theme.

Perhaps, we can test high-contrast with all the rules off for comparison?

Honestly, not really sure what to do here other than fix the errors, and make note of the exceptions.

Here are links to the repo for a deep dive though.
https://github.com/dequelabs/axe-core
https://github.com/dequelabs/axe-webdriverjs

tmcconechy:
The main things that dont pass on light theme are the Azure text on white. We would have to change that design.

If we do i think then its definitely an error but as of now i think we should ignore contrast errors on that theme because we dont support that in the design yet.

For high contrast it should pass so its valid to check there.

tmcconechy:
Is there any explanations of the coverage report? I'm not sure what the percents mean.
We had coverage of around 6% before on units which makes sense but these numbers dont add up to me.

nickwynja:
@tmcconechy We're still sorting that out. I don't yet love how this is reporting either but it's the only way we could get it to work properly.

What the `22.76%` is is that 20% of the code _which has tests_ is covered by one of the tests. So out of the ~115 components, 15 have tests and 20% of that 15/115 is covered.

I'd like to go back to reporting coverage across the whole code base but there were issues with it reporting based on the `dist` instead of the `src`. I think we'll be able to better isolate after the re-org when tests are in a different location.

clepore:
So that build [fails](https://travis-ci.org/hookandloop/project-hardcore/jobs/367769530#L674), but only after I pulled from `controls/4.6.0-rc`:

screen shot 2018-04-17 at 2 43 45 pm

tmcconechy:
The build error looks like an issue with the Date (server time zone). Its not really the point of the test. I'll just change the template a bit so it doesnt use a date there in PR on bit bucket you can merge.

clepore:
Sure lets do it

PR #9, opened 2018-04-20
Restore running of func tests, update docs

d-g-h:
Restore Protractor functionality
* Some tests are failing because mock/junk is throwing an error on GET

```sh
Error: Cannot find module '/Users/username/project-hardcore/src/js/getJunkDropdownData'
```


nickwynja:
The coverage reporting doesn't seem to be working here anymore. I can look into it on Monday.

PR #15, opened 2018-04-26
(demo)app Cleanup for File Reorganization

EdwardCoyle:
Related to SOHO-7803

Got all component paths and related documentation working:
- `/` goes to kitchen sink page
- `/index` goes to docs
- `/kitchen-sink` goes to kitchen sink page
- `/components` goes to generated docs page with list of components
- `/components/list` goes to Express app's automated component list
- `/components/:component` goes to generated docs page for that component
- `/components/:component/list` goes to Express app's automated example list for that component
- `/components/:component/:example` goes to a specific example or visual test

At this point, many of the other existing routes "work" as well, but we should consider generalizing what's been done to the component routes to work with patterns, layouts, (customer) examples.  There are also some routes for `/tests` that are no longer used, but have some specific Express layout rules that need to be migrated to `/components`.

Other work:
- Separated out various parts of the express app into more manageable bits
- Cleaned up some unused vars/dependencies
- Generated docs and existing Express pages are using the same scripts/CSS where possible.
- (demo)app now has a "pre-build" process that generates demo-specific CSS from source SASS files.
- Recent integration with Content Security Policy (CSP) has broken many inline scripts because the Express app isn't rendering the `nonce` attribute properly.  I fixed all the ones I could find in `components` but we still need this in a lot of places.  Consider this on-going.


nickwynja:
Reminder to all the you need to run `npm run documentation` to generate the docs for the components since that command no longer is part of `npm start`.

d-g-h:
On http://localhost:4000/api/states

I'm still seeing...

```sh
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at validateHeader (_http_outgoing.js:503:11)
    at ServerResponse.setHeader (_http_outgoing.js:510:3)
    at sendJSONFile (/Users/d/project-hardcore/app/data.js:883:9)
    at router.get (/Users/d/project-hardcore/app/data.js:889:5)
    at Layer.handle [as handle_request] (/Users/d/project-hardcore/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/d/project-hardcore/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/d/project-hardcore/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/d/project-hardcore/node_modules/express/lib/router/layer.js:95:5)
    at /Users/d/project-hardcore/node_modules/express/lib/router/index.js:281:22
    at param (/Users/d/project-hardcore/node_modules/express/lib/router/index.js:354:14)
```

nickwynja:
@d-g-h That's a separate issue we can address in a different PR. I have a lead on what's going on but need to make sure I'm not missing anything.

nickwynja:
@EdwardCoyle We removed it originally because it was running in CI and making build times longer. My first thought was to make a flag within the `grunt` command to _not_ run the docs generator and to use the flag when running on CI.

tmcconechy:
You can't do a flag in grunt but you could make another command. Fx

- grunt build
- grunt documentation
- grunt build&documentation

Something like that

tmcconechy:
A bit late to the party but i thought we agreed were using the file name convention of kebab case.

Fx 
 app/src/js/middleware/basepath-handler.js
not 
 app/src/js/middleware/basepathHandler.js

So this is a bit inconsistent.

clepore:
- [ ] Update casing structure to be kebabs    

(kebab:  --🌶--- 🍗-🍅--  ...nom nom)

PR #19, opened 2018-05-03
Move last of the newer files

clepore:
Moves the last of the untracked `./component` files to their `./app` hierarchy.


src/components/template.js
```diff
@@ -1,5 +1,5 @@
-import * as debug from '../../utils/debug'; // NOTE: fix this path when moving to a component folder
-import { utils } from '../../utils/utils'; // NOTE: fix this path when moving to a component folder
```
nickwynja: I think this comment about updating the path when you copy the template to a component is still helpful.

PR #21, opened 2018-05-07
Initial npm for ids-enterprise

clepore:


nickwynja:
Could we do something instead like`publish/@infor` and `publish/ids` or some sort of folder nesting inside of `publish` rather than two directories in the root?

clepore:
Yeah that's what I wanted to talk about and see what people think is best. We could just switch over entirely if we wanted to.

tmcconechy:
I think we can switch over entirely and just publish to npm (ids-enterprise). We don't have to / cant do both i think.

clepore:
You can do both, as long as you have a package.json for each "way"

nickwynja:
I don’t really know of a good way to migrate people from the local `@infor` npm except to just tell everyone to update their `package.json` to use the new one. Maybe we publish 4.6 on `@infor` as well as public for just this version and tell people this will be the last release on `@infor`… gives them a month basically migrate.

tmcconechy:
Yes, thats what i was also thinking. 4.6.0 is already publish on @infor now as part of the old process.
So its there. The next one 4.7.0 will be on npm public in a month. (Or on both?)

clepore:
We can still publish a new package at a new name at `4.6.0` if we wanted to have both exposed. Up to you guys.

tmcconechy:
I guess it doesn't hurt so your saying we have

4.6.0 - on public as ids-enterprise and on @infor as sohoxi ?
4.7.0 - on public as ids-enterprise and on @infor as ids-enterprise ?

I fine no matter what but the main thing to me will be 4.7.0 on public as ids-enterprise

nickwynja:
@clepore Yea, I'm good with that.

PR #23, opened 2018-05-07
Bugfix: Dropdown `HTMLElement.selectedOptions` usage

EdwardCoyle:

When opening the Dropdown component on Internet Explorer 11, Javascript errors were previously thrown due to the usage of `HTMLSelectElement.selectedOptions` for checking on the selected item.

**Related issue (required)**:  https://jira.infor.com/browse/SOHO-7901

Run the application server and open http://localhost:4000/components/dropdown/example-no-search-lsf in a browser, or run the Dropdown E2E tests.

**Test plan (required)**

Check that the first Dropdown on http://localhost:4000/components/dropdown/example-no-search-lsf correctly displays a result with a first character that matches an alphanumeric keypress (ex: pressing "R" while the dropdown has focus should change its selection to "R - Rocket Raccoon")

**Other Details:**

Closes SOHO-7901

### DIFF
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -1415,8 +1415,20 @@ Dropdown.prototype = {
     if (!this.settings.multiple && this.initialFilter) {
       setTimeout(() => {
         if (self.settings.noSearch) {
-          const selectedOptions = self.element[0].selectedOptions;
-          self.searchInput.val(selectedOptions.length > 0 ? selectedOptions[0].innerText : '');
+          let selectedOpt;
+          let selectedOptText = '';
+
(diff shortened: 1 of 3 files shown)


jake-harrington:
@EdwardCoyle Why do it in a different way for IE specifically? Does the IE code not work in other browsers as well? Seems like a manageable amount of code in either case so it seems strange to do it in two ways.

EdwardCoyle:
@jake-harrington doing it with `selectedOptions` is the better way because we can handle multiple selected options.  However, IE doesn't support that property, so we had to fall back to a `selectedIndex` based method.  Originally we were only doing it one way.

PR #25, opened 2018-05-08
Added fix for soho-7814

tmcconechy:

- removed a few onClicks as they fail csp
- added an example of a datagrid on a tab with scrollable body for support
- fixed soho-7814 by making label have line-height : normal. Thai language has a very tall character.

**Related issue (required)**:

https://jira.infor.com/browse/SOHO-7814

http://localhost:4000/components/input/test-thai-labels
- make sure the top of the tall thai word is visible


nickwynja:
As an admin, I'm going to merge this PR even though it's not passing all checks. It does [pass the build](https://travis-ci.org/infor-design/enterprise/builds/376455405) but is on the old system so doesn't show up appropriately.

PR #26, opened 2018-05-08
Updating NPM to publish newly as "ids-enterprise"

clepore:

So we can have an "open" version of `ids-enterprise` going forward

**ALSO** fixes a bug where icons don't show on the demo `app/`


clepore:
- [ ] I think we need to add a READ.me to the publish folder that will show on NPM.

publish/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ids-enterprise",
   "slug": "ids-enterprise",
-  "version": "4.6.0-alpha.2",
+  "version": "4.6.0",
```
tmcconechy: Should we/when should we update this to 4.7.0-dev or -rc or something?

clepore: I don't think we should until we are ready to release it.

clepore: Actually, nm. I think you're right.

PR #27, opened 2018-05-09
SOHO-7856 added spaces for the focus to be display properly

LarryAmisola:
Provide a margin only to first-child input elements like input field and checkbox preceded by a label with a class name audio. This will directly solved the focus cut-off issue.

**Related issue (required)**:
This will also solved the focus issue of the search field inside the modal component.
http://localhost:4000/components/modal/test-required-fields

Open this test page once fixed have been merge.
http://localhost:4000/components/modal/test-focus-box-shadow-cut

### DIFF
--- a/src/components/modal/_modal.scss
+++ b/src/components/modal/_modal.scss
@@ -70,9 +70,22 @@
         width: 400px;
       }
     }
+
+    &:first-child .audible ~ input:not(.spinbox):not(.colorpicker),
+    &:first-child .audible ~ .searchfield-wrapper {
+      margin-top: 4px;
+    }
+
+    &:first-child .checkbox-label {
+      margin-top: 2px;
+    }
+
+    &:last-child .checkbox-label {
+      margin-bottom: 1px;
+    }
   }
 
-  .field:last-child input:not(.spinbox):not(.colorpicker),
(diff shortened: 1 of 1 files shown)


tmcconechy:
If i open the example page and then enter the field and tab out. Then there is excessive space between the input and error icon/message.

Looks like we need to toggle the margin-bottom rule for errors as well.
screen shot 2018-05-09 at 8 58 53 am

Secondly there seems to be a double box shadow around the input when focused now.

screen shot 2018-05-09 at 8 58 46 am

LarryAmisola:
Thanks for the review Tim. I'm now working on this.

LarryAmisola:
1. The double box shadows was caused by the css properties of input:focus and .searchfield-wrapper.has-focus. To resolved the issue, I have added line 201. This will negate the box-shadow of the input:focus when the searchfield-wrapper.has-focus is available.
http://localhost:4000/components/searchfield/example-index.html
http://localhost:4000/components/modal/test-focus-box-shadow-cut

2. Added a css rule that will control only the display of the error message if immediately proceeded by the searchfield-wrapper.
http://localhost:4000/components/modal/test-required-fields

PR #34, opened 2018-05-10
Update README.md

nickwynja:

Updates readme to remove reference to `npm.infor.com`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,7 @@ We support the latest release and the release previous to the latest (R-1) for b
 
 ### npmjs
 
-This node module is hosted on a private repo [npm.infor.com](http://npm.infor.com:4873). To be able to install it in your project you need to tell npm on your system to fetch packages with the `@infor` scope from the correct registry.
-
-```bash
-npm config set @infor:registry=http://npm.infor.com:4873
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,13 +55,7 @@ We support the latest release and the release previous to the latest (R-1) for b
...
 ### npmjs
 
-This node module is hosted on a private repo [npm.infor.com](http://npm.infor.com:4873). To be able to install it in your project you need to tell npm on your system to fetch packages with the `@infor` scope from the correct registry.
-
-```bash
-npm config set @infor:registry=http://npm.infor.com:4873
-```
-
-Then to install, run:
+To install, run:
```
tmcconechy: Something missing here now? Do we just want to remove the whole section?

nickwynja: I think it's fine as is. You no longer need to configure the local `npm.infor.com` server before you can install.

tmcconechy: Yeah but it says:

+To install, run:

<nothing>

Or is that just the way the dif looks?

tmcconechy: Oh i see its the diff - sorry new to seeing these.

PR #39, opened 2018-05-11
This is a package for today (manual "nightly" build)

clepore:
I am bumping the version so we can get 2 more manual nightly builds up to npm for testing with NG and ng-quickstart.

### DIFF
--- a/publish/package.json
+++ b/publish/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ids-enterprise",
   "slug": "ids-enterprise",
-  "version": "4.7.0-dev",
+  "version": "4.7.0-dev.180511",
   "description": "Infor Design System (IDS) Enterprise Components for the web",
   "repository": {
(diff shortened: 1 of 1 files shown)


publish/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ids-enterprise",
   "slug": "ids-enterprise",
-  "version": "4.7.0-dev",
+  "version": "4.7.0-dev.180511",
```
tmcconechy: 4.6.0-dev.20180511 we had been using so far (4 digit date) - something me an nick settled on

nickwynja: I forgot we talked about `YYYY`. I'd be fine with either `YYYYMMDD` or `YYMMDD`. The latter is a bit shorter.

tmcconechy: ok im fine with either its a new system so doesnt matter i only did one so far

PR #40, opened 2018-05-11
Use apache 2.0 license

nickwynja:

Moves to Apache 2.0 license, which makes more sense for us at this point in time.


nickwynja:
@tmcconechy I was on a call with legal yesterday specifically about getting Terms and Conditions for `design.infor.com` and he was confused about the old license we had so figured this made more sense. He specifically had issues with the no-redistribution clause.

@benfenton What do you think about going to the Apache 2.0 license here for the soho codebase? Apache 2.0 still has an "as is" clause: https://github.com/infor-design/enterprise/pull/40/files#diff-9879d6db96fd29134fc802214163b95aR143

tmcconechy:
OK to merge or waiting on some discussion?

PR #43, opened 2018-05-14
SOHO-7935 Add findRowsByValue to datagrid control

micheal-cao:
current datagrid control does not provide a function for find rows by field value.  this delivery add such function to  datagrid control.

sometimes, we need find rows numbers based on a row's field name, for example, we need find a row based on a row's unique id, sometimes, such id is internal, so we could not find it by going through all cells, also, performance is another consideration.

SOHO-7935 Add findRowsByValue to datagrid control

the logic is similar to the existing function selectedRows(), which selectedRows checks row's selection status, this function checks value of the field of data for rows.

> Other Details:
n/a

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -5806,7 +5806,10 @@ Datagrid.prototype = {
   },
 
   /**
-   * Returns an array of row numbers for the rows containing the value for the field name.
+   * Returns an array of row numbers for the rows containing the value for the specified field.
+   * @param  {string} fieldName The field name to search.
+   * @param  {any} value The value to use in search.
+   * @returns {array} an array of row numbers.
    */
   findRowsByValue(fieldName, value) {
     const s = this.settings;
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -5805,6 +5805,33 @@ Datagrid.prototype = {
     return selectedRows;
   },
 
+  /**
+   * Returns an array of row numbers for the rows containing the value for the field name.
+   */
+  findRowsByValue(fieldName, value) {
+    const s = this.settings;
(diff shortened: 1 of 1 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -5805,6 +5805,33 @@ Datagrid.prototype = {
     return selectedRows;
   },
 
+  /**
+   * Returns an array of row numbers for the rows containing the value for the field name.
+   */
```
tmcconechy: Should add @params to the jsdoc

* @param {string}  [settings.fieldname] The field name to search
* @param {string}  [settings.value] The value to use in the search..

src/components/datagrid/datagrid.js
```diff
@@ -5805,6 +5805,33 @@ Datagrid.prototype = {
...
+  findRowsByValue(fieldName, value) {
+    const s = this.settings;
+    const dataset = s.treeGrid ? s.treeDepth : s.dataset;
```
tmcconechy: Does this work for tree grid? or untested?

tmcconechy:
Im wondering if you saw the keywordSearch feature? if that can be re-utilized but i guess i dont have a problem with adding something like this.

src/components/datagrid/datagrid.js
micheal-cao: Yes, It works for tree grid. I have tested in both regular table and tree table.

src/components/datagrid/datagrid.js
micheal-cao: thanks, i would update it later

micheal-cao:
keywordSearch searches cell text, but I hope that we could provide an ability of searching data model directly.

src/components/datagrid/datagrid.js
tmcconechy: Yes i'll hit that soon when i go through those.

PR #68, opened 2018-05-18
Better arguments for deploy script

nickwynja:

Provides more a more generic interface to `jenkins-deploy.sh` script so that it can also be used with publish scripts.

To build `master` to the demo server, run:
```
./scripts/jenkins-deploy.sh -b master
```
This is what will be used on merges into `master`.

To deploy a specific version to the demo server, like on release of a new version, run:
```
./scripts/jenkins-deploy.sh -b 4.6.1
```

You'll then want to run a second deploy on that version, to deploy to the `latest-enterprise` demo server, so that the `/code/ids-enterprise/latest/demo` links properly point to the `latest` version:

```
./scripts/jenkins-deploy.sh -b 4.6.1 -l
```

You can also add the `-w` flag to watch the status of the build and exit according to the status of the Jenkins build. This is good for CI where you want to track whether the Jenkins deploy was good but might not be wanted for a local deploy script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   - npm run functional:ci
 deploy:
   - provider: script
-    script: ./scripts/jenkins-deploy.sh master
+    script: ./scripts/jenkins-deploy.sh -b master -w
     skip_cleanup: true
     on:
       branch: master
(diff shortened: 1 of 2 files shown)


nickwynja:
@clepore `-b` is `BUILD_FROM` not `branch`. We're currently doing this in the Jenkins script:

```sh
if [ "$GIT_TAG_OR_BRANCH" = "branch" ]
then
    BUILD_NAME=$GIT_BRANCH-$CONTAINER
	git checkout $GIT_BRANCH
else
	VERSION_STRING=$(echo "${GIT_TAG//./}") # Removes . to not break subdomain
	BUILD_NAME=$VERSION_STRING-$CONTAINER
	git checkout tags/$GIT_TAG
fi
```

But it totally makes sense to just use `git checkout $BUILD_FROM` for both rather than using `git checkout tags/$BUILD_FROM` and not need the `-t` (tag) flag.

I'll update both scripts.

PR #76, opened 2018-05-21
SOHO-7225 - Added support to alert type icon with Validation

deep7102:
http://localhost:4000/components/validation/test-message-types-landmark


app/views/components/validation/test-alert-types.html
```diff
@@ -58,6 +68,23 @@
       type: 'info'
     };
 
+    $.fn.validation.rules.customIconRule1 = {
+      check: function (value) {
+        return (value !== '');
+      },
+      message: 'Something about your user profile',
+      type: 'icon'
```
tmcconechy: Does this need an icon name?

PR #86, opened 2018-05-23
SOHO-7902 Updated select and deselect functions to prevent toggling o…

tmcastro:
- Current issue is happening because select function is used as a toggle instead of just selecting the item
- Updated select and deselect functions so that using select function on already selected items will not toggle it
- Moved some code to deselect function so that it will no longer need to fire the select function
- http://localhost:4000/components/listview/test-deselect.html

- Firing the select function on already selected items should not deselect/toggle the item

### DIFF
--- a/src/components/listview/listview.js
+++ b/src/components/listview/listview.js
@@ -722,8 +722,65 @@ ListView.prototype = {
     if (typeof li === 'number') {
       li = $(this.element.children()[0]).children().eq(li);
     }
+
+    const self = this;
     if (li.is('.is-selected')) {
-      this.select(li);
+      this.selectedItems = [];
+      li.removeClass('is-selected hide-selected-color');
+
+      li.attr('aria-selected', false);
+      li.find('.listview-selection-checkbox input').prop('checked', false);
+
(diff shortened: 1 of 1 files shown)


src/components/listview/listview.js
```diff
@@ -747,7 +804,7 @@ ListView.prototype = {
     let isChecked = false;
     const isMixed = self.settings.selectable === 'mixed';
 
-    self.selectedItems = [];
+    // self.selectedItems = [];
```
tmcconechy: Remove vs comment out if un needed.

src/components/listview/listview.js
```diff
@@ -722,8 +722,65 @@ ListView.prototype = {
     if (typeof li === 'number') {
```
tmcconechy: Looks like this change is also on https://github.com/infor-design/enterprise/pull/87

tmcconechy:
Testing this out i see it introduced one bug i could find.

http://localhost:4000/components/listview/example-singleselect.html is now not behaving as single select anymore. http://usalvlhlpool1.infor.com/4.6.0-rc/components/listview/example-singleselect.html

PR #89, opened 2018-05-23
Adds exceptions for google fonts and *.design.infor.com

nickwynja:

Adds CSP exceptions to allow proxy of app in `design.infor.com`.


nickwynja:
Seems like the server isn't recognizing the updated `directives`. Looking into it.

PR #92, opened 2018-05-24
Patch app.js for demo server

nickwynja:

This change is necessary to run HTTPS and proxy on the `latest` demo server. This will essentially deploy `4.6.1` plus the app server patch.

It's not necessary at this time to bump a patch version of `4.6.x`. I think this fix is fine sitting on the tip of that branch and could go out if there's ever a need for `4.6.2`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -10,7 +10,7 @@ let express = require('express'),
   http = require('http'),
   BASE_PATH = process.env.BASEPATH || '/',
   fullBasePath = function (req) {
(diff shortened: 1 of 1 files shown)


nickwynja:
This seems like a strange case where CI isn't kicking off because I initially made this change in the Github web file editor. I'll keep an eye on CI since it _seeemed_ to be back to normal prior to this.

PR #93, opened 2018-05-24
Soho 7999 spinbox minmax

tmcconechy:

Fixed an issue with initializing spinbox with min, max and step
Added first pass at spinbox tests

soho-7999

Run the tests
and checkout http://localhost:4000/components/spinbox/list pages

> **Other Information**:

@d-g-h Take a look at the e2e tests and see what you think about the sleep situation.


d-g-h:
It's fine for now. Seems totally necessary, I will revisit once this gets in though... real quick.

EdwardCoyle:
Functionality changes seem good!  Ran tests with no problems.

Hitting the examples manually, I just found one thing with the body copy on the range limits example:  http://localhost:4000/components/spinbox/example-range-limits.html.  Probably want to change the text to not read "-50" and "50" for the limits, or remove it completely.

screen shot 2018-05-24 at 11 55 19 am

PR #94, opened 2018-05-24
SOHO-8006 fixed the extra padding in popupmenu

LarryAmisola:
Get the descendants of li element specifically an svg with class selector icon not close and dropdown. This will served as checker if a ul should have a class 'has-icon' or not. The aim of this fix is to removed the extra padding in the item list when a list doesn't contained an svg icon.

No padding
http://localhost:4000/components/popupmenu/example-index.html
http://localhost:4000/components/popupmenu/example-menubutton.html

with padding
http://localhost:4000/components/popupmenu/example-icons.html


tmcconechy:
As discussed with @LarryAmisola i pushed a similar PR with a test https://github.com/infor-design/enterprise/pull/95

So closing this one.

PR #84, opened 2018-05-22
Cleanup "publishing" for NPM package

clepore:
### Explain the **details** for making this change. What existing problem does the pull request solve?

- [x] Remove old documentation
- [x] Change the banner name for versioning the components
- [x] Reorg the npm scripts to make more "readable" and cooperate more with CI
- [x] Rename `publish/` to its package name (aka `ids-enterprise`)

### **Steps necessary to review your pull request (required)**:

- Run all build scripts
  - [ ] `npm start`
  - [ ] `npm run publish-setup`
- [ ] Verify 0 errors and that the app works
- [ ] Check/grep documentation to make sure that I updated it to match these changes

PR #105, opened 2018-05-28
Fix package.json location in nightly script

nickwynja:

The changes in infor-design/enterprise#84 broke the nightly script. This fixes it.

infor-design/enterprise#84


nickwynja:
Using admin rights to merge this since everyone is off today and can't approve.

PR #113, opened 2018-05-29
Added watch command

tmcconechy:
Fixed Csp Issue
Fixed issue on accordion sample page

Add watch to npm command list

Fixed sample page:
http://localhost:4000/components/accordion/example-ajax.html

https://jira.infor.com/browse/SOHO-8046

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "quickstart": "npx grunt demo && node app/server.js",
     "build": "npx grunt",
     "stop": "./scripts/stop.sh",
+    "watch": "npx grunt watch",
     "eslint": "npx eslint -c .eslintrc.js src test",
     "eslint:log": "npx eslint -c .eslintrc.js src test -o eslint.log",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
     "quickstart": "npx grunt demo && node app/server.js",
     "build": "npx grunt",
     "stop": "./scripts/stop.sh",
+    "watch": "npx grunt watch",
```
d-g-h: Drop for `npm run build -- watch` in the docs?

tmcconechy: At first i thought that was the correct command too. But im trying to put `grunt watch` in the script commands the `npm run build -- watch` seems to just build and stop and not actually watch.

d-g-h: screen shot 2018-05-29 at 3 48 17 pm

tmcconechy:
(thumbs) i didnt have a space before the --

PR #114, opened 2018-05-29
CSP Query Parameter fix for the Demo App

EdwardCoyle:

Fixed CSP query param setting to actually disable CSP if set to "false".  Hitting any URL like so:

http://localhost:4000/kitchen-sink?csp=false

...will disable CSP.  In some cases, we can use this for OS/Browsers/Environments that have issues with our current CSP definitions.


tmcconechy:
May need to merge with https://github.com/infor-design/enterprise/pull/113

EdwardCoyle:
@clepore we resolved the conflict here yesterday

PR #117, opened 2018-05-29
Soho 7816 dropdown performance

deep7102:
http://localhost:4000/components/multiselect/test-select-all-performance


src/components/dropdown/dropdown.js
```diff
@@ -1953,6 +1947,58 @@ Dropdown.prototype = {
     }
   },
 
+  /**
+   * Toggle all selection for items.
+   * @private
+   * @param {boolean} doSelectAll true to select and false will clear selection for all items.
+   * @returns {void}
+   */
+  toggleAllSelection(doSelectAll) {
```
tmcconechy: Maybe a better function name would be selectAll / deselectAll. And could we add a functional test?

tmcconechy:
This does seem a LOT faster before it would totally hang. But what i was noticing was it was only when clicking on the checkbox. Did you explore why that might be?

src/components/dropdown/dropdown.js
```diff
@@ -1953,6 +1947,58 @@ Dropdown.prototype = {
...
+      }
+    } else {
+      // Clear all
+      items.forEach(node => node.classList.remove('is-selected'));
+      options.forEach((node) => {
+        node.selected = false;
+      });
+    }
+
+    this.pseudoElem.find('span').text(text);
```
d-g-h: Try to drop jQuery here as well? `find, text, val`...

src/components/dropdown/dropdown.js
d-g-h: Yeah, we should add. Should have a e2e test for this already via dropdown/multiselect... Let's run the e2e on the branch in the current state, I will do it now also...

test/components/multiselect/multiselect.e2e-spec.js
```diff
@@ -193,3 +205,50 @@ describe('Multiselect example-clear-all tests', () => {
     });
   }
 });
+
+describe('Multiselect example-select-all-performance tests', () => {
+  beforeEach(async () => {
+    await utils.setPage('/components/multiselect/test-select-all-performance');
+  });
+
+  it('Should select all performance test', async () => {
```
d-g-h: @deep7102  I believe results are going to vary from machine to machine. For me, I had lots of apps open and my compy was slower, and all the tests failed. Upon closing of those open apps, the tests passed. My worry is that this will not work on the CI, or in BrowserStack.
Is there any other way we can test perform other than time? Is this test really needed once we show performance improvements?

d-g-h: screen shot 2018-06-04 at 4 27 10 pm

tmcconechy:
What if we set this up to 5000? Or would that slow down the tests 5 seconds - i think would be ok because it will finish earlier mostly.

d-g-h:
That time should be OK, but I thought Deepak had set an expected threshold time. So like, if we don't meet the performance/timing expectation the test fails...

tmcconechy:
The issue was it actually took so long it froze the browser so maybe that helps some. Unless we ignore this test or find another way. I.E.

The test could just test all and then see if all are selected after. That way its not based on time at all. And if it doesnt complete we would get a test timeout/fail? What you think @deep7102 ?

PR #121, opened 2018-05-30
Button tokens

jmacaluso711:
- Install ids-identity
- Starting to replace button vars with tokens. 
- Made minor adjustments to _button.scss where necessary

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -19,13 +19,13 @@ button {
   @include css3-extended(backface-visibility, hidden, (-webkit-, -moz-));
   @include css3-extended(transform, translate3d(0, 0, 0), (-webkit-, -moz-));
   @include css3-extended(color, transparent, (-webkit-tap-highlight-));
-  @include font-size($button-font-size);
   @include transition(color 300ms cubic-bezier(.17, .04, .03, .94), background-color 300ms cubic-bezier(.17, .04, .03, .94), border 300ms cubic-bezier(.17, .04, .03, .94), box-shadow 300ms cubic-bezier(.17, .04, .03, .94));
   @include antialiased();
   color: inherit;
-  font-weight: 700;
-  height: 34px;
-  line-height: normal;
+  font-size: $button-font-size;
+  font-weight: $button-font-weight;
+  height: $button-height;
+  line-height: $button-line-height;
   min-width: 100px;
   overflow: hidden;
   position: relative;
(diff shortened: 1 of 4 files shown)


src/components/button/_button.scss
```diff
@@ -207,7 +207,7 @@ a.btn-secondary,
 a.btn-tertiary,
 a.btn-menu {
   display: inline-block;
-  line-height: 31px;
+  line-height: 38px;
```
clepore: Should this use a token?

tmcconechy:
How can i see whats in ids-identity ? Maybe i / we dont have git access is i dont see it here https://github.com/infor-design?

src/components/button/_button.scss
tmcconechy: This also seems like a high level of change (7px) we might need to test text is centered on all browsers?

clepore: Yes I agree

clepore:
It's all open source so you should see it. `ids-identity` is part of https://github.com/infor-design/design-system and is also on npmjs.

src/components/button/_button.scss
clepore: screen shot 2018-05-30 at 12 57 28 pm

src/core/_config.scss
```diff
@@ -1,6 +1,10 @@
 // Variables
 //================================================== //
 
+// IDS Identity
+@import '../../node_modules/ids-identity/dist/tokens/web/theme-default.sass';
```
tmcconechy: What about the other 3 themes? Where they tested / should this be added for them?

src/core/_config.scss
```diff
@@ -152,36 +156,36 @@ $popupmenu-dark-box-shadow: 0 2px 5px $drop-shadow-depth;
...
-$primary-btn-bg-color: $primary-color;
-$primary-btn-color: $inverse-color;
-$primary-btn-hover-bg-color: $azure07;
-$primary-btn-focus-bg-color: $primary-color;
-$primary-btn-active-bg-color: $azure07;
-$primary-btn-disabled-color: $inverse-color;
-$primary-btn-disabled-bg-color: $primary-color; //Primary at 50%
-$primary-btn-disabled-border-color: $primary-color;
-$primary-btn-disabled-opacity: .5;
+$primary-btn-bg-color: $button-primary-bg-color;
```
tmcconechy: Wondering if any merrit to changing our variable names to match

i.e
$primary-btn-bg-color -> $button-primary-bg-color

And then all locations in all files (bit more of a change though)

tmcconechy:
This changes the button size to be much bigger. 

untitled

clepore:
@tmcconechy Here's a link to the issue (https://github.com/infor-design/design-system/issues/141) we are currently working off of. We have a meeting with the designs in a bit, but that should give you a bit of insight of the goal.  It's a bit of a work in progress.

PR #134, opened 2018-06-04
Soho 8016 data grid editing control records per page not working

tmcastro:
- Updated the settings to not include the "source" option
- Source option is being validated on the data grid component and "function() {}" is still considered true
- http://localhost:4000/components/datagrid/example-editable.html

- Pager should be working fine now


tmcastro:
Thanks, @tmcconechy! I'll do that..

Please disregard this PR.. I've created a new one.

PR #144, opened 2018-06-05
SOHO-8061 Run e2e on CI w/o errors

d-g-h:

Runs CI test on a non-BrowserStack environment on Trusty (Travis), where tests should run nearly the same as local:debug:conf.js

https://jira.infor.com/browse/SOHO-8061
https://jira.infor.com/browse/SOHO-7331

Run all e2e commands, use your newly created BS keys, especially `npm start && npm run e2e:ci`


tmcconechy:
Is it failing because it needs the change?

d-g-h:
Skipped it, I will rework later, and make a ticket

.travis.yml
```diff
@@ -14,7 +18,10 @@ before_script:
   - npm run eslint:error-only
 script:
   - npm run build
+  - npm run quickstart &
```
clepore: ❓ Do we need to run `build` and `quickstart`, or could we just do `npm start` (which does both)?

test/components/timepicker/timepicker.e2e-spec.js
```diff
@@ -9,12 +9,9 @@ const ddSelector = function (dropdown) {
   return `#timepicker-popup select.dropdown.${dropdown} + .dropdown-wrapper div[aria-controls="dropdown-list"]`;
 };
 
-let timepickerEl;
```
clepore: ❓ I see originally these were scoped global and then could be set in the `beforeEach()`, but now we are repeating them everywhere. Is there a reason for that? timing?

package.json
```diff
@@ -11,9 +11,10 @@
     "eslint:error-only": "npx eslint -c .eslintrc.js --quiet src test",
     "functional:local": "npx karma start test/karma.conf.js",
     "functional:ci": "npx karma start test/karma.conf.js --single-run",
-    "e2e:local": "npm run webdriver:update && npx protractor test/protractor.local.conf.js",
-    "e2e:bs": "npm run webdriver:update && npx protractor test/protractor.bs.conf.js",
     "e2e:ci": "npm run webdriver:update && npx protractor test/protractor.ci.conf.js",
+    "e2e:ci:bs": "npx protractor test/protractor.ci.bs.conf.js",
+    "e2e:local:bs": "npm run webdriver:update && npx protractor test/protractor.local.bs.conf.js",
+    "e2e:local:debug": "npm run webdriver:update && npx protractor test/protractor.local.debug.conf.js",
```
tmcconechy: Actually which one will run e2e local headless chrome now? with no browser poping up?

d-g-h: `npm run e2e:ci`

test/components/timepicker/timepicker.e2e-spec.js
d-g-h: Yes, timing, and isolation. The variable names are the same but the selectors are different (use different \#s). Best to requery an element, rather than use a cached selector. Selenium will often complaint about "stale elements". The test suite needs improving hence https://jira.infor.com/browse/SOHO-8098, and possibly more issues to follow.

.travis.yml
d-g-h: To save time after the build, we can start the demoserver w/`&` in a different process, and run the functional tests simultaneously which are not dependent on the demoserver

PR #148, opened 2018-06-06
Soho 8042 submit button validation

tjamesallen15:
Need to fix for the Submit Buttion Enable and Disable Validation

https://jira.infor.com/browse/SOHO-8042

http://localhost:4000/components/modal/example-validation.html
Fill up the required fields.
See if the Submit Button is now enabled.

> Other Details:

### DIFF
--- a/src/components/validation/validation.js
+++ b/src/components/validation/validation.js
@@ -90,8 +90,18 @@ function ValidationRules() {
             const elem = $(this);
             const fieldValue = elem.is('.editor') ? elem.html() : elem.val();
 
+            if (elem[0].getAttribute('data-validate').trim() !== 'required') {
+              const regex = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,16}(?:\.[a-z]{2})?)$/i;
+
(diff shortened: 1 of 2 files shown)


src/components/validation/validation.js
```diff
@@ -90,8 +90,18 @@ function ValidationRules() {
             const elem = $(this);
             const fieldValue = elem.is('.editor') ? elem.html() : elem.val();
 
+            if (elem[0].getAttribute('data-validate').trim() !== 'required') {
+              const regex = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,16}(?:\.[a-z]{2})?)$/i;
```
tmcconechy: I think what should actually happen is it should use the specific validation check (see rules).
Basically you have recreated the check that exists in the rules.

https://github.com/infor-design/enterprise/blob/master/src/components/validation/validation.js#L246

Perhaps there is a way to simply check if the field has an error on it. https://github.com/infor-design/enterprise/blob/master/src/components/validation/validator.js#L575

This way would work on any validation type.

tjamesallen15:
Thanks Tim and Ed! :)

PR #153, opened 2018-06-06
Cleanup Dist and Npm Publish Folders

tmcconechy:
Cleanup Publish to publish whole sass folder structure so imports work.
Recreate yarn and package.json lock

Cleanup Dist Folder to Remove incorrect "all" folder.  
Cleanup Publish to publish whole sass folder structure so imports work.  
Recreate yarn and package.json lock

https://jira.infor.com/browse/SOHO-7583

Nightly publish...will show sass folder updated  
`npm grunt packup` -> check folder in ids-enterprise  
`grunt` will not show all folder and zip now.


.gitignore
```diff
@@ -22,6 +23,7 @@ ids-enterprise/LICENSE
 templates/**/doc.html
 test-reports/
 
+
```
d-g-h: Drop unneeded whitespace?

scripts/configs/compress.js
```diff
@@ -1,18 +0,0 @@
-module.exports = {
-
-  compress: {
-    main: {
-      options: {
-        archive: 'dist/all.zip'
```
d-g-h: Dropping the .zip? Cool, this should speed things.

.gitignore
clepore: We should have a `publish/` folder to need ignored? The new publish folder is `ids-enterprise`

scripts/configs/compress.js
tmcconechy: Yeah it was only needed for the old bamboo build as it had an assets tab. But where would someone now download the assets if they dont use npm now? Hmmm

d-g-h: download the source or clone, and run `npm run install && build`, `zip -r dist`?

.gitignore
tmcconechy: When i ran `npm grunt packup` i ended up with a publish folder somehow. Maybe another change is needed. Let me see.

scripts/configs/compress.js
tmcconechy: Talking to chris, we can now upload a "build" to the release page in git. This will work well and replace the need for this...

d-g-h: Oh yeah, I forgot about, that's the expected location...

clepore: One can also download the package from the npm site too, i believe.

PR #158, opened 2018-06-07
Re-organize enterprise to use webpro's release-it script

clepore:
**Explain the **details** for making this change. What existing problem does the pull request solve?**

- [x] Consolidated package.json's
- [x] Added release-it configuration
    - github release includes zipped `dist/` file
    - npm commands for different types of releases
    - do NOT include release-it as a dependency (only needed globally by a few people)
- [x] updated release documentation
- [x] Added `sass/` dir to `dist/` for distribution

- [ ] Run `npm pack` and unzip that file to verify the contents are the same as before
- [ ] Verify the follow commands work correctly
    - `npm run start`
    - `npm run quickstart`
    - `npm run zip-dist`
    - any `grunt` commands I might have missed

**Other Details:**

Don't worry about the release script commands for now.


tmcconechy:
Can release-it run the AWS publish command (it published the release to a CDN)... Just an idea.

clepore:
@tmcconechy it can, should that run last with the documentation?

tmcconechy:
Yeah i'd say run it last at the end on actual release. Fx 4.8.0 

See: 
https://github.com/infor-design/enterprise/blob/master/docs/PUBLISH.md#deploy-to-aws
(Have to change from publish folder to dist now).

clepore:
We'll see if its still required now that the requesting parties can use NPM. Plus they can always download from github or npm.

PR #164, opened 2018-06-08
SOHO-5894 Fix Main Pane Text not visible when changing themes

tjamesallen15:

https://jira.infor.com/browse/SOHO-5894

http://localhost:4000/layouts/three-column-fixed-right.html
Change Theme to Dark Contrast. See middle pane if the text can be seen.

**Closing issues**

### DIFF
--- a/src/layouts/_layouts.scss
+++ b/src/layouts/_layouts.scss
@@ -236,10 +236,7 @@ body.no-scroll {
     border-left: 1px solid $graphite02;
     border-right: 1px solid $graphite02;
     width: calc(60% - 3px);
-
-    p, h1, h2, h3, h4, h5, h6 {
-      color: $font-color-inverted;
-    }
+    background-color: $panel-bg-color;
 
     &.scrollable,
     &.scrollable-y {
--- a/src/layouts/_layouts.scss
+++ b/src/layouts/_layouts.scss
@@ -237,6 +237,10 @@ body.no-scroll {
     border-right: 1px solid $graphite02;
     width: calc(60% - 3px);
 
+    p, h1, h2, h3, h4, h5, h6 {
+      color: $font-color-inverted;
+    }
+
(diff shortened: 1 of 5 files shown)


src/layouts/_layouts.scss
```diff
@@ -237,6 +237,10 @@ body.no-scroll {
     border-right: 1px solid $graphite02;
     width: calc(60% - 3px);
 
+    p, h1, h2, h3, h4, h5, h6 {
+      color: $font-color-inverted;
```
EdwardCoyle: I think using the inverted font color here isn't necessarily the right answer...  I see the following after testing the branch:

dark-theme
high-contrast-theme
light-theme

Honestly I think the only one here that's correct is the high contrast theme.  The Dark Theme's major problem is that the `.main` section's background shouldn't be white to begin with, since it's supposed to be a dark theme.  The font color and background color should probably be reversed there.

I think @LarryAmisola's assessment of the Light theme is still true - the font is white on a white background.  The fix here should probably be switching this font color to `$font-color-default`, and fixing the background color of `.main` on the Dark Theme.

@tmcconechy any thoughts?

tjamesallen15: How did you produce that error? I can't see it on my local.

tjamesallen15: sample

tmcconechy: You can reproduce it by changing the theme to dark / light / high contrast. Ed is correct.
Test http://localhost:4000/layouts/scrolling-two-column.html and compare the panes should be...

light - white and teh grey
dark - dark grey and black
high contrast - grey and darker grey

So the white pannel in the middle is wrong it should actually be using $panel-bg-color vs $body-bg-color

So the problem is really maybe that the middle is not using $panel-bg-color

PR #165, opened 2018-06-11
SOHO-7723 fix button alignment issue

LarryAmisola:
Added CSS fix for the button alignment issue
 
SOHO-7723

http://localhost:4000/components/tabs/example-beforeactivate-callback
- open the link above then resize the window
- Expected result: Buttons should have a 5px space/gap vertically when in stacked

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -642,7 +642,14 @@ a.btn-menu {
         }
       }
     }
- 	}
+  }
+  button {
+    margin-bottom: 5px;
+
+    &:last-child {
+      margin-bottom: 0;
+    }
+  }
 }
 
 // Short field buttons


src/components/button/_button.scss
```diff
@@ -642,7 +642,14 @@ a.btn-menu {
         }
       }
     }
- 	}
+  }
+  button {
+    margin-bottom: 5px;
```
tmcconechy: Im not sure if this is a good idea because it will effect every button all over the place.
I think instead we can do https://jira.infor.com/browse/SOHO-7994 and change the samples.

The layout issue is super minor on further review i think we should cancel this to avoid any impact and then just close SOHO-7723 as wont do.

tmcconechy:
Decided we shouldnt do https://jira.infor.com/browse/SOHO-7723 and just will fix it in the way mentioned on https://jira.infor.com/browse/SOHO-7994 sometime in the future.

PR #166, opened 2018-06-11
Jira Cleanup for SOHO-7402, SOHO-7411, and SOHO-7246 

tmcconechy:

**SOHO-7402** - Fixed bug on listview with scroll container after calling removeSelectedItems
**SOHO-7411** - Fixed bug on init of header menu popup
**SOHO-7246** - Allow exportToCsv and exportToExcel to work without a datagrid.

**SOHO-7402** 
http://localhost:4000/components/listview/example-multiselect.html
- select several items
- note that the list is the correct scroll height
- hit remove
- note that the list is the correct scroll height still

**SOHO-7402** 
http://localhost:4000/patterns/navigation-popupmenu.html
- Page should be set to "Page Two" (as specified in the markdown)

**SOHO-7246** - Allow exportToCsv and exportToExcel to work without a datagrid.
http://localhost:4000/components/datagrid/test-export-without-datagrid.html
- click export
check the file.

**Closing issues**

SOHO-7402
SOHO-7411
SOHO-7246


src/utils/excel.js
```diff
@@ -68,11 +50,13 @@ excel.exportToCsv = function (fileName, customDs, self) {
 
     // CHECK EXPORTABLE
     const nonExportables = [];
-    $.each($('th', self.headerRow).not('.is-hidden'), (index, item) => {
-      if ($(item)[0].getAttribute('data-exportable') && $(item)[0].getAttribute('data-exportable') === 'no') {
-        nonExportables.push(index);
-      }
-    });
+    if (self) {
```
EdwardCoyle: Not sure if this is right, Excel seems to throw a warning on the exported file:
screen shot 2018-06-13 at 10 05 27 am

Seems that the file opens and looks OK though...
screen shot 2018-06-13 at 10 05 41 am

tmcconechy:
This is a known issue. It happens no matter what you do with the excel type export and has always been that way. Its just saying "do you trust this file". Its a bit disconcerting but we havent ever found a way around this.

tmcconechy:
Fixed that @clepore - the example had the extension in the file name option, but it also gets added by the function so it duped

PR #169, opened 2018-06-12
SOHO-7511: Comment out possibly no needed case

guoliang:

Fixes issue in https://jira.infor.com/browse/SOHO-7511

https://jira.infor.com/browse/SOHO-7511

In my case, I could not open a new tab which is hosting a an Angular component.

**Closing issues**

### DIFF
--- a/src/components/tabs/tabs.js
+++ b/src/components/tabs/tabs.js
@@ -2247,7 +2247,7 @@ Tabs.prototype = {
           // Text Content or a Selector.
           if (hasId !== null) {
             const obj = $(sourceString);
-            sourceString = obj.length ? $(sourceString).clone() : sourceString;
+            sourceString = obj.length ? obj : sourceString;
           }
           // do nothing if it's just a string of text.
           break;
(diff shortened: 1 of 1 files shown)


src/components/tabs/tabs.js
```diff
@@ -2251,10 +2251,10 @@ Tabs.prototype = {
           }
           // do nothing if it's just a string of text.
           break;
-        case 'object':
-          // jQuery object or HTML Element
-          sourceString = $(sourceString).clone();
-          break;
+        // case 'object':
+        //   // jQuery object or HTML Element
```
EdwardCoyle: Completely remove this case instead of just commenting it out

src/components/tabs/tabs.js
```diff
@@ -2251,10 +2251,10 @@ Tabs.prototype = {
           }
```
EdwardCoyle: On line 2250:
```javascript
sourceString = obj.length ? $(sourceString).clone() : sourceString;
```
Should probably also remove the clone from the ternary statement as well

guoliang: Fixed

src/components/tabs/tabs.js
guoliang: Fixed

PR #176, opened 2018-06-14
SOHO-8124 - validateForm with an async rule does not return a sucess or failure to the callback function

marknolaninfor:
SOHO-8123 and SOHO-8124 fixes to the async validation, to ensure that the rule and deferred objects are correct when the rule result is being resolved.


marknolaninfor:
I have corrected the build issue and also added a change event to the form, so that the submit button is disabled until the validated event returns.

src/components/datagrid/datagrid.js
```diff
@@ -2858,7 +2858,7 @@ Datagrid.prototype = {
       // Set Width of table col / col group elements
       let colWidth = '';
 
-      if (this.recordCount === 0 || this.recordCount - ((activePage - 1) * pagesize) === 0) {
+      if (this.recordCount === 0) {
```
tmcconechy: Is this related/intentional?

EdwardCoyle:
@marknolaninfor seems like we're just in need of some cleanup on an unused var: https://travis-ci.com/infor-design/enterprise/builds/76404654#L509

I'll approve once that's fixed.

PR #177, opened 2018-06-14
SOHO-8030 Fix cursor tracker showing.

tjamesallen15:

https://jira.infor.com/browse/SOHO-8030

http://localhost:4000/components/colorpicker/example-show-label

**Closing issues**

### DIFF
--- a/src/components/colorpicker/_colorpicker.scss
+++ b/src/components/colorpicker/_colorpicker.scss
@@ -80,6 +80,11 @@
         border-color: $input-border-color;
       }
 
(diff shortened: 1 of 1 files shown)


tmcconechy:
I thought it want much of one. But when i tried this and since the field is readonly it seemed ok with this change to me. The cursor is never going to do anything because the field is readonly.

So to me its borderline bug - minor. But i would either agree to keep this or cancel the bug either way.

tjamesallen15:
Thanks for all the comments, guys!

PR #180, opened 2018-06-14
Fixing field filter wrapper on dropdown fields

### LINKED ISSUE #179: Dropdowns on Field Filter don't display correctly
**Describe the bug**
Dropdowns aren't being wrapped correctly by the field filter control.

**To Reproduce**
Add this html to field-filter/example-index.html

     <div class="field">
      <label for="example-dropdown" class="label">Dropdown</label>
      <select id="example-dropdown" class="dropdown field-filter input-md">
        <option value=""></option>
        <option value="1">Inquiry</option>
        <option value="2">Setup</option>
      </select>
    </div>

1. Navigate to /components/field-filter/example-index.html
2. See error

**Expected behavior**
Dropdown should match how the other fields looks for field filter

**Screenshots**
image

**Platform**
 - All

Sovia:
Fixing the field filter wrapper for dropdowns

[https://github.com/infor-design/enterprise/issues/179](https://github.com/infor-design/enterprise/issues/179)

Navigating to the example-index.html in field filter should now look like:
image

**Closing issues**
`closes #179`

### DIFF
--- a/src/components/field-filter/field-filter.js
+++ b/src/components/field-filter/field-filter.js
@@ -11,22 +11,22 @@ import '../dropdown/dropdown.jquery';
 const COMPONENT_NAME = 'fieldfilter';
 
 /**
-* Ability to have a dropdown next to the field.
-*
-* @class FieldFilter
-* @constructor
-*
-* @param {jQuery[]|HTMLElement} element The component element.
-* @param {object} [settings] The component settings.
-* @param {array} [settings.dataset]  Array of data
-* @param {object} [settings.dropdownOpts]  Gets passed to this control's dropdown
-* @param {string} [settings.template] An Html String with the mustache template for the view.
-*/
+ * Ability to have a dropdown next to the field.
+ *
+ * @class FieldFilter
+ * @constructor
+ *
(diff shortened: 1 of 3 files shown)


src/components/field-filter/field-filter.js
```diff
@@ -125,7 +125,7 @@ FieldFilter.prototype = {
 
       // Add css classes
       this.field.addClass('fieldfilter-wrapper')
-        .find('div.dropdown span').addClass('audible');
+        .find('div.dropdown.no-init span').addClass('audible');
```
tmcconechy: Why is the no-init needed?

Sovia: Without the no-init, it was also adding the audible class to the span inside the dropdown field that was being wrapped. This caused the value to never appear selected when the dropdown wasn't open (even when a value was selected)

EdwardCoyle: We should find a way to do it without the `no-init` class being a dependency.  `no-init` isn't really supposed to be used directly by the components in this manner - it's meant as a tool for app developers to control the flow of component initialization.  If we use it internally and remove it automatically, it could cause unexpected results for an app developer depending on it.

PR #189, opened 2018-06-18
Repurpose "test" as it was a duplicate cmd

clepore:

- `npm test` is a duplicate cmd, we can move the `all-checks` script content to it and avoid having to have another cmd in the list

- n/a

- make sure `npm run test` isn't used other places where the server might be required, yet unavailable


clepore:
Yeah `test` tends to be the general "all tests" for npm scripts, so I figured we get rid of the dupe (not from @tmcconechy ) and then migrate his script to `npm test` to avoid having another script.

tmcconechy:
Ok cool - works for me just wanted a way to run everything

PR #198, opened 2018-06-20
SOHO-4528 add focus on pagination numbers

LarryAmisola:
Add a blur event to handle callback that will remove a class

SOHO-4528

http://localhost:4000/components/pager/example-index.html
- Click the pagination number and a focus state will display
- Click the page then pagination number will be unfocus

### DIFF
--- a/src/components/pager/pager.js
+++ b/src/components/pager/pager.js
@@ -440,7 +440,7 @@ Pager.prototype = {
 
       for (i = pages; i > 0; i--) {
         if (i === (this.activePage || 1)) {
-          thisClass = 'class="selected"';
+          thisClass = 'class="selected is-focused"';
           thisText = Locale.translate('PageOn');
           isAriaSelected = 'aria-selected="true"';
           isAriaDisabled = 'aria-disabled="true"';
@@ -453,6 +453,11 @@ Pager.prototype = {
 
         $(`<li ${thisClass}${isAriaSelected}><a href="#" ${isAriaDisabled}><span class="audible">${thisText} </span>${i}</a></li>`).insertAfter(this.pagerBar.find('.pager-prev'));
       }
+      $(document).off('click').on('click', (e) => {
+        if (!$(e.target).parent().hasClass('pager-toolbar')) {
+          self.pagerBar.find(PAGER_NON_NUMBER_BUTTON_SELECTOR).removeClass('is-focused');
+        }
+      });
     }
 
     if (this.isTable && !this.settings.indeterminate && this.pagerBar.find('.pager-count').length === 0) {


src/components/pager/pager.js
```diff
@@ -453,6 +453,11 @@ Pager.prototype = {
 
         $(`<li ${thisClass}${isAriaSelected}><a href="#" ${isAriaDisabled}><span class="audible">${thisText} </span>${i}</a></li>`).insertAfter(this.pagerBar.find('.pager-prev'));
       }
+      $(document).off('click').on('click', (e) => {
```
LarryAmisola: I need your input on this lines. I'm still looking for a better way to remove a class name 'is-focus'

tmcconechy:
I would try to refocus the button after the page is loaded. Using the current code that adds the class.
Hopefully as simple as a .focus() in the right place?

src/components/pager/pager.js
EdwardCoyle: @LarryAmisola I had this problem with the Searchfield recently.  I think you're close here, but I don't recommend the `.off()` then `on()` method, especially with `$(document)` listeners, since it has the potential to wreak havoc with other components that are using `$(document).on('click')`.

I would suggest setting up a `$(document).on('click.pager')` (read: namespaced) event listener under `handleEvents` that persists the entire time, and do your element detection there.  I would also suggest hooking into the [existing blur handler](https://github.com/infor-design/enterprise/blob/soho-7580-merge-searchfield-with-tbsf/src/components/pager/pager.js#L270), and running the same event listener on a short setTimeout (probably 0 is enough), so that all the buttons will lose their focus when a keyboard key like Tab, or an arrow key, is used to change focus.  That way, you cover all the accessibility bases.

Also,  for reference, here's how I did that in the recent Searchfield changes:
- the [`handleSafeBlur()`](https://github.com/infor-design/enterprise/blob/soho-7580-merge-searchfield-with-tbsf/src/components/searchfield/searchfield.js#L889) method runs an internal check on a setTimeout, which staggers the check until the browser has time to swap focus.  This ensures that the element isn't still focused while the classes are removed.
- [`get isFocused()`](https://github.com/infor-design/enterprise/blob/soho-7580-merge-searchfield-with-tbsf/src/components/searchfield/searchfield.js#L820) is the actual check on what is the focused element.  Searchfield's a bit more complicated than pager because it needs to check that Popupmenus attached to the component don't have focus, as well as all the items inside the searchfield wrapper.
- Notice that in many handlers, `handleSafeBlur()` is called whenever something is clicked out, or defocused, etc.  The component's wrapper will only appear to lose focus and drop the attached event handlers if none of the attached elements are focused.

Hope that helps a bit.

LarryAmisola: Thank you Ed. I'll take a look and study your codes. I like to incorporate that same clean way of coding.

LarryAmisola:
Noted Tim. I'll keep that in mind to add a namespace in an event. Thank you.

tmcconechy:
Closing this for this release - but looking forward to the new solution you can just re-raise when done.

PR #204, opened 2018-06-20
SOHO-8147 - Locale.DateFormat functionality to convert Gregorian date…

boehmd1:
…s on demand

The hotkey for choosing Today did not work if the DatePicker was not open, got Today date in Gregorian instead of Islamic UmAlQura.

Adding optional argument to LocaleFormatDate attribs JSON.
1. fromGregorian
This should be used by clients who store dates in Gregorian.
If the code is trying to format a date that is Gregorian with a Locale using a non Gregorian calendar, then formatDate will first convert the Gregorian date to  Islamic UmAlQura before formatting.
2. toGregorian
This should be used by clients who store dates in Gregorian.
If the code is trying to format a date that is UmAlQura with a Locale using a non Gregorian calendar, then formatDate will first convert the UmAlQura date to  Gregorian before formatting.

SOHO-8147

Two functional test were added to test the use of Locale.formatDate using the directives toGregorian and fromGregorian
'Should convert date from Gregorian (if needed) before formatting date (when fromGregorian is true)'
'Should convert date to Gregorian (if needed) before formatting date (when toGregorian is true)'

Also added a page to visually test the methods. test-formadate-conversion.html

**Closing issues**


app/views/components/locale/test-formatdate-conversion.html
```diff
@@ -0,0 +1,44 @@
+<div class="row">
```
tmcconechy: We probably can remove this visual test if it is covered by a unit test. It doesnt add a whole lot i feel and we are trying to do that to reduce the number of pages to QA.

tmcconechy:
How should i test the "T" key. Tried

http://localhost:4000/components/datepicker/example-index.html?locale=ar-SA
- focus and hit t
But still giving Gregorian.

Are there any other tests i should be doing visually?

tmcconechy:
How should i test the "T" key. Tried

http://localhost:4000/components/datepicker/example-index.html?locale=ar-SA
- focus and hit t
But still giving Gregorian.

Are there any other tests i should be doing visually?

boehmd1:
That seems like the correct steps, at least the steps I user.  Let me try to duplicate that in my sandbox again.  I just merged in changes from your ml.

boehmd1:
Just tested locally, works for me.  I get 1439/10/07

app/views/components/locale/test-formatdate-conversion.html
boehmd1: No disagreement.

tmcconechy: ok just delete this file and then push the delete so its gone out of the pr 👍

PR #206, opened 2018-06-21
Fixed sort indicator was not showing with Datagrid

deep7102:

Column that is specified to be sortable, sort indicators was not showing when the column alignment is set to center.

https://jira.infor.com/browse/SOHO-7444
https://jira.infor.com/browse/SOHO-7918

http://localhost:4000/components/datagrid/example-filter
Open above link
Hover to column header “Status”
See sort indicator should be visible

http://localhost:4000/components/datagrid/test-rowdata-issue.html
Open above link
In "Datagrid B"
Hover to star column header “Default”
See sort indicator should not be visible, and row-height not breaking
Resize and make this column wider by dragging in header
See sort indicator should be visible
Hover to column header “Status”
See sort indicator should be visible

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -4053,7 +4053,6 @@ Datagrid.prototype = {
     this.saveColumns();
     this.saveUserSettings();
     this.headerWidths[idx].width = width;
-    this.setHeaderCenterTextAndSortIndicator();
   },
 
   /**
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -991,6 +991,28 @@ Datagrid.prototype = {
       this.restoreFilter = false;
       this.savedFilter = null;
     }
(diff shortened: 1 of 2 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -991,6 +991,28 @@ Datagrid.prototype = {
...
     }
+    this.setHeaderCenterTextAndSortIndicator();
+  },
+
+  /**
+  * Set header center-text sort-indicator.
+  * @private
+  * @returns {void}
+  */
+  setHeaderCenterTextAndSortIndicator() {
```
tmcconechy: I'd really prefer if we can find a way to do this with CSS only. This could get a bit expensive and not sure it will work if the column is resizing.

Are there any options for that or whats the reason for needing this.

deep7102: Using this to compare header width and text width, yes it will work with column resizing.

PR #262, opened 2018-06-25
SOHO-8128 Added onKeyDown setting

tmcastro:
- Added onKeyDown setting
- Enabled firing of other key down events via onKeyDown setting
- http://localhost:4000/components/datagrid/test-editable-onkeydown.html

- Click on editable cell
- Hit tab button
- There should be a log in console and currently selected editable cell should not be unselected

### DIFF
--- a/app/views/components/datagrid/test-editable-onkeydown.html
+++ b/app/views/components/datagrid/test-editable-onkeydown.html
@@ -1,0 +1,194 @@
+
+<div class="row">
+  <div class="twelve columns">
+    <div role="toolbar" class="toolbar">
+
+      <div class="title">
+        Data Grid Header Title
+        <span class="datagrid-result-count">(N Results)</span>
+      </div>
+
+      <div class="buttonset">
+          <button type="button" id="validate" class="btn-menu">
+            <span>Validate</span>
+          </button>
+          <ul class="popupmenu">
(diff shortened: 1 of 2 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -6138,7 +6138,16 @@ Datagrid.prototype = {
       // Tab, Left and Right arrow keys.
       if ([9, 37, 39].indexOf(key) !== -1) {
         if (key === 9 && !self.settings.actionableMode) {
```
tmcconechy: We probably want to remove this regardless of the actionable mode so maybe this should be

```
if (key === 9 && self.settings.onKeyDown) {
  const ret = self.settings.onKeyDown(e);
  if (ret === false) {
    e.stopPropagation();
    e.preventDefault();
    return;
  }
}

if (key === 9 && !self.settings.actionableMode) {
    return;
}
```

app/views/components/datagrid/test-editable-onkeydown.html
```diff
@@ -0,0 +1,194 @@
+
```
tmcconechy: This has a bit of un necessary code in it since the editable example was copied... I would..

- take the example-index
- add editable: true
- added input editors to the column
- show the onKeyDown setting

That way the examples arent that confusing and show lots of necessary setthings that might get bug reports

PR #319, opened 2018-07-03
SOHO-8056 Fixed issue regarding misasligned dirty tracker

tmcastro:
- Fixed alignment of dirty tracker when switching from Visual to HTML view
- http://localhost:4000/components/editor/example-dirty-tracking.html

- Upon initialization, edit text in editor to display dirty tracker
- Click on 'HTML' button on toolbar
- Dirty tracker should now be aligned correctly


src/components/editor/editor.js
```diff
@@ -301,8 +301,13 @@ Editor.prototype = {
     buttonset += '</div>';
     toolbar += `${buttonset}</div>`;
 
-    this.toolbar = $(toolbar).insertBefore(this.sourceViewActive() ?
-      this.element.prev() : this.element);
+    if (this.element.prev().hasClass('icon-dirty')) {
+      this.toolbar = $(toolbar).insertBefore(this.element.prev().prev());
+      this.element.prev().prev().css({ position: 'absolute' });
```
tmcconechy: Can this be done with a css class?

Also should this be done in the dirty tracking code instead here: https://github.com/infor-design/enterprise/blob/master/src/components/trackdirty/trackdirty.js

Because createToolbar only runs once, i would think this would need to be more dynamic as the dirty indicator is dynamically added / removed so you might have to handle it in there.

tmcastro: Hi Tim,

I've changed my approach for this issue.. instead of setting the position to absolute, I just rearranged the sequence the elements are being added in the editor

PR #324, opened 2018-07-05
soho-8119 contextmenu should be hidden when clicking accordion

### LINKED ISSUE #238: SOHO-8119: Context menu not hidden when clicking on accordion menu
Repro:

1. Setup an accordion menu bar with two menu groups.
2. Setup a context menu to display when clicking on a menu item.
3. Right-click on the menu item (the context-menu is correctly displayed)
4. Click on another menu group in the menu bar (the context menu remains visible).
5. Click somewhere else in the document (the context menu is correctly hidden).

!ContextMenuOnHeader.PNG|thumbnail!

LarryAmisola:
Added a new function attachMenu and variable menuId that will allow the context menu to be display inside the accordion and thru this can control the context menu API.

SOHO-8119
Closes #238 

http://localhost:4000/components/accordion/test-with-contextmenu.html
- visit the link above. once page is displayed. right click on the accordion and context menu should be display properly.
- click on any accordion items. Contextmenu should close.


src/components/accordion/accordion.js
```diff
@@ -361,6 +364,48 @@ Accordion.prototype = {
...
+      * @memberof Tree
+      * @event menuopen
+      * @type {object}
+      * @property {object} event - The jquery event object
+      * @property {object} args for node element, item
+      * @property {HTMLElement} args.menu The menu item
+      * @property {HTMLElement} args.node The DOM Element.
+      */
+      // self.element.triggerHandler('menuopen', { menu: $(`#${menuId}`), node });
+      self.element.triggerHandler('contextmenu', e);
```
tmcconechy: I know this is how datagrid does it. But we ended up with a lot of issues having the code maintain adding the menu and handling the selection events ect.

Can we do this differently such that all we do is trigger the context menu event?

```
self.element.triggerHandler('contextmenu', e);
```

Then in the example code you just listen for that event and manually invoke the context menu fx

```
 ...on('contextmenu', function(e) {
    $(e.target).on('selected', function(args, a) {
      console.log(args, a);
    }).popupmenu({
      menu: menu,
      trigger: 'immediate'
    });
  });
```

This would reduce code and any need for future changes. See https://github.com/infor-design/enterprise/blob/master/app/views/components/popupmenu/example-trigger-immediate.html for immediate menu support.

LarryAmisola:
Thanks Tim for the help. Your solution looks so simple and clean. Actually, I would like to ask about line 1394 return self[`handle${type}Click`](e, element);

How does this code works? Specially this `handle${type}Click`. This is very different from accordion.click right?

tmcconechy:
That code was there. Looks like @EdwardCoyle was trying to make all the click events funnel through that function so its calling a different function depending on the type namely

handleAnchorClick
handleHeaderClick

PR #334, opened 2018-07-10
332 validation fixes

### LINKED ISSUE #332: Validation is Attaching Submit Handlers unnessarily
**Describe the bug**
The validation code is adding an event handler that blocks form submission.

**To Reproduce**
1. Go to https://ustawvqa1.infor.com/efinancials
2. Select batches from the menu
3. Click add
4. Fill in required values and submit

**Expected behavior**
Form should submit.

**Platform**
 - Chrome
 - App specific.

**Additional context**
Its getting into `if (this.element.is('form')) {` unnecessarily in validate.js
Secondarily we should add a line to not validate on `disabled input elements`.

This started in 4.8.0 and was made aware of on a chat with @pilotbob

tmcconechy:

A team noted an issue where forms are no longer submitting. Added an additional check for disabled fields to the validation.

Closes #332  

1. npm run test
1. on sample page http://localhost:4000/components/validation/test-validation-form-manual.html
1. try to submit - should be blocked
1. fill in required fields 
1. try to submit - page should reload/submit

1. on sample page http://localhost:4000/components/validation/test-disable-validation.html
1. hit submit
1. only the top field will validate. the other three have a setting to "skip" validation on them.


app/views/components/datagrid/test-wide-grid-expandable-row-with-form-buttons.html
```diff
@@ -2,7 +2,7 @@
   <div class="row">
     <div class="twelve columns">
 
-      <form class="test-form" data-validate-on="submit" data-validate-form="">
+      <form class="test-form" data-validate-on="submit" data-validate-on="submit">
```
PilotBob: you have the same attribute twice here. Was that intentional?

tmcconechy: No i had it wrong on the attribute that was previously used and changed it last minute.
Will remove the second one we had been using `data-validate-on="submit"` here

tmcconechy:
OK, Good call. Yes i made a better disabled example that shows ALL the ways to disable. Will link it in the .md

PR #346, opened 2018-07-12
Fixing more tooltip on programmatically created more action buttons on toolbar

### LINKED ISSUE #345: Tooltip not appearing on the toolbar's more button when the more button is created programmatically.
**Describe the bug**
The 'More' tooltip isn't showing up on for the more actions button on a toolbar when the button is created programmatically. This is because the button isn't initialized before the button invoking loop.

**To Reproduce**
1. Go to [toolbar/example-index.html](http://480-sohoxi.demo.design.infor.com/components/toolbar/example-index.html)
2. Hover over the  '...' button in the blue header. You should see the more tooltip appear.
3. Hover over the  '...' button in the regular toolbar in the contents area. 
4. No tooltip appears

**Expected behavior**
The more tooltip should appear for any more actions button on a toolbar, no matter if it's created programmatically or in the html.

**Screenshots**
Working in header:
image
since the `<div class="more">` exists:
image

Not working in regular toolbar:
image
since the `<div class="more">` doesn't exist:
image

**Platform**
 - All

Sovia:
Fixing an error where the tooltip "More.." wouldn't appear when the more actions button was created programmatically in toolbar

Closes #345 

1. After pulling in my change, go to /toolbar/example-index.html and hover over the ... button in the toolbar in the content area:
image

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -153,7 +153,6 @@ Toolbar.prototype = {
         }<span class="audible">${Locale.translate('MoreActions')}</span>`)
         .attr('title', Locale.translate('More'))
         .appendTo(moreContainer);
-      this.more = $(this.more).initialize();
     }
 
     // Reference all interactive items in the toolbar
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -151,7 +151,9 @@ Toolbar.prototype = {
       this.more = $('<button class="btn-actions" type="button"></button>')
         .html(`${$.createIcon({ icon: 'more' })
         }<span class="audible">${Locale.translate('MoreActions')}</span>`)
+        .attr('title', Locale.translate('More'))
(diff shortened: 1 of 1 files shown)


src/components/toolbar/toolbar.js
```diff
@@ -151,7 +151,9 @@ Toolbar.prototype = {
       this.more = $('<button class="btn-actions" type="button"></button>')
         .html(`${$.createIcon({ icon: 'more' })
         }<span class="audible">${Locale.translate('MoreActions')}</span>`)
+        .attr('title', Locale.translate('More'))
         .appendTo(moreContainer);
+      this.more = $(this.more).initialize();
```
tmcconechy: Would it work without this line (see section // Invoke buttons). If not maybe just needs `.button().tooltip()` ?

Sovia: Just checked, and it does work without that line. Thanks!

PR #347, opened 2018-07-12
Allow for Populating Dropdowns/Multiselect menu items when filtering

### LINKED ISSUE #267: SOHO-7494: Multiselect using ajax missing expected search term behavior
### Original ticket via @pilotbob: [SOHO-7494](https://jira.infor.com/browse/SOHO-7494)

When implementing the Multilist using ajax there seemed to be no support for a search term. 

You can observe the following when accessing the sample at http://usalvlhlpool1.infor.com/4.3.4/components/multiselect/example-ajax

There appears to be no passing of a typed in search value to the source function or endpoint. Example text implies it will be passed as "term", or the second parameter passed to the source callback function. 

The page implies that a search term is sent with the ajax request but I’m not seeing that.
 
When I tab into the control the first letter I type causes it to drop down, make the ajax request without a term. 
 
Also, once I select one item, I can’t see a way to type in a second item.
 
I am expecting to be able to type in a “new” item and have it make an ajax request to get matching items that they can select from. Is that not possible?
 
I can only make it work by returning the full dataset and then it does searching client side. It also makes the ajax call every time I open the control. I’ve found I can specify the endpoint in data-source as well it calls it every time I open the dropdown.

Unable to type in a second value once an initial value has been typed.

Expected behavior... 

The ability to type into the text box and on change a call is made to the calldback function or to the endpoint (specified in data-source) passing the typed in value. It also would be nice to specify how many letters must be typed before a call to the remote endpoint is made, but that can of course be done in the callback function. 

The ability to type in an additional item. While I was able to type in the first item I was only able to enter multiple items using the mouse and selecting items in the drop down. (not specific to AJAX usage)

The ability to delete a single item, similar to how you can do that in the Jira multiselect. See the affects versions multiselect above. (not specific to AJAX usage)

EdwardCoyle:
This PR adds a mode to the Dropdown component that allows AJAX calls to an external source to populate its dataset each time filtering occurs.  Previously, this wasn't possible (AJAX calls were only allowed to populate when the menu opened).  This new setting also extends to the Multiselect component, and will cause reloading from source to happen whenever the list is filtered, and an option is chosen from the filtered results.

In addition to the new setting, some cleanup of the basic Dropdown interactions is included here:
- Removed typeahead occurrences on `keypress` events, in favor of `input` events.
- Cleaned up filtering codepaths (previously there were several, now there is one).
- _related to the above bug_, there was an issue where the `delay` setting was not being respected during filtering/ajax calls.  This is now always respected in any case where the menu opens or becomes filtered.
- Separated the `keydown` event handlers for the Dropdown/Multiselect's pseudo element and its Search Input element (previously they were tied together and made troubleshooting keyboard issues difficult)
- The legacy setting `reloadSourceOnOpen` was somehow disconnected in this component.  It's been replaced with `reload`, which powers the new typeahead reloading, as well as maintaining the original abilities to reload simply when the menu opens, or never load from an external source at all.
- Added some e2e tests in both Dropdown and Multiselect for the new features, and modified some existing e2e tests to account for the changes in the event handler structure.

Closes #267 (SOHO-7494)

- Pull down the branch
- Run both the functional and e2e tests for Dropdown/Multiselect (all should pass).
- Open http://localhost:4000/components/dropdown/example-index.html and do the following:
  - Open a developer tools console to the Network tab
  - Open the dropdown, and type anything into the filter.
  - Ensure that no AJAX calls are occurring.
- Open http://localhost:4000/components/dropdown/example-ajax.html and do the following:
  - Open a developer tools console to the Network tab
  - Open the dropdown.  It should cause an AJAX call to populate the dropdown list.
  - Type "New" into the search input.  This should NOT cause an AJAX call to populate the list.
- Open http://localhost:4000/components/dropdown/test-reload-typeahead.html and do the following:
  - Open a developer tools console to the Network tab
  - Open the dropdown.  It should cause an AJAX call to populate the dropdown list.
  - Type "New" into the search input.  This should also cause an AJAX call to populate the list.
- Open http://localhost:4000/components/multiselect/test-reload-typeahead.html and do the following:
  - Open a developer tools console to the Network tab
  - Open the multiselect.  It should cause an AJAX call to populate the dropdown list.
  - Type "New" into the search input.  This should also cause an AJAX call to populate the list.
  - Select "New Jersey" from the list.  After the selection happens, the list should reset.  Another AJAX call should populate the list.

**Additional items**:

Paging @PilotBob


PilotBob:
Something I don't see your test cover. 

If there are selected values in the drop down, they still show as selected even though the AJAX data may not contain those items since the term doesn't include them?

Is it possible to set it to only pull data from the server if they type a term in?

EdwardCoyle:
> If there are selected values in the drop down, they still show as selected even though the AJAX data may not contain those items since the term doesn't include them?

Right -- in the typeahead-reloading mode, it stores selected values this way since it would theoretically be possible to search for one item (say "New Jersey"), select it, then search for another with a completely different term (say "Colorado").  If you've made a previous selection, I'd think it should be retained.

> Is it possible to set it to only pull data from the server if they type a term in?

It should actually be doing this already, if you focus the dropdown's pseudo (in-page) element, and type the search term.  It will bypass the full results set in favor of the filtered one.

tmcconechy:
ok just saw one thing so far, i ran the test and smoke tested a bunch of examples.

http://localhost:4000/components/dropdown/example-index.html
- click the label
- type new
Now the word is all selected making it so you cant type anymore. This didnt do that before.

PR #362, opened 2018-07-16
Clear searchfield when new data is loaded

### LINKED ISSUE #307: Clear Searchfield when new data is loaded
I have a datagrid where I can dynamically change dataset and if I have something in the search field the result will be 0 when I load a new dataset. (Even if I have a hit on that search string).

I would support having the datagrid just clear it when you call updateDataset.

A clear-method on the toolbar API.

tmcastro:
- Cleared searchfield on toolbar upon firing updateDataset function
- http://localhost:4000/components/datagrid/test-clear-searchfield-on-update-dataset.html

**Related Issues**
Closes #307 

- Type in search keyword on searchfield on toolbar and hit Enter key
- Click on "Load Data Set"
- Searchfield should be cleared and datagrid should not be filtered


app/views/components/datagrid/test-clear-searchfield-on-update-dataset.html
```diff
@@ -0,0 +1,69 @@
...
+      }, {
+        activity: 'Inspect and Repair',
+        quantity: 6
+      }, {
+        activity: 'Inspect and Repair',
+        quantity: 7
+      }
+    ];
+
+    // $('#datagrid1').datagrid({ columns, dataset });
```
tmcconechy: Could remove

app/views/components/datagrid/test-updatedataset.html
```diff
@@ -5,6 +5,9 @@
   <div class="twelve columns">
     <div id="datagrid2"></div>
   </div>
+  <div class="twelve columns">
+    <button class='btn-primary' id='load-dataset'>Load Data Set</button>
```
tmcconechy: Use secondary button or toolbar buttons in examples.

tmcconechy: Also would one example/test page be enough to cover this instead of two?

src/components/searchfield/searchfield.js
```diff
@@ -1007,6 +1007,10 @@ SearchField.prototype = {
       e.preventDefault();
     }
 
+    if (e.ctrlKey && key === 8) {
```
tmcconechy: Not sure that this part is needed? the key here is alt + delete (mac) or ctrl + delete (windows)
Backspace should just behave normally and clear char by char.

tmcconechy:
screen shot 2018-07-16 at 9 29 40 am

src/components/searchfield/searchfield.js
EdwardCoyle: @tmcastro we merged your other PR into `master`, so if you simply remove this code and run a `git pull origin master` on this branch, you should be set.

Issue #270, opened 2018-06-25
Unselectable item with badge in the Dropdown


CindyMercadoReyes:
**Describe the bug**
Unselectable item with badge in the Dropdown.

**To Reproduce**
1. Go to url: http://480-beta0-sohoxi.demo.design.infor.com/components/dropdown/test-badges.html
2. Click the Dropdown to open
3. Try to select the last item with badge but doesn't have text'
4. See screen shot
<img width="1279" alt="dropdown badges" src="https://user-images.githubusercontent.com/39101346/41868542-59ae9c92-7884-11e8-9656-9f936b16eeaa.png">
<img width="1279" alt="dropdown badges" src="https://user-images.githubusercontent.com/39101346/41868548-5d5e6be2-7884-11e8-86a4-42e32adf8479.png">

**Expected behavior**
badge should appear.

**Screenshots**

**Platform**
 -
 - All Browsers 
 
**Additional context**

brianjuan:
Bug fix is working as expected across all browsers and passed QA testing.

Environment: http://490-beta0-sohoxi.demo.design.infor.com/components/dropdown/test-badges.html
<img width="612" alt="screen shot 2018-07-19 at 4 52 44 pm" src="https://user-images.githubusercontent.com/39508864/42932258-684bd7b2-8b74-11e8-861e-8335164c922d.png">

tmcconechy:
This issue should have been closed.

PR #369, opened 2018-07-17
270 fix unselectable item in dropdown

LarryAmisola:

change the dropdown anchor css display property to inline so that badge will be align side by side with the anchor.

#270

http://localhost:4000/components/dropdown/test-badges.html
Check the link above, click the dropdown list. Label and badge should be align to each other.


tmcconechy:
screen shot 2018-07-17 at 8 04 15 am

app/views/components/dropdown/test-badges.html
```diff
@@ -4,7 +4,7 @@
 
     <div class="field">
       <label for="badge-test">Badge Test</label>
-      <select id="badge-test" name="badge-test" class="dropdown">
+      <select id="badge-test" name="badge-test" class="dropdown" data-options="{'cssClass': 'with-badge'}">
```
LarryAmisola: This will make the fix more specific for the badge. This will allow us to display the badge side by side with the anchor without affecting the other examples.

src/components/dropdown/_dropdown.scss
```diff
@@ -298,6 +304,8 @@ div.multiselect {
 
       .badge {
         border-color: $inverse-color;
+        background-color: $inverse-color;
```
LarryAmisola: Change the background to white on hover

tmcconechy:
Pushed a change that just moves the badge inside the span. This is simpler and doesnt require someone to update their code and add the setting.

PR #480, opened 2018-07-23
424 Prevented expanding of row templates on search

### LINKED ISSUE #424: SOHO-7396: Row templates expanded when searching in Datagrid
Variation of SHUB-6744  - All rows expanded in datagrid when keywordFilter blanked.

In this case the row template is expanded for matching rows when a match is found in the template. The example used mentioned in SHUB-6744 cannot be used to illustrate the issue as it uses empty templates.

Using this example to illustrate the issue:
http://usalvlhlpool1.infor.com/4.3.4/components/datagrid/example-expandable-row.html

!row_expanded_template_text.png|thumbnail!

Notice that a search for 'Lo' finds a row and expands the template as it matches 'Lorem'.

Empty templates can be used as a workaround for this issue.

tmcastro:
- Prevented expanding of row templates on data grid search
- Updated validation for hiding data grid rows
- http://localhost:4000/components/datagrid/example-expandable-row.html

- Open search filter on upper right side of data grid
- Use 'lo' as keyword and hit Enter
- Row template should no longer be expanded

Closes #424

### DIFF
--- a/app/views/components/datagrid/test-expand-row-on-search.html
+++ b/app/views/components/datagrid/test-expand-row-on-search.html
@@ -1,0 +1,61 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+{{={{{ }}}=}}
+
+<script>
+  $('body').one('initialized', function () {
+
+      var grid,
+        columns = [],
+        data = [];
+
+      // Some Sample Data
+      data.push({ id: 1, productId: 2142201, sku: '<b>SKU #9000001-237</b>', productName: 'Compressor', activity:  'Assemble Paint', quantity: 1, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action'});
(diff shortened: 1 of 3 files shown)


tmcconechy:
Can we make this into a datagrid option? That was mentioned on the issue. Maybe just add a 

expandRowOnSearch: true / false

Option on top of this (default can still be true). Then we will need a specific test page for the feature showing it and a e2e test for that page.

tmcconechy:
This looks good but im thinking some more about it and i think when the option is set it should not even attempt to highlight the expandable area. So maybe.

1. Change option name to  searchExpandableRow: true
2. Make it not highlight the text or use the expandable row in the search at all when thats set. (what you see now is if you search and open the row it is still highlighted. And it may be using the values to display rows that have data in the expandable row when the developer asked for that to not be the case).

tmcastro:
Hi @tmcconechy, should I also prevent showing of expandable row upon searching?

tmcconechy:
No, i think it should still show the expandable row with the row it finds... So.

searchExpandableRow: true
- behaves like now, it will highlight text in the expandable row. If no text is found in the expandable row it still shows the expandable row along with the row thats found

searchExpandableRow: false
- it will never highlight text in the expandable row.  it still shows the expandable row along with the row thats found

This will be more work but its really what the bug is about. The expandable row should stay with the rows found and we have an option to include it in the search or not

PR #523, opened 2018-07-31
371 Fixed issue regarding incorrect focusing of cells when clicking a…

### LINKED ISSUE #371: Datagrid- Nested Grid : focus still on the parent rows
**Description**
When you select a link / row within the nested grid, the focus is still on the parent link/row. As a result, the click event within the nested grid doesn’t work as expected. 

**To Reproduce**
1. http://master-enterprise.demo.design.infor.com/components/datagrid/example-nested-grids.html
2. Expand any row
3. click directly on the link i the subgrid

Result: The parent link will focus, and the event info on the column click event should be retested.

**Expected behavior**
Each row within the nested datagrid should trigger the corresponding 'click' event .

**Platform**
 - Browsers used: chrome, safari
 - SOHO version: 4.8.0

tmcastro:
- Fixed issue regarding incorrect focusing of cell when clicking 'a' tags on nested grids
- Added a new 'targetLevel' parameter for setActiveCell and cellNodes functions
- Issue was caused because more than 1 row node was being found when searching for aria-rowindex with a specific index
- http://localhost:4000/components/datagrid/example-nested-grids.html

Closes #371 

- Expand 4th row, click on any of the links inside nested grid
- Before, the cells on the outer grid is being focused instead of the clicked cells
- Now, the cells containing the clicked links should now be selected

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -4390,7 +4390,7 @@ Datagrid.prototype = {
    * @param  {boolean} includeGroups If true groups are taken into account.
    * @returns {object} The dom node
    */
-  cellNode(row, cell, includeGroups) {
+  cellNode(row, cell, includeGroups, targetLevel) {
     let cells = null;
     let rowNode = null;
 
@@ -4411,7 +4411,7 @@ Datagrid.prototype = {
       return $();
     }
 
-    cells = rowNode.find('td');
+    cells = $(rowNode[targetLevel]).find('td');
     return cells.eq(cell >= cells.length ? cells.length - 1 : cell);
   },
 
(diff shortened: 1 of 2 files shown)


tmcastro:
Hi @tmcconechy, I've tested your suggested code but I'm still seeing some issues. I've attached a video of what I'm getting. I've also added to the code I've pushed yesterday to fix the issue regarding not being able to click on other cells on the nested grid.

[ISSUE-371.zip](https://github.com/infor-design/enterprise/files/2248114/ISSUE-371.zip)

tmcconechy:
@tmcastro I pushed a fix on a new branch on 371-datagrid-nested-link-click maybe just test it out some more there.... If you agree then close this PR and make a new one from that.

Issue #354, opened 2018-07-13
Refactor old sass vars with corresponding ids-identity sass vars


clepore:
**Is your feature request related to a problem? Please describe.**
Currently we have a temporary implementation of ids-identity tokens where we map them to their existing sass variable counter part, for example in [_config.scss#165](https://github.com/infor-design/enterprise/blob/master/src/core/_config.scss#L160):

```sass
// existing sass var : ids-identity token
$primary-btn-bg-color: $button-primary-bg-color;
```

**Describe the solution you'd like**
The goal is to replace the older variables with their corresponding token sass variable. This will help cleanup `_config.scss` and remove confusion from having variables tied to one another with similar names.

To continue with the above example:

1. Remove [_config.scss#L160](https://github.com/infor-design/enterprise/blob/master/src/core/_config.scss#L160)
1. Find all instances of the original variable `$primary-btn-bg-color`
1. Replace those instances with the ids-identity token variable `$button-primary-bg-color`.
1. Repeat for each "mapping" pair to clean up redundancy

*Note that it might be easier to break this ticket into a few pull requests for smaller, easier peer reviews.*

**Additional context**
See https://github.com/infor-design/design-system/issues/158 for ids-identity and token related information.

brianjuan:
Hi, can you provide us the page for QA testing and also the necessary steps. Thanks!

clepore:
This is going to be a hard one to test. Since we mostly got rid of duplicate style variables, the testing method would be to make sure the site styles didn’t change. The easiest way to do this would be to test everything you normally would and keep an eye out for any differences (mostly colors) that look out of place.

PR #530, opened 2018-07-31
Refactor old vars with corresponding tokens (buttons)

jmacaluso711:
- Replaces old vars with ids-identity vars. 
- Removed unneeded vars from config and used ids-identity vars in component specific `scss` files. 
- This PR only relates to buttons. (primary, default and link)

related #354

- `npm install` to get `ids-identity@v1.2.0`

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -116,35 +116,35 @@ button {
 
 // standard, primary, destructive
 .btn-primary {
-  background-color: $primary-btn-bg-color;
-  border: 1px solid $primary-btn-bg-color;
-  color: $primary-btn-color;
+  background-color: $button-primary-bg-color;
+  border: 1px solid $button-primary-bg-color;
+  color: $button-primary-text-color;
 
   &.btn {
-    color: $primary-btn-color;
+    color: $button-primary-text-color;
   }
 
   .icon {
-    fill: $primary-btn-color;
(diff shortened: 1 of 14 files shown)


src/components/button/_button.scss
```diff
@@ -116,59 +116,59 @@ button {
...
-  background-color: $primary-btn-bg-color;
-  border: 1px solid $primary-btn-bg-color;
-  color: $primary-btn-color;
+  background-color: $button-primary-bg-color;
+  border: 1px solid $button-primary-bg-color;
+  color: $button-primary-text-color;
 
   &.btn {
-    color: $primary-btn-color;
+    color: $button-primary-text-color;
```
EdwardCoyle: Not sure exactly what's off here... I think it's on buttons that contain both a `.btn` and `.btn-primary` class.  The focus state is not the solid-border, and is instead using a box shadow:

screen shot 2018-08-01 at 12 36 29 pm

jmacaluso711: is it buttons that don't have the `hide-focus` class?
```
&:focus {
    border: 1px solid transparent;

    &:not(.hide-focus) {
      box-shadow: 0 0 0 2px $body-bg-color, 0 0 0 3px $button-primary-bg-color, $focus-box-shadow;
    }
```

PR #515, opened 2018-07-30
410 selection icons in inverse popupmenu too low contrast in high contrast theme

ericangeles:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

I added a variable ('$popupmenu-checked-color-light') and reuse the variable ('$azure06'), and I also add a style where it will target the high-contrast-theme as a parent and will get the selector '::before' and override the border-color of it.

```
/* Added it in Popop Menu declaration */
$popupmenu-checked-color-light: $azure06;

/* Located at the bottom of the file */
.high-contrast-theme {
  .masthead {
    .popupmenu {
      li {
        &.is-checked > a::before {
          border-color: $popupmenu-checked-color-light $popupmenu-checked-color-light $popupmenu-checked-color-light transparent;
        }
      }
    }
  }
}
```
**Related github/jira issue (required)**:

Closes #410

**Steps necessary to review your pull request (required)**:

- Run http://localhost:4000/components/masthead/test-different-button-types?theme=high-contrast
- Select a value from the status dropdown
- note the check...

Issue #410, opened 2018-07-17
SOHO-7788: Selection icons in inverse popupmenu are too low contrast in high contrast theme


clepore:
The blue selection icons -- particularily single-select -- of an inverse popupmenu e.g. in the masthead are barely legible when using high contrast theme.

See https://jira.infor.com/browse/SOHO-7788 for screen shots and more info.

brianjuan:
http://master-enterprise.demo.design.infor.com/components/masthead/test-different-button-types
Passed QA testing

PR #535, opened 2018-08-01
410 selection icons in inverse

ericangeles:

Fixes check icon color specifically in high contrast theme but it is also apply to the rest of themes. Added a new variable ```$popupmenu-checked-color-inverse: $azure06;``` in _masthead.scss and created a new css rule to target the check icon specific only in the masthead section. I noticed that the css rule in the masthead was also used in the header menu section.

```
.popupmenu {
      li {
        &.is-checked > a::before {
          border-color: $popupmenu-checked-color-inverse $popupmenu-checked-color-inverse $popupmenu-checked-color-inverse transparent;
        }
      }
    }
```
_This was wrapped in the masthead class_

Closes https://github.com/infor-design/enterprise/pull/515 - deleted / original PR. Renamed the remote branch to SOHO-7788-selection-icons-in-inverse
Closes https://github.com/infor-design/enterprise/issues/410

- Run http://localhost:4000/components/masthead/test-different-button-types
- Select a value from the status dropdown
- note the check...


tmcconechy:
Did you see my suggestion on the other PR?

Maybe a better fix:
- add a new variable $popupmenu-checked-color-inverse to all three themes
- use that in a rule in masthead.scss 
- color would be the same in 2 / 3 themes

This way we have a defined variable for the inverse check color thats not just in one file.

src/components/masthead/_masthead.scss
```diff
@@ -7,6 +7,7 @@ $masthead-text-color: $slate03;
 $masthead-button-color: $slate04;
 $masthead-button-hover-color: $white;
 $masthead-button-focus-color: $white;
+$popupmenu-checked-color-inverse: $azure06;
```
tmcconechy: Might not need this?

PR #539, opened 2018-08-01
Datagrid inadvertenly showing dirty flag for dropdowns

### LINKED ISSUE #538: Datagrid: With showDirty displaying row dirty when dropdown value hasn't changed
A row's status is showing the dirty flag, when a cell has a dropdown and the user selects the same value from the dropdown.  Landmark issue, since the column.field property will be set to something like "fields.ReasonType.value".  So within the updateCellNode function, the code for getting the oldVal needs to be changed.

**Expected behavior**
Dirty flag is not set when the user opens the dropdown and selects the current value from the drop-down list.
withfix
withoutfix

**Screenshots**

**Platform**
 - Device (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name [e.g. chrome, safari, stock browser]
 - Browser Version [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**
Made the following change to the method updateCellNode() within datagrid.js
```
    // const oldVal = (col.field ? rowData[col.field] : '');
    let oldVal = '';
    if (col.field) {
      if (col.field.indexOf('.') > -1) {
        var parts = col.field.split('.');
        if (parts.length === 2) {
          oldVal = rowData[parts[0]][parts[1]];
        }

        if (parts.length === 3) {
          oldVal = rowData[parts[0]][parts[1]][parts[2]];
        }
      } else {
        oldVal = rowData[col.field];
      }
    }

```

claudenbach:
A row's status is showing the dirty flag, when a cell has a dropdown and the user selects the same value from the dropdown. Landmark issue, since the column.field property will be set to something like "fields.ReasonType.value". So within the updateCellNode function, the code for getting the oldVal needs to be changed.

Closes #538 "

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7119,7 +7119,22 @@ Datagrid.prototype = {
       rowNode = this.visualRowNode(row);
       cellNode = rowNode.find('td').eq(cell);
     }
-    const oldVal = (col.field ? rowData[col.field] : '');
+    // const oldVal = (col.field ? rowData[col.field] : '');
+    let oldVal = '';
+    if (col.field) {
+      if (col.field.indexOf('.') > -1) {
+        var parts = col.field.split('.');
+        if (parts.length === 2) {
+          oldVal = rowData[parts[0]][parts[1]];
+        }
+
+        if (parts.length === 3) {
+          oldVal = rowData[parts[0]][parts[1]][parts[2]];
+        }
+      } else {
+        oldVal = rowData[col.field];
+      }
+    }
 
     // Coerce/Serialize value if from cell edit
     if (!fromApiCall) {
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7119,22 +7119,7 @@ Datagrid.prototype = {
       rowNode = this.visualRowNode(row);
       cellNode = rowNode.find('td').eq(cell);
     }
-    // const oldVal = (col.field ? rowData[col.field] : '');
-    let oldVal = '';
-    if (col.field) {
-      if (col.field.indexOf('.') > -1) {
-        var parts = col.field.split('.');
-        if (parts.length === 2) {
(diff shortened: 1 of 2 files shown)


claudenbach:
Thanks, I'll see if simply calling that method works.

claudenbach:
Yes, changing the statement works great.  A nice solution.
 old:
`const oldVal = (col.field ? rowData[col.field] : '');`

new:
`const oldVal = this.fieldValue(rowData, col.field);`

tmcconechy:
OK perfect, then you can just change that and then commit && push to update the PR

tmcconechy:
(from phillip)

To fix up the crlf line feeds in your soho components.
 
git config core.autocrlf false
git rm --cached -r .
git reset --hard

(Basically we use LF but windows likes CRLF and changes files on git checkout)

claudenbach:
I've updated the pull request.

When I run the tests, several fail but they aren't related to my change since they don't even hit my code change?

tmcconechy:
We have some tests failing that need a merge from 4.9 branch. If we rerun the build it will sometimes be ok. Lets see when it runs.

tmcconechy:
@claudenbach to fix the test can you change it to

```
  it('Should show currently dirty rows', (done) => {
    datagridObj.rowStatus(1, 'dirty');
    datagridObj.rowStatus(3, 'dirty');

    setTimeout(() => {
      const dirtyRows = datagridObj.dirtyRows();

      expect(dirtyRows.length).toEqual(2);
      done();
    }, 0);
  });
```

In datagrid-validation.func-spec.js .. I tried to but i cant update your master. Just replace the test with that and it should work.

Not entirely sure why it is a problem just on this PR tho.

claudenbach:
What I figured out is that the two items added to the output array, have a percent and phone property that does not exist in the sample data.  And the price value in the output array for item 4 is set to ' ' (empty string) but in the sample data its value is a null. 

But I like your solution, so the sample data may change without impacting the test.

Issue #311, opened 2018-07-02
Colorpicker hides color names for certain languages


gushanson:
**Describe the bug**
The colorpicker is only optimized for en-US. When using for example fi-FI the text is cropped.
I suggest that the colorpicker should get enough width to display its longest color name (if it fits within the container).

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/colorpicker/example-show-label.html?locale=fi-FI
2. Click on the colorpicker menu
3. Click a the first green color.

**Expected behavior**
The color name "Smaragdinvihreä10" is visible.

**Screenshots**
Only "Smaragdinvi" is visible.

**Platform**
 - Windows 10
- Chrome, IE11, Edge

**Additional context**

tmcconechy:
Whoever takes this. We should make the colorpicker take these sizes:

http://localhost:4000/components/input/example-sizes.html

Particularily input-mm and input-md -> if those classes are tagged in should go to that size.
Possibly consider an auto size (input-auto)

ericangeles:
I also noticed that there is a span container (with a class of ```.colorpicker-container```) in the input field that sets the width to just ```150px``` only. Maybe we change it to min-width instead of width.

PR #544, opened 2018-08-02
311 - Fixed colorpicker text being cropped

ericangeles:

Created a new template for colorpicker sizes. **app/views/components/colorpicker/example-sizes.html**
Changed the ```width``` to ```min-width``` instead in the span container of input field so that the width of the input field will not look odd. Seems that the text will go beneath the swatch.

Closes https://github.com/infor-design/enterprise/issues/311

- Run http://localhost:4000/components/colorpicker/example-show-label.html?locale=fi-FI
- You will see different sizes of colorpicker

**File Changes**
- **Modified:** app/views/components/colorpicker/example-show-label.html - revert back to original structure.
- **Modified:** src/components/colorpicker/_colorpicker.scss
- **New File:** app/views/components/colorpicker/example-sizes.html
- **Modified:** src/components/colorpicker/readme.md


app/views/components/colorpicker/example-show-label.html
```diff
@@ -3,7 +3,7 @@
 
     <div class="field">
       <label for="show-label">Show Label</label>
-      <input class="colorpicker" id="show-label" type="text" value="#0E5B52"
+      <input class="colorpicker input-mm input-md" id="show-label" type="text" value="#0E5B52"
```
tmcconechy: input-mm and input-md are two different sizes so i would expect this to have one or the other.
i suggest making a new example in colorpicker/example-sizes.html that shows all the sizes. Like this but for color picker.

http://master-enterprise.demo.design.infor.com/components/datepicker/example-sizes.html

Also remember to add the link to the colorpicker.md (slug section)

src/components/colorpicker/_colorpicker.scss
```diff
@@ -22,7 +22,7 @@
   border-radius: 2px;
   display: inline-block;
   overflow: hidden;
-  width: 150px;
+  min-width: 90px;
```
tmcconechy: I think we need to remove this min-width as this broke the "just color" example.
http://localhost:4000/components/colorpicker/example-short-field.html

src/components/colorpicker/_colorpicker.scss
```diff
@@ -78,6 +78,8 @@
       input[readonly] {
         background-color: $input-bg-color;
         border-color: $input-border-color;
+        color: $graphite10;
```
tmcconechy: Use `$input-color` here for a variable..

Run http://localhost:4000/components/colorpicker/example-sizes.html in dark theme you cannot see the text. So we need to use a variable thats set for all 3 themes.

Issue #245, opened 2018-06-22
SOHO-7978: Right click (long press) for context menus on mobile


clepore:
Need a way to open context menus on mobile devices. On lists / datagrid and inputs.

EdwardCoyle:
@tmcconechy think this behavior should be limited to touch events only?  Or should the "long press" also translate to held-down mouse buttons?

tmcconechy:
@EdwardCoyle would say touch device only. We don't need it because on a desktop you would hover. So its swapping non existant hover functionality for touch longpress.

EdwardCoyle:
That's actually a whole other issue then... we should probably have a way for field options to be permanently accessible in a touch environment

EdwardCoyle:
Noting that work on #400 can go ahead now that #504 is merged

EdwardCoyle:
Awaiting the merge of #573 to put together an example of the context menu being replaced by a Field Options button

EdwardCoyle:
Ready for Q/A.  _To test:_ 

- open http://localhost:4000/components/contextmenu/example-field-options.html on a mobile device (iOS Safari/Android Chrome)
- all examples should show a working context menu attached to each field options button

brianjuan:
http://master-enterprise.demo.design.infor.com/components/contextmenu/example-field-options.html

Contextmenu is cropped and not scrollable in iOS devices, contextmenu will close if you attempt to scroll it.
Another issue is the focus of the previously selected field is not disappearing.
<img width="528" alt="screen shot 2018-08-16 at 4 22 43 pm" src="https://user-images.githubusercontent.com/39508864/44197701-fba3ea00-a171-11e8-8f9e-c625c2deaecf.png">

EdwardCoyle:
@brianjuan, after some digging, the scrolling in the Popupmenu on iOS may require a larger fix than we have time for.  I've raised #645 for that one.  As far as the scrolling goes, please disregard that for this issue.

I've got a PR coming that fixes the active state issues.

EdwardCoyle:
@brianjuan, #646 should have fixed the active state issues.  Please retest against the latest cut of `master`.  #645 contains all the popupmenu scrolling items, and has been spun off as a separate issue.

brianjuan:
@EdwardCoyle Thanks for the update, I've tested the active state issue and now passed QA testing, will move this to Done.

PR #573, opened 2018-08-07
555 - Permanently enable the Field Options button's visibility on touch-based devices

### LINKED ISSUE #555: Enable permanently-displayed field options button in touch environments
The user experience surrounding the field options button on input fields doesn't currently work all that well (as of `v4.9.0`) in iOS environments, and doesn't work at all in Android environments.

To test, open http://master-enterprise.demo.design.infor.com/components/field-options/example-index.html in a mobile browser.

_In iOS Safari:_
- Field Options are not accessible unless the field is previously focused, and the button doesn't always stay in-tact when navigating away from the input field.

_In Android Chrome:_
- Field options are completely inaccessible on dropdown/multiselect components
- On most input fields, field options are visible once the input becomes focused, but immediately disappear and close their menus when the button is tapped.
- On checks/radios, interestingly enough everything works fine.

- Field Options attached to other components should always be visible on the page and ready for interactions when a touch environment is detected (the Environment core component has a check for this as of `v4.10.x` that can be used)

**Additional context**
For an example of why we might want this, see the discussion about touch-enabled context menus on input fields in #245

EdwardCoyle:
Previously, mobile devices had difficulty using Field Options buttons due to their dependency on hover and focus events.  This PR sets up a touch-device mode, which permanently displays the field options button in environments where a touch screen is present.

Closes #555

Pull this branch and run the demoapp, then:

_In a desktop browser:_
- open http://localhost:4000/components/field-options/example-index
- hover each field and ensure that the field options button for each is displayed
- focus each field and ensure that the field options button for each is displayed

_In a mobile browser:_
- open http://localhost:4000/components/field-options/example-index
- ensure that the field options button is permanently displayed next to each field, and can be accessed with a tap.

### DIFF
--- a/app/views/components/field-options/example-index.html
+++ b/app/views/components/field-options/example-index.html
@@ -410,9 +410,6 @@
         fullWidth: true
       }
     }
-  }).on('change', function (e, args) {
-    console.log(args[0]);
-  })
-  .fieldoptions();
+  }).fieldoptions();
 
 </script>
--- a/src/components/field-options/_field-options.scss
+++ b/src/components/field-options/_field-options.scss
@@ -37,7 +37,7 @@
 }
 
 .field-options {
-  max-width: $field-md;
+  max-width: ($field-md - 40px);
   width: calc(100% - 40px);
 
   &.input-xs {
@@ -219,6 +219,11 @@
     }
   }
 
+  .dropdown,
+  .multiselect {
+    max-width: ($field-md - 40px);
+  }
+
   .searchfield-wrapper {
     box-shadow: none;
 
@@ -233,6 +238,14 @@
     .btn-actions {
       left: -8px;
       width: 36px;
+    }
+  }
+
+  .textarea {
+    width: ($field-md - 40px);
+
+    ~ .btn-actions {
+      left: auto;
     }
   }
 }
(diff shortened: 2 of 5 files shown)


EdwardCoyle:
@davidcarlsonberg, just pushed... changed `mouseenter` and `mouseleave` to `mouseover` and `mouseout` to fix those issues with Timepicker, Fileupload, Lookup, and close buttons.  These events bubble, so it helped clean up the lifecycle a bit while giving us the ability to delegate.

I think the fact that field options doesn't display when hovering a disabled item is by design (the logic being: no extra commands should be available if a component's primary action is disabled).  I added some specific logic for this into the `.field`'s event handler for dealing with visibility changes.

EdwardCoyle:
Noting for visibility that fixing #245 depends on this being merged

tmcconechy:
screen shot 2018-08-08 at 8 46 14 am

EdwardCoyle:
@tmcconechy the intention is that it's always visible on touch screen devices.  I'm not sure that it's going to translate perfectly in the Chrome dev tools.

EdwardCoyle:
Yeah, we don't really have any other precedent for disabling hover state-related visuals on elements that don't have focus, although I could see the argument for it.  Let's leave it for now

PR #576, opened 2018-08-07
fix for indeterminate paging not calling previous sometimes

### LINKED ISSUE #567: datagrid indeterminate paging - previous page fails if initially loaded data is in the middle of the set of records
**Describe the bug**
There is an case where the source(req.type === prev) doesn't get called.  If you load a page of records that isn't the first page (req.firstPage/lastPage both equal false) and click previous the source function is never called. 

**To Reproduce**
1. save example on next comment as example-previssue.html to your project.
2. open example-previssue.html in your browser
3. notice a page of recs is displayed (to illustrate loading a page of recs from the middle of a dataset).
4. click previous page and notice request type is never called on the datagrid source function.

**Expected behavior**
The source(req.type === 'prev') should be called in this case.

**Platform**
 - OS Version: Windows 10
 - Browser Name: chrome
 - Latest build

**Additional context**
This is only an issue with indeterminate paging AFAIK

pwpatton:
Issue: When paging is indeterminate prev page was never called when starting a datagrid mid page
What I did: Removed the drawing optimization for indeterminate paging since pagNum can't really change when paging is indeterminate, it isn't relevant to indeterminate paging. 

Closes #567

See git hub issue for these details.

### DIFF
--- a/src/components/pager/pager.js
+++ b/src/components/pager/pager.js
@@ -236,22 +236,40 @@ Pager.prototype = {
       }
 
       if (li.is('.pager-prev')) {
-        self.setActivePage(self.activePage - 1, false, 'prev');
+        if (self.settings.indeterminate) {
+          self.activePage = -1; // so setActivePage doesn't skip rendering the row data
+          self.setActivePage(1, false, 'prev');
+        } else {
+          self.setActivePage(self.activePage - 1, false, 'prev');
+        }
         return false;
       }
 
       if (li.is('.pager-next')) {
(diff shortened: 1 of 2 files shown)


pwpatton:
Ok, I've checked in a jasmine test for this too. Ready to be reviewed.

tmcconechy:
How do i test this on the page? (Page mentioned isnt there but i assume can test on the indeterminate paging example).

or is it all covered by the e2e test?

pwpatton:
It's definitely converted by the jasmine/unit test.  I made a simple sample but didn't check it in as it was obscure and the test covers it anyway.

PR #585, opened 2018-08-09
453 - SOHO-6041 show tooltips on disabled inputs and buttons

### LINKED ISSUE #453: SOHO-6041: Show tooltips on disabled inputs and buttons.
**Summary**

When an {{input}} or {{button}} is disabled, it's mouse hover tooltip is not shown.

**Expected behavior**

As standard HTML does show title regardless of whether an input is disabled or enabled, I'd expect IDS to do the same. It is actually pretty useful to allow user to hover over a disabled field and give him feedback why is it disabled.

Just create an HTML page with this input:

```
<input type="text" disabled="disabled" title="I'm title"></input>
```

Without IDS, it shows the title. With IDS initialized, the title is removed at runtime (by the `Tooltip` component) and is not shown.

**Possible solution**

Enable tooltips to display on these elements whenever `settings.hover === true`, similar to what's been done for icons in #554

davidcarlsonberg:
Disabled elements are not allowed to handle events in most browsers. A hover event to display element title is desired on disabled inputs and buttons.

Closes #453 

- Pull branch and run app.
- Visit example page to verify title renders as tooltip: http://localhost:4000/components/button/example-disabled-button-tooltip
- Review documentation for further instruction or documentation on purpose of / reasoning behind this workaround: ./src/components/button/readme.md

### DIFF
--- a/app/views/components/button/example-disabled-button-tooltip.html
+++ b/app/views/components/button/example-disabled-button-tooltip.html
@@ -1,0 +1,17 @@
+<div class="row">
+
+  <div class="twelve columns">
+
+    <div class="field">
+      <button class="btn-secondary" type="button" disabled><div title="div title - tooltip">Button</div></button>
+    </div>
+
+    </div>
+
+  </div>
+
+</div>
+
+<script type="text/javascript">
+  $('div[title]').tooltip();
+</script>
--- a/app/views/components/button/example-disabled-button-tooltip.html
+++ b/app/views/components/button/example-disabled-button-tooltip.html
@@ -1,16 +1,108 @@
+
 <div class="row">
+  <div class="six columns">
 
-  <div class="twelve columns">
+    <span class="label">Primary</span>
 
-    <div class="field">
-      <button class="btn-secondary" type="button" disabled><div title="div title - tooltip">Button</div></button>
-    </div>
+    <button class="btn-primary" type="button" title="btn-primary title/tooltip" disabled><div class="disabled-tooltip" title="btn-primary title/tooltip"></div>Action</button><br><br><br>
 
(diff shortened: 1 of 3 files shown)


app/views/components/button/example-disabled-button-tooltip.html
```diff
@@ -0,0 +1,17 @@
+<div class="row">
+
+  <div class="twelve columns">
+
+    <div class="field">
+      <button class="btn-secondary" type="button" disabled><div title="div title - tooltip">Button</div></button>
+    </div>
+
+    </div>
```
tmcconechy: Extra Div?

app/views/components/button/example-disabled-button-tooltip.html
```diff
@@ -0,0 +1,17 @@
+<div class="row">
+
+  <div class="twelve columns">
+
+    <div class="field">
```
tmcconechy: Field may not be needed here

app/views/components/button/example-disabled-button-tooltip.html
```diff
@@ -0,0 +1,17 @@
...
+      <button class="btn-secondary" type="button" disabled><div title="div title - tooltip">Button</div></button>
```
tmcconechy: Can we try to add more button types and see what works? fx icon, menu, primary, link ect..
I think the icon might not work

davidcarlsonberg: The icon doesn't work. The `div` screws with the styling. Fussing with styling now on div to preserver `button` styling but I may have to pursue the overlay option as part of the `Tooltip.plugin`.

Issue #614, opened 2018-08-15
Hierarchy - Need a way to dynamically update actions in leaf action menu


krishollenbeck:
**Describe the solution you'd like**
Hierarchy control needs a way to allow actions in action menu to be updated dynamically. Some actions may not be available until a certain leaf is selected.

![menu](https://user-images.githubusercontent.com/1056684/44151116-3018ee68-a067-11e8-975b-5b990b865c1d.png)

tmcconechy:
This fix was merged

PR #615, opened 2018-08-15
Fix for 614 - Hierarchy - Need a way to dynamically update actions in leaf action menu

krishollenbeck:
Adds the ability to update the actions menu on a specific leaf.

https://github.com/infor-design/enterprise/issues/614

-  With suggested changes. Navigate to http://localhost:4000/components/hierarchy/example-context-menu-with-details.html
- Action menu will have two new actions ("Delete, Move") which were added after the initial render.

### DIFF
--- a/src/components/hierarchy/hierarchy.js
+++ b/src/components/hierarchy/hierarchy.js
@@ -235,12 +235,31 @@ Hierarchy.prototype = {
   },
 
   /**
+   * @public
+   * @param eventInfo eventType, target, data, ect..
+   * @param newActions new actions to be appended to the menu
+   */
+  updateActions(eventInfo, updatedActions) {
+    const leaf = $(eventInfo.targetInfo.target).closest('.leaf');
+    const nodeData = eventInfo.data;
+    const popupMenu = $(leaf).find('.popupmenu');
+    const lineItemsToRemove = popupMenu.find('li').not(':eq(0)');
+
+    $(lineItemsToRemove).each((idx, item) => {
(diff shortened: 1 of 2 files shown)


src/components/hierarchy/hierarchy.js
```diff
@@ -234,15 +234,39 @@ Hierarchy.prototype = {
     });
   },
 
+  /**
+   * @public
+   * @param eventInfo eventType, target, data, ect..
```
tmcconechy: This should be giving a warning as they need a type.

`* @param {object} eventType Infor about the event such as eventType, target, data, ect..`

Also should add a title.

src/components/hierarchy/hierarchy.js
```diff
@@ -252,16 +276,24 @@ Hierarchy.prototype = {
...
-      const items = `${data.menu.actions.map(a => "<li><a href='" + a.url + "' data-action-reference='" + JSON.stringify(a.data) + "'>" + a.value + "</a></li>").join('')}`;
-      template.push(items);
+      template.push(this.getActionMenuItems(data));
     }
 
     template.forEach((i) => { popupMenu.append(i); });
   },
 
+  /**
+   * @private
```
tmcconechy: Same here

src/components/hierarchy/hierarchy.js
krishollenbeck: Okay, just pushed a fix for that.

Issue #633, opened 2018-08-20
Dropdown configured as multiselect: tooltip does not display on initial load or when readonly


fitzorama:
**Describe the bug**
When the dropdown is configured as a multi-select and enabled with a tooltip, the tooltip does not display until a selection has been made or changed and does not display when the component is configured as readonly.

**To Reproduce**
Using the updated attached test-multiselect-with-tooltip.html in the same named .zip file:
1. Load the form
2. Hover over the fields labelled Configured like Multiselect with Tooltip - Pre-selected and Configured like Multiselect with Tooltip - Pre-selected Readonly
3. Tooltip does not display
4. Make an additional selection for the field labelled Configured like Multiselect with Tooltip - Pre-selected
5. Tooltip displays

**Expected behavior**
Expect the tooltip to display without opening the dropdown or making or changing a selection. Expect the tooltip to display when the component readonly state is set to true.

**Screenshots**

[test-multiselect-with-tooltip.zip](https://github.com/infor-design/enterprise/files/2303731/test-multiselect-with-tooltip.zip)
![dropdown](https://user-images.githubusercontent.com/20212079/44363256-55b1f180-a489-11e8-9257-febc6a42328c.gif)

EdwardCoyle:
The fix for this is to make (this codepath)[https://github.com/infor-design/enterprise/blob/master/src/components/dropdown/dropdown.js#L2363] run during initialization  of the Dropdown... currently nothing is setting up tooltips on init.  This would be an issue with any Dropdown/Multiselect component that has text long enough to be cut off with ellipsis.

brianjuan:
http://4100-beta0-enterprise.demo.design.infor.com/components/dropdown/test-tooltips.html
Passed QA testing

PR #641, opened 2018-08-22
633 - Added tooltip in initial load

ericangeles:

Created a method `initTooltip()` in `dropdown.js` and triggered it in initialization.
The current tooltip runs only in the condition when firing the change event with the new value in multiselect.

`dropdown.js` (L#495)
```
initTooltip() {
    const opts = this.element.find('option:selected');
    const optText = this.getOptionText(opts);

    this.tooltipApi = this.pseudoElem.find('span').tooltip({
      content: optText,
      trigger: 'hover',
    });
  },
```

Run it inside of init() (L#314)
```
this.initTooltip();
```

Also, run it inside of the condition where firing the event. Start from line (L#2367)
```
if (!noTrigger) {
      // Fire the change event with the new value if the noTrigger flag isn't set
      this.element.trigger('change').triggerHandler('selected', [option, isAdded]);

      if (this.pseudoElem.find('span').width() >= this.pseudoElem.width()) {
        this.initTooltip();
        
      } else if (this.tooltipApi) {
        this.tooltipApi.destroy();
      }
    
```

Closes https://github.com/infor-design/enterprise/issues/633

- Run http://localhost:4000/components/dropdown/test-tooltips.html
- Load the form
- Hover over the fields labelled Configured like Multiselect with Tooltip - Pre-selected and Configured like Multiselect with Tooltip - Pre-selected Readonly
- In Multiselect with Tooltip, select values until the ellipsis appeared
- Tooltips should be display upon hovering

**File/s**

- **M** - src/components/dropdown.js


tmcconechy:
Ok, if you put in a fix, it should rebuild and hopefully the tests pass after 👍

PR #646, opened 2018-08-22
245 and 555 - Fixes for cutoff popupmenus and incorrect active states on fields with options

### LINKED ISSUE #245: SOHO-7978: Right click (long press) for context menus on mobile

### LINKED ISSUE #555: Enable permanently-displayed field options button in touch environments

EdwardCoyle:
This PR addresses some mobile usability issues that came up as a result of the work on #245 (and related #555):

- Popupmenus that open up while a software keyboard is visible on screen would sometimes become half-sized or cut off.  Often the menus were positioned in a way that covered the trigger buttons, due to a timing issue that was causing the menus to open during the keyboard animations.
- Active states of input fields and field-options buttons would sometimes remain if a field options menu was opened, and focus was set on an outside input field.
- It was sometimes possible to incorrectly scroll out of bounds on an iOS device (read: the blue header was not always present on-screen because it had been unexpectedly scrolled out of view, even if the body tag explicitly had a `.no-scroll` CSS class set).

**Related github/jira issues (required)**:
Fixes #245
Fixes #555

Pull this branch and run the demoapp.  Then:

_For popupmenu related changes:_
- open http://localhost:4000/components/field-options/example-input-fields on iOS
- focus the middle input field
- tap the middle input field's field options button
- the keyboard should roll out, and a menu should open up aligned to underneath the field options button.

_For testing active state changes:_
- open http://localhost:4000/components/field-options/example-input-fields on iOS
- focus any input field
- tap the input field's field options button
- focus any other input field
- the active state (blue border) on both the input field and the field options button should disappear.

### DIFF
--- a/src/utils/environment.js
+++ b/src/utils/environment.js
@@ -4,6 +4,9 @@ import { breakpoints } from './breakpoints';
 // jQuery Components
 import './debounced-resize.jquery';
 
+// Utility Name
+const UTIL_NAME = 'environment';
+
 /**
  * @class {Environment}
  */
@@ -26,6 +29,7 @@ const Environment = {
     $('html').attr('data-sohoxi-version', SOHO_XI_VERSION);
     this.addBrowserClasses();
     this.addGlobalResize();
+    this.addGlobalEvents();
   },
 
   /**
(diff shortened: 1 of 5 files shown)


src/utils/environment.js
```diff
@@ -26,6 +29,7 @@ const Environment = {
     $('html').attr('data-sohoxi-version', SOHO_XI_VERSION);
     this.addBrowserClasses();
     this.addGlobalResize();
+    this.addGlobalEvents();
```
tmcconechy: Duplicate call?

tmcconechy: Woops bad eyes

PR #676, opened 2018-08-27
NG#51 - Fixing an issue when clearing datepicker with backspace

dmcodrum:
We found as noted on #51 that the change event was not firing when a user will backspace out a field on the datepicker (using a mask). Because of this data can be lost.

Closes infor-design/enterprise#51

- follow the steps on the ng issue after building this project and copying it over to ids-enterprise-nt

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ids-enterprise",
   "slug": "ids-enterprise",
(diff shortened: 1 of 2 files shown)


tmcconechy:
will use https://github.com/infor-design/enterprise/pull/677 instead

Issue #671, opened 2018-08-27
Showing whitespace while adding a tab in a card


Venket35:
**Describe the bug**
Showing whitespace while using tab in a card component.

**To Reproduce**
Use below sample adding by tmcconechy

**Expected behavior**
Whitespace should not come.
![error1](https://user-images.githubusercontent.com/42734126/44649378-4c80d180-aa01-11e8-8301-5d7f69f0bd7f.jpg)

**Screenshots**

**Platform**
 Infor_HomepagesWidgetSDK_1.0.24_20180523

**Additional context**

tmcconechy:
I was able to reproduce this but putting a tab on a card...

```
  <div class="two-thirds column">
    <div class="card">
      <div class="card-header">
          <h2 class="widget-title">Card Title</h2>
          <button class="btn-actions" type="button">
            <span class="audible">Actions</span>
            <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
              <use xlink:href="#icon-more"></use>
            </svg>
          </button>
          <ul class="popupmenu">
            <li><a href="#">Action One</a></li>
            <li><a href="#">Action Two</a></li>
          </ul>

      </div>
      <div class="card-content">
        <div id="tabs-normal" class="tab-container">
          <ul class="tab-list">
            <li class="tab"><a href="#tabs-normal-contracts">Contracts</a></li>
            <li class="tab"><a href="#tabs-normal-opportunities">Opportunities</a></li>
            <li class="tab is-selected"><a href="#tabs-normal-attachments">Attachments</a></li>
            <li class="tab"><a href="#tabs-normal-contacts">Contacts</a></li>
            <li class="tab"><a href="#tabs-normal-notes">Notes</a></li>
          </ul>
        </div>
        <div class="tab-panel-container">
          <div id="tabs-normal-contracts" class="tab-panel">
            <p>Facilitate cultivate monetize, seize e-services peer-to-peer content integrateAJAX-enabled user-centric strategize. Mindshare; repurpose integrate global addelivery leading-edge frictionless, harness real-time plug-and-play standards-compliant 24/7 enterprise strategize robust infomediaries: functionalities back-end. Killer disintermediate web-enabled ubiquitous empower relationships, solutions, metrics architectures.</p>
          </div>
          <div id="tabs-normal-opportunities" class="tab-panel">
[...]

PR #681, opened 2018-08-28
671 - Fixes whitespace in tablist in a card

ericangeles:

The issue is very inconsistent when resizing the browser and it flickers.

I changed the right: 51px to right: 0; to eliminate the white space when the button arrow is not yet visible.

This is under the `.tab-container` and `.horizontal` container.
```
&::after {
    right: 0; // bleed into the tab list (width of this element + 1 for border) & eliminate unncessary white space
  }
```

Then I created a css rule where I will back the position of the fade edge in the tab list so that it will not overlap when the button arrow is visible.

```
.has-more-button.tab-container.horizontal,
.has-more-button.tab-container.header-tabs {
  &::after {
    right: 51px;
  }
}
```

I created a test page for this,

http://localhost:4000/components/cards/test-tabs-card.html

Closes https://github.com/infor-design/enterprise/issues/671

- Run http://localhost:4000/components/cards/test-tabs-card.html
- Try to play it around by resizing the browser. Check the behavior of the fade edge and the button arrow until it appears.
- White space shouldn't come anymore.

**File/s**
**N** - app/views/components/cards/test-tabs-card.html
**M** - src/components/tabs/_tabs.scss

### DIFF
--- a/app/views/components/cards/test-tabs-card.html
+++ b/app/views/components/cards/test-tabs-card.html
@@ -1,4 +1,4 @@
-<div class="two-thirds column">
+<div class="two-thirds column row">
     <div class="card">
       <div class="card-header">
           <h2 class="widget-title">Card Title</h2>
--- a/app/views/components/cards/test-tabs-card.html
+++ b/app/views/components/cards/test-tabs-card.html
@@ -1,0 +1,48 @@
+<div class="two-thirds column">
+    <div class="card">
+      <div class="card-header">
+          <h2 class="widget-title">Card Title</h2>
+          <button class="btn-actions" type="button">
+            <span class="audible">Actions</span>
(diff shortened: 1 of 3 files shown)


app/views/components/cards/test-tabs-card.html
```diff
@@ -0,0 +1,48 @@
+<div class="two-thirds column">
```
tmcconechy: Minor but add class top-padding here ...

tmcconechy:
Yeah we should probably fix that as well. Someone reported this on MS teams but wasnt able to reproduce it

app/views/components/cards/test-tabs-card.html
ericangeles: Probably a row class? It seems that top-padding class only works on the same level of the page-container.

Issue #389, opened 2018-07-17
SOHO-8075: Datagrid - tabbing into editable cells not working


clepore:
Landmark requires the ability for power uses to quickly add rows to a datagrid.  

With editable cells in the datagrid, the tab key (to move to cell to the right) or tab + shift key (to move to the cell to the left) does not move across the cells of a row.  When tabbing into an editable cell, the cell should be in edit mode.

Tried adding the grid options of actionableMode: true and cellNavigation: true, but these don't affect the desired behavior.

This worked in 3.5 with the grid option of allowTabs: true. But allowTabs doesn't seem to available in 4.0.  

I've attached an updated version of example-editable.html to illustrate. Also attached a gif file to demonstrate how it was working in 3.5 to show what's needed in 4.0.

brianjuan:
Moving to Failed QA. When the user tab into the Order Date column and accidentally typed an invalid date, for example: 2/29/2018 then hit Tab again, the focus will go on the Pagination control not to the next column.

You can also notice that there still a focus on the Order Date column even you click on the other cells.

tmcconechy:
Im seeing an error "Uncaught TypeError: Cannot read property 'getDate' of undefined" in the console. This might be why... @deep7102

tmcconechy:
As noted
- type 11/ in the date cell gives an error
- tab into dropdown column and it opens (should need to do down arrow to open)
- tab out of dropdown stays open (related to previous)
- datepicker when in error has an uneeded box shadow

PR #689, opened 2018-08-28
389 - Fixed locale js error and dropdown editor with Datagrid

deep7102:
Using keys to date picker editor to create validation error was generating js error then focusing out from grid. Dropdown editor was opening it self with focus so this PR will fix those issues with Datagrid.

#389
https://jira.infor.com/browse/SOHO-8075

http://localhost:4000/components/datagrid/example-editable.html
- Open above link
- Open developer tools
- Use keyboard to get to grid
- Press F2 to toggles actionableMode "true"
- Use tab key to get date picker editor cell
- Type "00/00/0000" and press tab
- It should make validation error here and go to next cell (dropdown editor)
- See dropdown editor should not open (it should be focused state only, use arrow down key to open)
- See console should not be any js error

### DIFF
--- a/src/components/datagrid/_datagrid.scss
+++ b/src/components/datagrid/_datagrid.scss
@@ -1838,20 +1838,31 @@ $datagrid-short-row-height: 23px;
         // scss-lint:enable NestingDepth
 
         &:focus {
-          box-shadow: none;
+          // scss-lint:disable ImportantRule
+          box-shadow: none !important;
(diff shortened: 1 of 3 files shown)


tmcconechy:
@deep7102 I think you should push this to the 4.10.x branch since it is a regression issue on a failed QA item. You will have to redo the PR on that branch i think :(

PR #693, opened 2018-08-28
Add uplift theme

### LINKED ISSUE #690: Update theme switching to use ids-identity as a source
We already have an existing theme switcher in Enterprise, we just need to switch the source to pull the themes from `ids-identity`

n/a

n/a

**Additional context**
Once all is linked properly, you might not see much change and this is expected. The original 3 themes (light/dark/high-contrast) should look identical to the previous setup and the uplift theme is at most, 10% done/implemented.

tmcconechy:

Added the new theme into the switcher and the building out of the new theme from the ids-identity package. Updated ids. Now work can continue on the new uplift theme. 

As of now the only thing i see fixed / changed in the theme is the background color on the inputs. More will be added by design in the future. On #690 i noted some of the issues necessary to fix along the way (colors / button height) to get more working as we progress with the theme.

Closes #690 

http://localhost:4000/components/input/example-index.html  http://localhost:4000/components/checkboxes/example-index.html
- select "uplift" from the theme picker in the top right ... button


app/views/components/personalize/example-settings-page.html
```diff
@@ -27,6 +27,7 @@ <h3>Themes & Personalization</h3>
           <option selected value="light">Light</option>
           <option value="dark">Dark</option>
           <option value="high-contrast">High Contrast</option>
+          <option value="uplift">High Contrast</option>
```
EdwardCoyle: I'm guessing this should be "Uplift"

tmcconechy: I'll fix this in next pr...

Issue #692, opened 2018-08-28
Checkbox Component. Dirty tracker not appearing on high contrast


CindyMercadoReyes:
**Describe the bug**
[All Browsers] Checkbox Component. Dirty tracker not appearing on high contrast on "Option 3".

**To Reproduce**
1. Go to url: http://4100-beta1-enterprise.demo.design.infor.com/components/checkboxes/example-horizontal.html
2. Click on the menu and select "High Contrast'
3. Click on "Option 3"
4. The dirty tracker is not appearing. 

**Expected behavior**
The dirty tracker should appear on all three themes

**Screenshots**

![checkbox tracker_2](https://user-images.githubusercontent.com/39101346/44743783-6db0f300-aad1-11e8-964e-482f2073d178.png)

![checkbox _1](https://user-images.githubusercontent.com/39101346/44743839-920ccf80-aad1-11e8-9b13-1893d0edfb6b.png)

**Platform**
- All Browsers

**Additional context**
None

tmcconechy:
Looks like this needs zindex and a but of adjustment on top/left as well.

brianjuan:
http://master-enterprise.demo.design.infor.com/components/checkboxes/example-horizontal.html
Passed QA testing, bug fix is verified working across all browsers.

PR #711, opened 2018-08-30
692  - Added a fix for dirty tracker

ericangeles:

Adding an `z-index: 1;` will display the dirty tracker in high contrast theme. Also, added some styling for position of the icon.

Closes https://github.com/infor-design/enterprise/issues/692

- Run http://localhost:4000/components/checkboxes/example-horizontal.html
- Click on the menu and select "High Contrast'
- Click on "Option 3"
- The dirty tracker should now appear and in proper position


tmcconechy:
Also its just a pixel or so off on the horizontal example.

screen shot 2018-08-30 at 12 45 27 pm

ericangeles:
The position of the icon dirty is dynamic. Will look into the trackdirty.js.

app/views/components/checkboxes/example-horizontal.html
```diff
@@ -14,7 +14,7 @@
         <label for="checkbox2" class="checkbox-label" >Option 2</label>
       </div>
 
-      <div class="field">
+      <div class="field field-dirty">
```
deep7102: I think we should not add extra classes if we do not need it

src/components/icons/_icons.scss
```diff
@@ -189,12 +189,20 @@ svg.icon-logo {
   width: 34px;
 }
 
+.field-dirty {
```
deep7102: Instead of this new/extra rule, if we just adjust `margin` it should fix the alignment for selector `.compound-field > .field.field-checkbox .dirty-checkbox`

deep7102:
This file does not need any changes, please revert and remove this file from this PR's changes list.
```git checkout src/components/trackdirty/trackdirty.js```

Issue #282, opened 2018-06-27
Accordion: focus styling remains when clicking headers (in safari)


gushanson:
**Describe the bug**
The blue focus styling remains on a clicked header even after another header has been clicked.

**To Reproduce**
- Go to url: http://master-enterprise.demo.design.infor.com/components/accordion/example-index.html
- Click Warehouse Location header
- Click Sort By header
- Click brand name header

**Expected behavior**
The Warehouse Location is not styled as focused.

**Actual behavior**
The Warehouse Location is styled as focused.

**Additional context**
Version 4.7.0
**Safari only**

tmcconechy:
This is actually correct we have two types.

Panel-style
http://master-enterprise.demo.design.infor.com/components/accordion/example-accordion-panels.html
- used for navigation/selection

In-page / Traditional
http://master-enterprise.demo.design.infor.com/components/accordion/example-index.html
- just has expandable sections and content inside.
- focus the headers to activate 

So this example is correct.

gushanson:
![selected](https://user-images.githubusercontent.com/4129417/42166912-be0372de-7e0c-11e8-90cb-2b7cac222c97.png)
What I'm seeing is the scenario in the image. I sthis really by design?

tmcconechy:
That is an older link. How did you get that? By clicking each one at a time as im not seeing that here o the latest link (Pre 4.8.0)

http://master-enterprise.demo.design.infor.com/components/accordion/example-index.html

tmcconechy:
OK, Nevermind i do see that on safari. I'll reopen.

ericangeles:
I also noticed that the button arrow is also focused in safari(v.11.1.2) when toggling. But in the other browsers looks fine.

<img width="712" alt="screen shot 2018-08-30 at 10 25 55 am" src="https://user-images.githubusercontent.com/8839327/44825907-46783400-ac3f-11e8-9242-b8241208c1ee.png">

brianjuan:
http://master-enterprise.demo.design.infor.com/components/accordion/example-index.html
Passed QA testing. Bug fix is working as expected on Safari.

PR #712, opened 2018-08-30
282 - Added fix for accordion focus

ericangeles:

A css hack/trick that fix the issue of accordion focus in safari. It seems that the click event in anchor tag doesn't firing but the button tag with icon is somehow working.

```
//Button for Safari issue
.is-safari {
  .accordion-header {
    > [class^="btn"]:focus:not(.hide-focus) {
      border-color: transparent;
      box-shadow: none;
    }
  }
  .accordion-header.has-chevron {
    > [class^="btn"] {
      position: absolute;
      left: 0;
      height: 45px;
      width: 100%;
      margin: 0;
      padding: 13px 0 28px 13px;

      &:focus:not(.hide-focus) {
        border-color: transparent;
        box-shadow: none;
      }

      .icon.chevron {
        right: 0;
        position: absolute;
        width: 50px;
        margin-top: 5px;
      }
    }
  }
}
```

Closes https://github.com/infor-design/enterprise/issues/282

- Run http://localhost:4000/components/accordion/example-index.html
- Click the title header of the accordion
- Focus style should not be visible anymore


EdwardCoyle:
Just noting for other reviewers - the original issue #282 specifically pegged this as an issue on OSX Safari, so make sure you test there.

src/components/accordion/_accordion.scss
```diff
@@ -573,8 +577,8 @@
 
   // Selected color scheme
   &.is-focused:not(.hide-focus) {
-    border-color: $accordion-focused-border-color;
-    box-shadow: $focus-box-shadow;
+    border-color: none;
```
EdwardCoyle: This isn't going to work because it's completely disabling the focus styles.  See the GIF below:

aug-30-2018-12_37-pm

What's happening in the video is me navigating the accordion with the keyboard, using the Tab key to change the active header, then the Spacebar to expand/collapse the header.  With these changes, it's impossible for the user to understand at a glance which accordion header is currently the one that's focused.

ericangeles:
Gotcha. I see now the issue here. I believe the .hide-focus class is not triggering when clicking to the other accordion in Safari.

clepore:
@ericangeles If you make the changes @EdwardCoyle requested, this seems like an okay fix to move forward with. We'll review it tomorrow if you have any questions.

EdwardCoyle:
@davidcarlsonberg, RE:

> I do see a separate issue where if there is an accordion open and then click an accordion other than the one open, there is a slight horizontal flicker movement. This also happens on close of the second accordion. Perhaps this is for another issue.

I saw that too, but I think that's unrelated.  It's been present for awhile.  We could make a separate ticket to fix that -- I think it's padding-related and we'd need to find a way to smoothly animate across all the accordion styles.

src/components/accordion/_accordion.scss
```diff
@@ -1384,3 +1384,35 @@ html[dir='rtl'] {
...
+
+//Button for Safari issue
+.is-safari {
+  .accordion-header {
+    > [class^="btn"]:focus:not(.hide-focus) {
+      border-color: transparent;
+      box-shadow: none;
+    }
+  }
+  .accordion-header.has-chevron {
```
EdwardCoyle: Since we're using SASS, we can probably move this into the `.accordion-header` rule above this and use `&.has-chevron {` here

src/components/accordion/_accordion.scss
```diff
@@ -1384,3 +1384,35 @@ html[dir='rtl'] {
...
+    > [class^="btn"] {
+      position: absolute;
```
EdwardCoyle: I'm agree with @LarryAmisola here, alphabetizing the CSS properties in each rule is actually built into our code style guidelines as a requirement.  I'm realizing as I write this that our SASS linter might be working.  Can you re-organize these manually?

davidcarlsonberg:
A note that another ticket spawned from this one: https://github.com/infor-design/enterprise/issues/774

PR #715, opened 2018-08-30
700 - Fixed indeterminate paging with Datagrid

### LINKED ISSUE #700: Datagrid Paging Indeterminate: Paging controls doesn't work after initial interaction
**Describe the bug**
[All Browsers] Datagrid Paging Indeterminate: Paging controls doesn't work after initial interaction
http://4100-beta1-enterprise.demo.design.infor.com/components/datagrid/example-paging-indeterminate.html
http://master-enterprise.demo.design.infor.com/components/datagrid/example-paging-indeterminate.html

**To Reproduce**
1. Go to URLs above
2. Click Next Page icon or any icons in the paging control
3. Datagrid will update but the paging controls no longer work, First Page and Previous Page are disabled and nothing happens when you click Next Page and Last Page

**Expected behavior**
Paging control should work

**Screenshots**
No applicable screenshots.

**Platform**
 - All devices and browsers

**Additional context**
No additional context.

deep7102:
Using indeterminate paging was not working after initial interaction with Datagrid.

Closes #700

http://localhost:4000/components/datagrid/example-paging-indeterminate.html
http://localhost:4000/components/datagrid/example-paging-indeterminate-landmark.html
- Open any above link
- Click Next Page icon or any icons in the paging control
- Paging should work fine

### DIFF
--- a/app/views/components/datagrid/example-paging-indeterminate.html
+++ b/app/views/components/datagrid/example-paging-indeterminate.html
@@ -33,6 +33,7 @@
 
       //Init and get the api for the grid
       var pageSize = 10;
+      var lastPage = -1;
 
       grid = $('#datagrid').datagrid({
         columns: columns,
@@ -48,6 +49,9 @@
             //req.activePage = 2;
             return; //Or can ignore and call loadData
           }
+          if (req.type === 'last') {
+            req.activePage = lastPage;
+          }
 
           var url = '{{basepath}}api/compressors?pageNum='+ req.activePage +'&pageSize='+ req.pagesize;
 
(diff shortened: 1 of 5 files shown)


app/views/components/datagrid/example-paging-indeterminate.html
```diff
@@ -48,6 +49,9 @@
             //req.activePage = 2;
             return; //Or can ignore and call loadData
           }
+          if (req.type === 'last') {
+            req.activePage = lastPage;
+          }
```
pwpatton: Anyway we can avoid doing this in the implementing code? ActivePage really has no meaning when it comes to indeterminate paging anyway. Could we just say if it's indeterminate it's -1 all the time?

pwpatton: I see in the test you always set to -1 if indeterminate. That should be fine, but it would be nice to not have to set it at all.

pwpatton:
http://localhost:4000/components/datagrid/example-paging-indeterminate.html starts on the 2nd page or recs. This was for issue #576 that I had previously fixed. Does it work on the example-paging-indeterminate.html sample to 
1. click prev button to get ot the first page
2. then click next button until last page is reached.

pwpatton:
I just did some testing on this and it seems to work quite nicely in the samples. I'll know more once this makes it into landmark.

app/views/components/datagrid/example-paging-indeterminate.html
deep7102: Yeah! I agree you right less code is always good to manage too, but unfortunately its been already there so I try to fix around and not to break other thing might related to ActivePage. We should revisit and dig more to remove ActivePage for indeterminate option.

Issue #721, opened 2018-08-30
Document how to include and use Enterprise


clepore:
**Is your feature request related to a problem? Please describe.**
In the course of the migration to github, we lost some documentation on how to use/apply Enterprise into a system.

**Describe the solution you'd like**

1. Relocate the development information
    - [ ] Clean up [docs/DEVELOPER.md](https://github.com/infor-design/enterprise/blob/master/docs/DEVELOPER.md)
    - [ ] Move any developer instructions from [./README](https://github.com/infor-design/enterprise/blob/master/README.md) into [docs/DEVELOPER.md](https://github.com/infor-design/enterprise/blob/master/docs/DEVELOPER.md)

1. Add implementation instructions
    - [ ] Clean up [./README](https://github.com/infor-design/enterprise/blob/master/README.md)
    - [ ] Add instructions on how to include/use the controls into [./README](https://github.com/infor-design/enterprise/blob/master/README.md)
    - [ ] Maybe somehow add a link to the `docs/` folder "for more info" or something

1. Update [docs/index.md](https://github.com/infor-design/enterprise/blob/master/docs/index.md) (@clepore)
    - [ ] Link back to [./README](https://github.com/infor-design/enterprise/blob/master/README.md)
    - [ ] Make it more of a general introduction to EP (as this is what's shown on the website)

1. Make sure the changes pass `npm run mdlint:docs`

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

tmcconechy:
Some thoughts...

1) point out the links if using angular
2) go through the process of 
- npm install
- what to include in the page
- adding a component
- point to the example

tmcconechy:
Seems like this should have been closed.

PR #722, opened 2018-08-30
721 - Normalize line endings for Windows users

EdwardCoyle:
Rather than relying on Windows users to need to change their global Git settings, this PR adds a `.gitattributes` file with some reasonable line ending defaults for web projects.  This selectively  prevents Git's autoconversion from LF to CRLF on Windows.

Previously this was solved by doing the following on Windows machines:
```
git config core.autocrlf false
```

This should no longer be necessary after this PR.

#721

Pull down the repo on Windows and open any Javascript file.  In an editor that reports it, make sure these files contain all `LF`-style line endings.


EdwardCoyle:
@krishollenbeck @pwpatton 

Would you guys mind checking out this PR and let us know if you have any issues with line ending conversions?

krishollenbeck:
@EdwardCoyle I checked out the branch. And I am not positive it is working. Some of the .JS files show LF. But if I open the HTML files. Some of them appear as CRLF. See the bottom right in the attached screenshot

[
intellij-crlf
](url)

PR #733, opened 2018-08-31
724 autocomplete chars

### LINKED ISSUE #724: Autocomplete search term that contains a value not Alphanumeric does not work
We are using the autocomplete to select a UNC path on a remote server, \\Server\Share, and the service is returning the filtered values, but then the control excludes them.

**To Reproduce**
1. http://master-enterprise.demo.design.infor.com/components/autocomplete/test-xss-security.html
2. Type in "\\"
3. Check the term that is passed to the openList function 

**Expected behavior**
The term "\\" is passed to the openList function.

**Platform**
All

**Additional context**
The xss security on the searchTerm is stripping the double backslash.
return dfd.resolve(xssUtils.ensureAlphaNumeric(searchTerm), response);

xssUtils.ensureAlphaNumeric('\\') returns an empty string.

tmcconechy:

The security rules in autocomplete are two strict and removing special characters. Relaxing them to only remove tags.

Closes #724

http://localhost:4000/components/autocomplete/test-xss-security.html
- open dev tools
- stop debugger on 
```
openList(term, items) {
    if (this.element.is('[disabled], [readonly]') || this.isLoading()) {
      return;
    }
```
- type a / in the input
- watch the term


tmcconechy:
Redoing this fix.

PR #754, opened 2018-09-05
748 - add disable method on Tree

### LINKED ISSUE #748: Add Tree disable() method
A feature method to disable the entire Tree.

The assumed solution is the ability to disable the entire Tree by disabling each node.

An alternative solution could be disabling the parent element in Tree and styling the nodes as disabled based on a parent class.
An additional method could be to disable a single node.

**Additional context**
Parent issue: #561 .

davidcarlsonberg:
Add disable() method on tree by grabbing all anchors, putting into array and looping over to add `.is-disabled` class.

Resolves #748 

- Pull branch and run app.
- Go to http://localhost:4000/components/tree/example-disable.html
- Click `#disable` button to ensure disabling works on all nodes. There are two versions of calling the method. Reviewer can comment out one to try the other.

Suggestions welcome as this is my first pass at Tree and will inform other method creations such as `enable()` and `readonly()`.


app/views/components/tree/example-disable.html
```diff
@@ -0,0 +1,187 @@
...
+  // $('#disable').on('click', () => {
+  //   $('#tree').disable();
+  // });
+
+  const tree = $('#tree').tree().data('tree');
+
+  // method call on tree api
+  $('#disable').on('click', () => {
+    tree.disable();
+  })
```
deep7102: add `;` semicolon at the end of statement.

src/components/tree/tree.js
```diff
@@ -2140,6 +2140,20 @@ Tree.prototype = {
...
   },
+
+  /**
+   * Disables all nodes in the Tree component
+   * @returns {void}
+   */
+  disable() {
+    const nodes = this.element[0].querySelectorAll('a');
+    nodes.forEach(node => {
+      if ( !(node.classList.contains('is-disabled')) ) {
```
deep7102: Looks good! just to improve performance. Please remove this if statement, since `classList.add` method will take care for not to duplicate. So in this case it's just an extra check we can avoid it.

PR #763, opened 2018-09-06
749 - add enable method on Tree

### LINKED ISSUE #749: Add Tree enable() method
A feature method to enable the entire Tree.

The assumed solution is the ability to enable the entire Tree by enabling each node.

An alternative solution could be enabling the parent element in Tree and styling the nodes as enabled based on a parent class.
An additional method could be to enable a single node.

**Additional context**
Parent issue: #561 .

davidcarlsonberg:
Add `enable()` to enable all nodes on Tree.

Resolves #749 

Pull branch and run app.
Browse to http://localhost:4000/components/tree/example-enable.html
Verify that click of `#enable` button enables all nodes on Tree.


src/components/tree/tree.js
```diff
@@ -2140,6 +2140,18 @@ Tree.prototype = {
     this.element.empty();
     $.removeData(this.element[0], COMPONENT_NAME);
   },
+
+  /**
+   * Enables all nodes in the Tree component
+   * @returns {void}
+   */
+  enable() {
+    const nodes = this.element[0].querySelectorAll('a');
```
EdwardCoyle: Just a minor thing:  this method should probably also remove the `aria-disabled` attribute from each node as well, for better screen reader support:

screen shot 2018-09-06 at 3 34 51 pm

The nodes are definitely enabled, but a screen reader may interpret them otherwise while the attribute's present.

davidcarlsonberg:
Will need merge of master into this branch once branch `751-tree-restore-method` is merged into master.

PR #772, opened 2018-09-07
First pass at replacing some color palette color variables with token variables

### LINKED ISSUE #708: Replace color palette sass vars with more contextual tokens (first pass)
Currently we use a color palette in Enterprise that isn't dynamic per theme like the ids-identity package we are slowly moving to.

As a first pass, we want to go through the sass files and replace as many `$color##` (i.e. `$azure06`) variables from the internal Enterprise color palette as we can with more contextual token variables from ids-identity.

n/a

**Additional context**

As described this will be a first pass (probably out of a few), so we want to:
- try to replace as many as possible
- note ones that we can't replace just yet
- open a single issue (in the design-system repo) to add any tokens we may need

On completion, follow up with someone to get https://github.com/infor-design/design-system/issues/194 started.

clepore:
We want to slowly replace references to direct color variables from the color palette and use contextual token variables in their place.

_i.e_ Replace `color: $azure06` with `color: $color-primary`

Closes #708 

1. `npm start` to start the demo app
1. Verify
    - Colors look normal on the normal theme
    - Some colors change when switching themes ~, specifically the new Uplift theme~

_Edit: The themes aren't changing properly as there is a major structure/cascade issue on the tokens side. This can still be merged, but don't expect to see anything different on the Uplift theme when switching and the hc/dark themes will show some things different, but are also flawed._

### DIFF
--- a/app/src/sass/_directory.scss
+++ b/app/src/sass/_directory.scss
@@ -13,7 +13,7 @@
 
 .directory-listing {
   .icon {
-    fill: $azure07;
+    fill: $color-primary-alt;
 
     &.example {
       fill: $ruby05;
--- a/src/components/accordion/_accordion.scss
+++ b/src/components/accordion/_accordion.scss
@@ -1388,7 +1388,7 @@ html[dir='rtl'] {
 .is-safari {
   .accordion-header {
     &.has-chevron {
-      > [class^="btn"] {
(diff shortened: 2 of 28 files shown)


clepore:
Merge after https://github.com/infor-design/enterprise/pull/771.

PR #777, opened 2018-09-10
447 - Added internal grid tooltip with Datagrid

### LINKED ISSUE #447: SOHO-6568: Allow tooltips for datagrid cells where data is not fully displayed.
From Discussion With Tim:
- for a performant tooltip have it only initialize and display on mouseover of it's datagrid cell.
- for keyboard users have a keyboard key that opens the tooltip for the selected cell on focus

Columns with 'tooltips' should also render without the need for https://jira.infor.com/browse/SOHO-6568

deep7102:
For a performant tooltip added algorithm so grid tooltip should not use tooltip component. It only initialize and display on mouseover or keyboard keys with Datagrid.
Keys to toggle tooltip as `show: Shift + F1 and hide: Escape`

Closes #447 (formerly) https://jira.infor.com/browse/SOHO-6568

http://localhost:4000/components/datagrid/example-tooltips.html
- Open above link
- Open developer tools
- Use mouseover or keyboard to see tooltip


deep7102:
Thanks! @davidcarlsonberg and here is some explanation for the area you mention in your comments about you not clear.
 
Row 4 column 8 tooltip is not functioning because if column text have enough space to view then tooltip will not show. You can try dragging column header to make small width and see it should show up the tooltip on this cell too.

Duplicate id of "5", this is part of data "productId: 2542205" from user. This example to show so user can use as their need in data and it should not break.

src/utils/xss.js
```diff
@@ -150,4 +150,21 @@ xssUtils.isUrlLocal = function (url) {
...
 
+/**
+ * Calculate the width for given text string.
+ * @private
+ * @param {string} text string to process
+ * @param {number} padding value for left + right
+ * @param {string} font size and family used with the given text string
+ * @returns {number} calculated width
+ */
+xssUtils.textWidth = function textWidth(text, padding, font) {
```
tmcconechy: Lets put this in stringUtils instead for proper placement. This file is for x-site scripting utils.

test/components/datagrid/datagrid-api.func-spec.js
```diff
@@ -101,7 +101,7 @@ describe('Datagrid API', () => {
   });
 
   it('Should be able to call addRow and removeRow', () => {
```
tmcconechy: Can we add anymore tests for tooltips?

Issue #363, opened 2018-07-16
Dropdown: Autocomplete does not work when caps lock is ON


awbuboltz:
Dropdown will not auto-select a value when pressing a key that starts with it. This only happens when the Caps Lock is ON or while holding the Shift key.

**To Reproduce**
1. Create a dropdown with a, b, c as values:
2. Turn caps lock on, type B

**Expected behavior**
b should be selected

awbuboltz:
Relates to: [LSF-39040: selecting an option in the dropdown (when capslock is on) is not working](http://jira.lawson.com/browse/LSF-39040)

ericangeles:
Where can I replicate this issue?

tmcconechy:
I think you can reproduce this on any example. For example

http://localhost:4000/components/dropdown/example-no-search-lsf.html
or
http://localhost:4000/components/dropdown/example-index

Me and @EdwardCoyle discussed that this might now be working ok due to recent changes. But i see a one issue

http://localhost:4000/components/dropdown/example-index
- hit caps lock
- type n
Bug: the n is not shown in caps.

- hit backspace
- type n
Bug: the n is now in caps and filtering case insenstive as expected.

So now the filtering is working but on the first key stroke the value is not copied in caps over to the search input as a caps value. Maybe just need a check there.

awbuboltz:
Which release can we expect to find this fix in?

tmcconechy:
4.11.0

awbuboltz:
Great, thanks.

awbuboltz:
This was not fully solved, see http://master-enterprise.demo.design.infor.com/components/dropdown/example-no-search.html

tmcconechy:
Yeah i see, its not working on the no-search example :( . I made a new issue https://github.com/infor-design/enterprise/issues/1301 go ahead and follow that one.

PR #780, opened 2018-09-10
363 - Added fix for caps lock

ericangeles:

It might need to put the original value if it pass actual character. Adding:

```
if (e.type === 'input') {
      this.filterTerm = this.searchInput.val();
    } else {
      this.filterTerm += $.actualChar(e);
      if (e.key !== this.filterTerm && e.key.toLowerCase() === this.filterTerm) {
        this.filterTerm = e.key;
      }
    }
```

Closes https://github.com/infor-design/enterprise/issues/363

- Run http://localhost:4000/components/dropdown/example-index
- hit caps lock
- type n -> Bug: the n is not shown in caps.
- hit backspace
- type n
- The first letter should be capitalize when caps lock is activated


awbuboltz:
This was not fully solved, see http://localhost:4000/components/dropdown/example-no-search.html

PR #799, opened 2018-09-13
788 - Fixed actionable mode was not working with Datagrid

### LINKED ISSUE #788: Actionable Mode is not working in 4.10.0 - to move down cells
**Describe the bug**
In actionable mode, after typing in a cell, hitting enter should move down a cell. I also think it should create a new row on last row.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/test-editable-actionable-mode.html
2. Click on an editable cell
3. Type 'xyz'
4. Hit enter (should go to a new row)

**Expected behavior**
Typing in a cell and hitting enter should move to the next row (like google sheets). On last row a new row should be inserted.

**Additional context**
This works on http://490-enterprise.demo.design.infor.com/components/datagrid/example-editable.html (when toggling the mode with F2)

deep7102:
In actionable mode after editing cell and pressing enter key should move to down row in same column cell, but this feature was not working so this PR will fix this actionable mode issue.

Closes #788

http://localhost:4000/components/datagrid/test-editable-actionable-mode.html
- Open above link
- Click on an editable cell
- Type something like 'xyz' 
- Press enter key (should move to down row in same column cell)

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7467,7 +7467,7 @@ Datagrid.prototype = {
 
   setNextActiveCell(e) {
     const self = this;
-    if (e.type === 'keydown' && !self.settings.actionableMode) {
+    if (e.type === 'keydown') {
       if (this.settings.actionableMode) {
         setTimeout(() => {
           const evt = $.Event('keydown.datagrid');


tmcconechy:
Do you maybe need to merge master ? The test that failed was fixed a while ago. 
Maybe this was made from an old branch?

Issue #305, opened 2018-06-29
Datepicker - Time options not working correctly if passed in updated


tmcconechy:
**Describe the bug**
When calling update to reset the settings things are not working right when changing some of the time related settings.

**To Reproduce**
1. Copy this attachment to app/views/components/datepicker

[test-updated.html.zip](https://github.com/infor-design/enterprise/files/2371082/test-updated.html.zip)

2. Run http://localhost:4000/components/datepicker/test-updated.html
3. Click update
4. Selecting a date and set time in the picker.
5. Click to select a different time / date.

At this point an error occurs and the time portion is not updated.
The problem only occurs if using updated. So workaround is to destroy and recreate (this is commented out in the example).

**Expected behavior**
Expect the time options to work the same on updated as they do on init.

**Platform**
tested on chrome

**Additional context**
Problem submitted by Milan Premgi - Email

EdwardCoyle:
I would think this just requires swapping out the contents of the input field as part of the `updated()` method.  Whatever process sets the contents of the input field when a date is chosen from the picker should also happen in `updated()`.

brianjuan:
http://master-enterprise.demo.design.infor.com/components/datepicker/test-updated.html
Passed QA testing across all browsers

PR #821, opened 2018-09-19
305 - datepicker time options

ericangeles:

Additional condition under the mask. Seems that for some reason the format and mask is not updating in updated method.

Closes https://github.com/infor-design/enterprise/issues/305

- Run http://localhost:4000/components/datepicker/test-updated.html
- Click update
- Selecting a date and set time in the picker.
- Click to select a different time / date.

### DIFF
--- a/src/components/mask/mask-input.js
+++ b/src/components/mask/mask-input.js
@@ -97,7 +97,7 @@ MaskInput.prototype = {
           // Check for an instance of a Datepicker/Timepicker Component, and grab the date format
           const datepicker = $(this.element).data('datepicker');
           if ($.fn.datepicker && $(this.element).data('datepicker')) {
-            if (!this.settings.patternOptions) {
+            if (!this.settings.patternOptions && !this.settings.patternOptions.format) {
               this.settings.patternOptions.format = datepicker.settings.dateFormat;
             }
           }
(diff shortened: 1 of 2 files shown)


src/components/mask/mask-input.js
```diff
@@ -97,7 +97,9 @@ MaskInput.prototype = {
           // Check for an instance of a Datepicker/Timepicker Component, and grab the date format
           const datepicker = $(this.element).data('datepicker');
           if ($.fn.datepicker && $(this.element).data('datepicker')) {
-            this.settings.patternOptions.format = datepicker.settings.dateFormat;
+            if (!this.settings.patternOptions) {
+              this.settings.patternOptions.format = datepicker.settings.dateFormat;
```
EdwardCoyle: @ericangeles good call, but `patternOptions` gets setup as an object during the creation of the Component API, so I'm not sure it's ever going to reliably return as empty here.  You may want to also check for a defined `format` property to be safe:

```js
if (!this.settings.patternOptions && !this.settings.patternOptions.format) {
  //...
}
```

For reference, [this](https://github.com/infor-design/enterprise/blob/master/src/components/mask/mask-input.js#L80) is where that object setup happens.

ericangeles:
Hi @EdwardCoyle, changes has been done.

PR #836, opened 2018-09-25
Enabled filtering of child rows

### LINKED ISSUE #437: SOHO-6932: Searchfield filters out child rows in tree grid
For a datagrid tree with X amount of entries and these entries have child rows.

When using the search filter it will only consider rows with that searched word. So in case this row had sub-entries in it, it will not be displayed in the search result.

Expected Result
Sub entries if any shouldn't be filtered out.

tmcastro:
- Updated applyFilter function to enable filtering of datagrid rows using tree child elements
- http://localhost:4000/components/datagrid/test-tree-filter.html

- Closes #437 

- Click on searchfield under "Task" column header
- Test keywords "quotes", "contact" or other keywords that can be found on child nodes
- Rows should be filtered accordingly

### DIFF
--- a/src/components/datagrid/datagrid.formatters.js
+++ b/src/components/datagrid/datagrid.formatters.js
@@ -366,15 +366,15 @@ const formatters = {
 
   // Tree Expand / Collapse Button and Paddings
   Tree(row, cell, value, col, item, api) {
-    const isOpen = item.expanded;
-    const depth = api.settings.treeDepth[row] ? api.settings.treeDepth[row].depth : 0;
+    const isOpen = item ? item.expanded : true;
+    const depth = item && item.depth ? item.depth : 0;
     const button = `<button type="button" class="btn-icon datagrid-expand-btn${(isOpen ? ' is-expanded' : '')}" tabindex="-1"${(depth ? ` style="margin-left: ${(depth ? `${(30 * (depth - 1))}px` : '')}"` : '')}>
       <span class="icon plus-minus ${(isOpen ? ' active' : '')}"></span>
       <span class="audible">${Locale.translate('ExpandCollapse')}</span>
       </button>${(value ? `<span> ${value}</span>` : '')}`;
     const node = `<span class="datagrid-tree-node"${(depth ? ` style="margin-left: ${(depth ? `${(30 * (depth - 1))}px` : '')}"` : '')}> ${value}</span>`;
 
-    return (item[col.children ? col.children : 'children'] ? button : node);
+    return (item && item[col.children ? col.children : 'children'] ? button : node);
   },
 
   // Badge / Tags and Visual Indictors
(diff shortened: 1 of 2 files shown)


tmcastro:
Errors from Travis is caused by the function I've updated (applyFilter). I think I will need help updating the test script if my fix for the ticket is ok.

tmcconechy:
Looks like its just no longer working to set the filter with the API. Since the test "Should be able to applyFilter with the API" is failing. What you can do is try to debug this. Are are some notes on debugging a functional test.

https://github.com/infor-design/enterprise/blob/master/docs/TESTING.md#debugging-functional-tests

src/components/datagrid/datagrid.formatters.js
```diff
@@ -365,16 +365,16 @@ const formatters = {
   },
 
   // Tree Expand / Collapse Button and Paddings
-  Tree(row, cell, value, col, item, api) {
-    const isOpen = item.expanded;
-    const depth = api.settings.treeDepth[row] ? api.settings.treeDepth[row].depth : 0;
+  Tree(row, cell, value, col, item) {
+    const isOpen = item ? item.expanded : true;
+    const depth = item && item.depth ? item.depth : 0;
```
tmcconechy: Why do we remove the treeDepth here?

src/components/datagrid/datagrid.js
```diff
@@ -1540,7 +1540,9 @@ Datagrid.prototype = {
 
       for (let i = 0; i < conditions.length; i++) {
         const columnDef = self.columnById(conditions[i].columnId)[0];
-        let rowValue = self.fieldValue(rowData, columnDef.field);
+
+        let rowValue = rowData ? rowData[columnDef.field] :
```
tmcconechy: Im questioning this change as well, because i think it will cause some nested objects not to work

src/components/datagrid/datagrid.formatters.js
tmcastro: I replaced this code because the previous code is sometimes giving out the wrong depth and giving out wrong indentions for nodes. This happens because the index value of the parent/child nodes are being mixed up since the parent/child nodes are merged into one array in treeDepth.

tmcastro:
- Fixed the expand/collapse feature (I forgot to set the value of depth variable)
- Filter on HMM still works but we are now considering the values of the child nodes so some of the other nodes are still being displayed (Did I misunderstand this?)
- On previous version, filtering "due" displays no results but since we are now checking the child nodes, one of the parent nodes is now displayed together with it's child nodes

tmcconechy:
Yes, i would expect this to work differently. Regarding searching for HHM.
It should only show rows that either
- have HHM in them 
- all the parents up of any child row 
- nothing else

As it stands its showing all kinds of children rows and other rows that do not have HHM in them.

tmcastro:
Ahh ok, TIm. I get it now. I shouldn't display other child nodes that don't have the search key in them. I thought that I should display the parent of the child row and everything under it. I'll work on this now. Thanks!

Issue #791, opened 2018-09-12
Datagrid select header doesn't save filter after reload page


CristianCosta:
**Describe the bug**
The selected filter option is not displayed on the column header, in a datagrid using the saveUserSettings option after reloading the page.

**To Reproduce**
1. Edit http://localhost:4000/components/datagrid/test-save-settings.html so that it contains a filterType: select column
1. Select an option from the select (dropdown) filter.
1. Reload the page (F5)
1. The datagrid stills filtered (so, the filter is being applied) but the selected filter on header is empty.

**Expected behavior**
Select header displays the filter after reloading the site.

**Platform**
 - Browser Name: chrome, firefox.

**Additional context**
This issue is related from [769](https://github.com/infor-design/enterprise/issues/769)

PR #846, opened 2018-09-26
791 - Added fixes for save filter

ericangeles:

It doesn't return an array or object, instead, it returns a scalar value. There's a condition that if there's an array, that's the only time that it will execute the selected property.

```
if (input.is('select')) {
        if (conditions[i].value instanceof Array) {
          for (let j = 0; j < conditions[i].value.length; j++) {
            input.find(`option[value="${conditions[i].value[j]}"]`).prop('selected', true);
          }
        } else {
          input.find(`option[value="${conditions[i].value}"]`).prop('selected', true);
        }
        input.trigger('updated');
      }
```

Closes https://github.com/infor-design/enterprise/issues/791

- Run http://localhost:4000/components/datagrid/test-save-settings.html
- Add some filter in header
- Reload the page
- Filter should still display after reloading


davidcarlsonberg:
A note that if you've hit this page before you may need to clear cache to see the newest version of the markup.

PR #858, opened 2018-09-26
Implementing Custom Builds

### LINKED ISSUE #417: Test and Document Custom Builds
**Original Issue:**  https://jira.infor.com/browse/SOHO-7579

We need to provide a system for slimming down IDS builds to only include the components an app developer wishes to implement.  Given a list of component names (fx: "toolbar, button, datepicker, form, validation"), it should be possible to generate a "custom" version of `sohoxi.js` with just the required components and their dependencies. 

- [x] Document the current process for making a custom build.
- [x] Clean up any imports to make it clean and simple.
- [x] Create a CLI-based tool for generating custom builds.
- [x] Explore custom builds for SASS files, linked to the Rollup bundler.
- [x] Should have test coverage for all individual dependencies.
- [x] Deprecate all mapper, and builder dependencies after review with stakeholders.
- [x] Should include tests. Specifically, a test build methods, and if possible the actual build, and the quality of it.

A good testing for this process should do the following:
- Create a custom build of either a single component or multiple components
- Segment out just the relevant functional/e2e tests for only those components and their dependencies
- generate a coverage report for the custom build

**References/Issues/Possible Blockers:**
https://github.com/rollup/rollup-plugin-multi-entry
https://github.com/rollup/rollup/issues/2049
https://github.com/rollup/rollup/issues/2092

```
// ================================================= /
// Soho Xi - Rollup Configuration
// ================================================= /
import resolve from 'rollup-plugin-node-resolve';
import babel from 'rollup-plugin-babel';
import json from 'rollup-plugin-json';
import license from 'rollup-plugin-license';
import multiEntry from 'rollup-plugin-multi-entry';

const bundleBanner = require('./build/generate-bundle-banner');
const getComponents = (listComponent) => {
  if (listComponent) {
    return listComponent.split(',');
  }

  return ['components/index.js'];
};
module.exports = {
  input: getComponents(process.env.ROLLUP_COMPONENTS),
  output: {
    file: 'dist/js/sohoxi.js',
    format: 'iife',
    sourcemap: true,
    sourcemapFile: 'dist/js/sohoxi.js.map',
    name: 'Soho',
    globals: {
      jquery: '$',
      d3: 'd3',
      alert: 'alert',
      console: 'console'
    }
  },
  plugins: [
    multiEntry({ exports: true }),
    resolve(),
    json(),
    babel({
      exclude: 'node_modules/**', // only transpile our source code
      plugins: ['external-helpers']
    }),
    license({
      sourceMap: true,
      banner: bundleBanner
    })
  ],
};
```

This seems to work, and shouldn't be breaking, again test are required for this modification.

```
env ROLLUP_COMPONENTS='components/dropdown/dropdown.js' npx grunt
env KARMA_SPECS='components/dropdown/unit/dropdown-api.spec.js' npm run ci:unit
```

EdwardCoyle:
This PR implements a custom build system for slimming down IDS builds to only what a developer requires.

**NOTE:** that this PR should be considered a first pass, and there are multiple improvements necessary that are [documented on the original issue](https://github.com/infor-design/enterprise/issues/417#issuecomment-424767286).

**ALSO NOTE:** this issue is not one that Q/A will currently be able to test.

Closes #417

Pull this branch, and run: 

```
npx grunt clean && npm run build:custom -- --components=button,input,listview,popupmenu,tabs-multi,validation
```
Ensure the following is true: 
- there are files generated to a `temp/` folder in the project root
- the paths in the `import/export` statements inside of each of those generated files make sense
- each CSS/JS file generated to the `/dist` folder only includes artifacts from the above list of components and their dependencies (might take some digging)

If you want to examine the demoapp, run `npm run quickstart` after the build completes.  The cleaning process gets rid of pre-built demoapp styles as well as the `dist/` folder.


EdwardCoyle:
@davidcarlsonberg thanks!  Good call on both of those but I've got some reasons:

- Those empty files have to exist during the sass/rollup bundling processes.  If they don't exist, the compilers will throw a fit
- All those extra components you're seeing are definitely dependencies.  We have to do a better job of breaking up components like Tabs so we don't need things like Searchfield, unless someone's specifically using the Module/Header Tabs, so that interdependencies like that don't happen (Evolution ticket?).

EdwardCoyle:
@tmcconechy @davidcarlsonberg, I pushed some changes that should address Tim's points.  The build script now includes running the vendor lib copy process.  I also fixed a couple really minor IDS style issues that were causing that `list` view in each folder not to render correctly with some of the layout styles missing from the bundle (they aren't required).

To test, make sure you run a `grunt clean` before running the build script.  I'll update the testing notes in the PR to reflect that.

Just one note about the demoapp layouts.  When using the demoapp to test custom bundles, we should make a habit of using the `No Frills` layout by opening a URL like this:  http://localhost:4000/components/validation/list?nofrills=true.  Pretty much all of our demoapp pages include the Header component, which contains a Searchfield, Toolbar, and Application Menu trigger.  None of those components are "core", so they won't necessarily be included inside custom bundles.  That can throw off testing when on the regular page layout, so just use the stripped down one instead.

PR #872, opened 2018-10-01
430 - listview performance issue

### LINKED ISSUE #430: SOHO-7307: Listview Performance issue when showing 4000 items
Ticket is intended to improve performance of list view and combines SOHO-8172

1. Using SOHO Listview to display 4000 items in the list 
causes slowness in rending the list, searching items, item selection/deselection and lag in scrolling.

2. Also, when transferring items in one listview to another, the scrollbar gets cut. 
Similar thing happens when deselecting multiple items at once. (see attached image)- This should be fixed on 4.8.0

3. When paging the entire list renders then the first page is re-rendered would be more efficient to simply render the first page initially once.

See attached sample application for case 1.

tjamesallen15:

The listview freezes when it contains 4000 data.

Closes #430 

Check test-listview-performance.html
Test the performance of the listview


EdwardCoyle:
I'm seeing the same things that @tmcconechy.

I wonder if we need to set a couple of goals surrounding this ticket.  What kinds of speed improvements are we looking for?

deep7102:
Yeah! its still bit in IE 11 if try to filter something. It wait takes much longer for each key stroke

tmcconechy:
I would say we should try to see if we can get loading time down to around 5 seconds or so in IE for 4000 items.

Then also see if the keyboard can be improved (i believe it probably can by instead changing the logic to not have a tabindex on any of them except the one that is the active one, when you arrow up or down remove the tabindex for the current one and add it on the next/previous one and focus it, this way we dont need to touch all the other ones)

tjamesallen15:
Thanks for all those comments guys! I will work on this more! I just only tried this on EDGE. That's the only thing available in my browserstack. The loading there is 3 - 5 seconds.

Any installers or way to test it on IE11? I'm using MacBook already.

tmcconechy:
You can also make a VM on your Mac. You can download here https://www.virtualbox.org/
Then you need to borrow a windows ISO and install it. Its useful to have. I can track down an ISO if you need it

clepore:
FYI - If we need to purchase browserstack testing accounts, let us know so we can find out.

clepore:
I'm going to close this PR as we are digging into the issue a bit deeper and conversing about it.

Issue #684, opened 2018-08-28
Show Mobile Device and Version on About page component


CindyMercadoReyes:
**Is your feature request related to a problem? Please describe.**
Show Mobile Device and Version on About page component

**Describe the solution you'd like**
Currently on the About page for mobile the information being displayed is being pulled from the browser. It would be great if the About page displays the device and version.

I'm attaching a screen shot of what is currently being displayed on the About page for mobile.

**Describe alternatives you've considered**
Reference below:
https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser

**Additional context**

![bs_realdroid_mobile_samsung galaxy s9 plus-8 0-1440x2960](https://user-images.githubusercontent.com/39101346/44729812-94f5c900-aaad-11e8-86b3-0dced1550b48.jpg)

EdwardCoyle:
Would need to extend the `Environment` core component to announce the device type.  Also note that what gets reported through the user agent string will not always 100% match the device type provided, using the method suggested.  Doing this correctly (or close to it) would need some research.

tmcconechy:
yeah ok, was thinking about using Environment as well , i not sure if we can get a version or not yet but we can definitely get the device name. Lets make it a 5 then.

EdwardCoyle:
@CindyMercadoReyes when you say "version", which do you mean?

- Device's hardware version (like Samsung Galaxy S8)
- Device's software version (like Android 8.1.0, or iOS 11.2)
- IDS version

Not sure if we could do the first one based on User Agent string... the other two we could make happen via the `Environment` core component the same way

ericangeles:
@tmcconechy definitely.

PR #913, opened 2018-10-03
684 - Added a device name via environment.js

ericangeles:

Extending `environment.js` and created `addDeviceSpecs()`.

Closes https://github.com/infor-design/enterprise/issues/684

- Run http://localhost:4000/components/about/example-index.html
- Click 'Show About Screen' button
- Should have an additional information about the device you are using.


ericangeles:
I researched about getting the hardware version via JavaScript but it seems that there's none. Some advise to use Platform.js (https://github.com/bestiejs/platform.js) but am not sure if we really needed it. Your thoughts?

tmcconechy:
Let's not add a third party lib for this. I suggest just doing what is possible and leaving it there as this is not critical, as its information only. So when you have spent a little time on it and the rest is not possible or taking too long. Let's just leave it where it is.

src/components/locale/cultures/es-ES.js
```diff
@@ -364,6 +366,7 @@ Soho.Locale.addCulture('es-ES', {
     UseEnter: { id: 'UseEnter', value: '. Para la búsqueda utilice Entrar o Flecha abajo.', comment: 'Instructional comments for screen readers' },
     User: { id: 'User', value: 'Usuario', comment: 'User tooltip' },
     UserProfile: { id: 'UserProfile', value: 'Perfil de usuario', comment: 'User Profile tooltip' },
+    Version: { id: 'Version', value: 'SOHO Versión', comment: 'Version of SOHO' },
```
tmcconechy: Call it IDS now not SOHO

PR #989, opened 2018-10-11
378 - Datagrid csv export

### LINKED ISSUE #378: SOHO-8122: Datagrid CSV export - issue with column separator & different locales
When exporting a datagrid to CSV format the file always uses a comma (",") for the field separator. In many locales the comma is the accepted default for the column separator character for .CSV files. For example, Excel will ,by default, directly open the file without issue on an en-US locale machine. But if a user 's machine is configured for a Swedish locale, for example, the default separator character is a semicolon (";"); a default open of the same file in Excel treats each row as a single field thus the data is corrupted. There are workarounds for this but a solution that works 'by default' for every user regardless of locale is preferred.

Here are a 3 proposed solutions:

1. Excel will accept a separator identification as the first line of an imported CSV file,  "sep=,". Solution would be to add that as a first line to every CSV export file. The downside of this is that a non-Excel app may have an issue with this but we're ok with it. There exists functionality in our application where we do CSV exports with our own code and this is what we do.
2. Allow for an optional parameter to be sent to the exportToCsv method that tells the datagrid code what character to use as the separator.
3. A combination of 1 & 2 - an optional parameter for the exportToCsv method that instructs the datagrid code to add the "sep=," as the first line in the exported file. This probably makes the most sense and won't break existing implementations.

The "sep=," appears in older versions of datagrid.js code, was specific for IE9.  

Open to other ideas of course.

Our application is used extensively to query data, sometimes lots of it, and the export function is used quite a bit so having this functionality work by default is important.

jmacaluso711:
Adds an optional parameter of "sep=," to the `exportToCsv` method. Fixes issue when a user's locale setting uses `;` as default separator in csv files. 

closes #378 

* change your machine's language setting to Swedish in System Preferences > Language & Region
* go to http://localhost:4000/components/datagrid/example-export-from-button.html 
* export to csv

### DIFF
--- a/src/utils/excel.js
+++ b/src/utils/excel.js
@@ -309,7 +309,7 @@ excel.exportToExcel = function (fileName, worksheetName, customDs, self) {
  * @param {string} self The grid api to use (if customDs is not used)
  * @returns {void}
  */
-excel.exportToCsv = function (fileName, customDs, self) {
+excel.exportToCsv = function (fileName, customDs, self, seperator = 'sep=,') {
   const formatCsv = function (table) {
     const csv = [];
     const rows = [].slice.call(table[0].querySelectorAll('tr'));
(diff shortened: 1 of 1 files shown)


dmcodrum:
I'm PTO until mid-next week. I can check it when I get back. Thanks Tim!

tmcconechy:
Ok, i'll merge and anything else we need to do to follow up we can. Seems like this is working to me.

dmcodrum:
I'm sure it's fine Tim. I will check and let you know. Thx again.

PR #991, opened 2018-10-11
Fix modal test-dynamic-content page console error.

davidcarlsonberg:
Chrome console error:
_Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-012427b8' http://squizlabs.github.io http://myserver.com". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution._

N/A

- Pull branch
- Run app
- Browse to http://localhost:4000/components/modal/test-dynamic-content
- Open Chrome Dev Tools
- Click "Open Modal"
- Select second option in Type input field ("This Value Shows More Content")
- Ensure there are no errors in Chrome Dev Tools Console

### DIFF
--- a/app/views/components/modal/test-dynamic-content.html
+++ b/app/views/components/modal/test-dynamic-content.html
@@ -103,7 +103,7 @@
       // Initial update of view
       update();
 
-      $('#context-type').on('change', () => {update()});
+      $('#context-type').on('change', function() {update()});
    </script>
 </div>
 </div>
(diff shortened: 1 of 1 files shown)


app/views/components/modal/test-dynamic-content.html
```diff
@@ -102,6 +102,8 @@ <h4>
 
       // Initial update of view
       update();
+
+      $('#context-type').on('change', () => {update()});
```
EdwardCoyle: @davidcarlsonberg we can't use arrow functions on the test pages.  If an IE11 user tries to use this page it will appear broken because ES6 isn't supported natively, and we don't polyfill it

PR #1015, opened 2018-10-16
822 - Added fix for memory leaks

### LINKED ISSUE #822: Memory leak in modal
**Describe the bug**
The modal component (modal.js) causes memory leaks by attaching key event handlers to the document that are never detached. This causes leaks of detached DOM elements when opening and closing modal dialogs. The keepFocus function in modal.js is called one or more time for each modal and each call attaches a key handler to the document object. After the modal dialogs are closed parts of the DOM elements are kept alive from the key event handlers.

The leaked DOM elements can be found by using the Chrome Memory profiler. They show up as "Detached HTMLDivElement" etc.

**To Reproduce**
1. Open and close modal dialogs a couple of times
2. Take a heap snapshot in the Chrome Memory profiler
3. There are several detached DOM elements kept alive through the key event handlers.

**Expected behavior**
Opening and closing modal dialogs should not cause leaks of detached DOM elements.

**Platform**
- Device: N/A
- OS Version: Windows
- Browser Name: Chrome
- Browser Version: 69.0.3497.92 (Official Build) (64-bit)

**Additional context**
The issue was found when using the enterprise components in Infor Homepages. The modals were used through the Angular SohoModalDialogService but the issue seems to be in the jQuery modal implementation.

### LINKED ISSUE #1032: Enter on colorpicker is submitting the form for modal
**Describe the bug**
The enter key when having a colorpicker in a modal closes the modal.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/colorpicker/test-modal.html
2. Click on the button and open the modal
3. Click into either color picker
4. type 1a1a1a and hit enter

**Expected behavior**
Should not close modal.

**Additional context**
Should be able to fix here: https://github.com/infor-design/enterprise/pull/915/files
Also need to adjust the test: Should tab to, and enter hex, and press enter, display value, and close modalhttps://github.com/infor-design/enterprise/blob/master/test/components/colorpicker/colorpicker.e2e-spec.js#L375

ericangeles:

Every time that the modal opens, the event listener on key event also trigger. So it's like, you're always adding up a listener for that element. Adding an `.off` will remove an event handlers that were attached with `.on()`, and detaching it every time the modal is open.

Closes #822
Closes #1032

- Run http://localhost:4000/components/modal/example-index.html
- Open and close modal dialogs a couple of times
- Take a heap snapshot in the Chrome Memory profiler
- There should be no key events handler in detached DOM elements.


ericangeles:
Adding the specific ID did fixed the nested modal, but the issue in memory leak gets back.

tmcconechy:
That seems strange, are you sure the id is the same both times? It should be just a namespace so should work i would think? If not can you push the change and i can check it out?

ericangeles:
Yup. Seems really strange.

tmcconechy:
Created related issue for the test https://github.com/infor-design/enterprise/issues/1032 can pick this up if its blocking..

Is the actual leak fixed on this PR? If not perhaps we close and reraise when it is?

ericangeles:
Will try to test it again if it's working now.

ericangeles:
It fixed the issue, but the nested modal breaks again.

tmcconechy:
@ericangeles i pushed an update, seems the id function is not being used so its not unique. Combined with yours and mine this seems good no.

PR #1029, opened 2018-10-17
412 - Fixed row activation index with tree datagrid

### LINKED ISSUE #412: SOHO-7730: Tree datagrid issue with row activation
With a mixed selectable grid option, there is an intermittent issue that exists were occasionally clicking on a row does not also make the row visually active (does not add the class 'is-rowactivated' to the tr element).

Click on the rows of the tree grid and eventually the issue occurs.  This is a problem for landmark in that we have a dependent child grid that is updated when the row click occurs on the parent grid.  There is a visual disconnect in that the active row of the parent grid will now not match what is being displayed in the dependent child grid. 

Seeing this in the angular-components.  

I've attached the following:
datagrid-treegrid.demo.html
datagrid-treegrid.demo.ts

deep7102:
With a mixed selectable tree grid row activation index was showing the wrong index.

Closes #412

http://localhost:4000/components/datagrid/test-tree-row-activated.html
- Open above link
- Open developer tools
- Click on cell where value is "6" in column "Id"
- See log "rowactivated" in console

```
there cold be some extra data too along this in console

rowactivated:
-row: 5,
-item:
--idx: 6,
--node:
---taskName: "Follow up action with HMM Global"
---id: 6
---desc: "Contact sales representative with the updated purchase order."
---time: "20:10 PM” 
```


deep7102:
Added functional and e2e tests

Issue #393, opened 2018-07-17
SOHO-8039: [All Browsers] Contextual Action Panel Control: Modal is not closing when clicking "Cancel"


clepore:
1. http://470-beta-enterprise.demo.design.infor.com/components/contextualactionpanel/example-workspaces.html

2. [All Browsers] Contextual Action Panel Control: Modal is not closing when clicking "Cancel" 

Steps to reproduce:
- Click Show Contextual Action Panel Button
-  Click Cancel
- The cancel button stays highlighted. Please see screen shot attached
Second Issue:
- When resizing the pop up there are alignment issues. Please see screen shot attached.

ericangeles:
It seems that the cancel button is working. Will proceed to the other issues such as, highlighting cancel button & alignment issue when resizing.

CindyMercadoReyes:
This issue is now resolved. Moving this ticket to the "Done" bucket.

PR #1057, opened 2018-10-23
393 - Contextual Action Panel Control

ericangeles:

It seems that the cancel button is working properly now.

The remaining issues:
1. Cancel Button is at highlighting state when modal pops up.
2. Alignment issue upon resizing.

_Fix (1. Cancel Button)_

In the `contextualactionpanel.js`, changing the `focus()` event to `blur()`, instead of removing the condition for selecting the first button element in the `.flex-toolbar`.

```
if (!selected.length && self.toolbar.is('.flex-toolbar')) {
  selected = self.toolbar.find('button').first();
  selected.blur();
  return;
}
```

_Fix (2. Alignment issue)_

Alignment issue is visible in the mobile view, at max width of 500px.

Since the title is a bit longer, what I did was to not display it on the mobile perspective at the max width of 480px and leave the icon. Pretty sure that the icon will be use on the functional side.
And also, I made some adjustments on the margin, padding to make the flex-toolbar align properly even when resizing.

Closes https://github.com/infor-design/enterprise/issues/393

- Run http://localhost:4000/components/contextualactionpanel/example-workspaces.html
- Click Show Contextual Action Panel Button
- Click Cancel (this should be fixed)
- Upon opening the modal, Cancel Button should be focused
- Resize the browser. Check all the elements if it's aligned correctly and should be responsive too.

### DIFF
--- a/src/components/contextualactionpanel/_contextualactionpanel.scss
+++ b/src/components/contextualactionpanel/_contextualactionpanel.scss
@@ -26,7 +26,8 @@
   }
 
   .flex-toolbar {
-    margin: 0 16px;
+    margin: 0 auto;
+    padding: 0 16px;
   }
 
   .toolbar {
--- a/src/components/contextualactionpanel/_contextualactionpanel.scss
+++ b/src/components/contextualactionpanel/_contextualactionpanel.scss
@@ -27,7 +27,7 @@
 
   .flex-toolbar {
     margin: 0 auto;
-    padding: 0 16px;
+    padding: 0 5px;
   }
 
   .toolbar {
--- a/src/components/contextualactionpanel/contextualactionpanel.js
+++ b/src/components/contextualactionpanel/contextualactionpanel.js
@@ -331,7 +331,7 @@ ContextualActionPanel.prototype = {
           }
           if (!selected.length && self.toolbar.is('.flex-toolbar')) {
             selected = self.toolbar.find('button').first();
-            selected.focus();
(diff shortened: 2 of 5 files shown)


src/components/contextualactionpanel/contextualactionpanel.js
```diff
@@ -331,7 +331,7 @@ ContextualActionPanel.prototype = {
           }
           if (!selected.length && self.toolbar.is('.flex-toolbar')) {
             selected = self.toolbar.find('button').first();
-            selected.focus();
+            selected.blur();
```
tmcconechy: I would revert this. This is just focusing the first button and its desired behavior despite what the bug says. So just ignore the highlight part of the bug.

src/components/contextualactionpanel/_contextualactionpanel.scss
```diff
@@ -26,7 +26,8 @@
   }
 
   .flex-toolbar {
-    margin: 0 16px;
+    margin: 0 auto;
+    padding: 0 16px;
```
tmcconechy: Is this just to get the focus state to appear? If so maybe 0 5px is enough or it changes the characteristics of the top toolbar.

src/components/contextualactionpanel/contextualactionpanel.js
ericangeles: Gotcha.

PR #1085, opened 2018-10-24
293 - Replaced `transitionend` event in Tabs in favor of renderLoop callback

### LINKED ISSUE #293: Tabs: fix scaffolding around `afteractivated` event / renderloop
**Describe the bug**

The `afteractivated` event fired by Tabs components doesn't always get fired properly.  The reason for this is it's being triggered at the end of a `transitionend` event.  While this event should be triggered the end of the fade-out animation of the `.tab-panel` element, in some cases such as the window losing focus, the event isn't triggered.

**To Reproduce**
This behavior is sporadic.  Steps to reproduce the behavior:
1. Go to http://localhost:4000/components/tabs/example-activated-event.html
2. Open this test in a code editor and change the event listener's condition from `activated.test` to `afteractivated.test`
3. In a browser, change the active tab with the keyboard and mouse
4. Occasionally, the `afteractivated` event will not fire.

**Expected behavior**

The `afteractivated` event should always be reliably triggered, even if the window loses/regains focus.

**Possible Solutions**

- Replace the [event listener for `transitionend`](https://github.com/infor-design/enterprise/blob/master/src/components/tabs/tabs.js#L1897) with a new RenderLoop Item (driven by `requestAnimationFrame`) that will more reliably trigger the event

**Screenshots**
n/a

**Platform**
n/a

**Additional context**

- A skipped test for this was added by @d-g-h here:  https://github.com/infor-design/enterprise/pull/271#pullrequestreview-132907278

EdwardCoyle:
This PR gets rid of a `transitionend` event listener in Tabs that was being used to fire a key lifecycle event, `afteractivated`.  The new behavior triggers this event after a `renderLoop` callback instead, which is more reliable than depending on the CSS transition.

Closes #293

- pull this branch, build, and run the demoapp
- open http://localhost:4000/components/tabs/example-activated-event.html
- ensure that a Toast appears notifying that the `afteractivated` event has fired (Tabs runs the actual activation process on init).
- Navigate to another tab with a click or a keypress.  Ensure that `beforeactivated`, `activated`, and `afteractivated` all fire in that order.

Additionally, you can run the Tabs functional test suite.  A new test has been added to cover this case.


EdwardCoyle:
@tmcconechy it will definitely add to that problem.  However I've got that issue assigned to me this sprint and we have a fix for it, so it won't be an issue for long.

PR #1091, opened 2018-10-25
 1079 - Lookup Page Indicators Hidden when Grid Width is small

### LINKED ISSUE #1079: Lookup Page Indicators Hidden when Grid Width gets smaller
**Describe the bug**
Hi, noticed an inconsistent behavior for the lookup control. When the grid size is smaller, some of the grid page indicators are being hidden. The indicators are hidden inconsistently. Please see steps/screenshots below for further explanation.

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/components/lookup/example-paging.html'
2. Don't open the lookup yet. In the console of the browser, enter this:

`$('#paging').data('lookup').settings.options.columns.pop(5);
$('#paging').data('lookup').settings.options.columns.pop(3);
$('#paging').data('lookup').settings.options.columns.pop(2);`

This removes some of the columns of the lookup.

3. Now open the lookup.
4. The "Records per Page" and "Page Number" indicators are hidden.
image

5. Click on the Next Page. The "Records per Page" is shown. The "Page Number" indicators are still hidden.
image

6. Now click on the previous page. The "Page Number" indicators are shown but the "Records per Page" are hidden.
image

Not sure if this expected, but the hidden indicators varies from page to page. Hope we can clarify if this is an issue or not. Thanks.

**Platform**
- Google Chrome

marknolaninfor:
Displaying a lookup that requires some stretching to display the full pager bar.

Closes #1079 

Open http://localhost:4000/components/lookup/example-paging.html
Click on the paging small lookup


tmcconechy:
There is still a minor lint problem preventing this from being merged:
https://travis-ci.com/infor-design/enterprise/jobs/154332447#L512

marknolaninfor:
I have now created a separate example page and also set the default value for the minwidth parameter, as the second lookup in the example does not have the value set.

Previously this would have been showing what the current behaviour is for a lookup that has a small grid width.

The new change will display the paging controls on a larger screen without the need to change the lookup definition, use the default. The strange behaviour may still happen on a smaller screen, under 400px wide as the pager control is trying to be responsive to the available space.

marknolaninfor:
Hi @EdwardCoyle,

Are you able to review the latest change on this pull request.

I have now created a separate example page and also set the default value for the minwidth parameter, as the second lookup in the example does not have the value set.

Previously this would have been showing what the current behaviour is for a lookup that has a small grid width.

The new change will display the paging controls on a larger screen without the need to change the lookup definition, use the default.

tmcconechy:
I assume it was accidental to close?

PR #1093, opened 2018-10-25
398 - Autocomplete beforeOpen

### LINKED ISSUE #398: SOHO-7995: Add a beforeOpen on autocomplete
I got a use case scenario like this:

_I would like to get the AutoComplete functionality ONLY if I start my search with “{“._
_If I use anything but that I want the input field to behave just like a normal input field._

If I had a way before the popup opens to determine if this is a right string or not would be the best case, such as a *beforeOpen.*

Case is used in App Builder.

tmcastro:
- Added autocomplete beforeOpen event
- http://localhost:4000/components/autocomplete/test-beforeopen-event.html

Closes #398

- Click on text field and type anything
- Check console log
- "beforeOpen" should be logged


src/components/autocomplete/autocomplete.js
```diff
@@ -204,6 +204,8 @@ Autocomplete.prototype = {
       return;
     }
 
+    this.element.triggerHandler('beforeOpen', [this]);
```
tmcconechy: Following our standards this should be `beforeopen`. Also i would add something like...

```
const canOpen = this.element.triggerHandler('beforeOpen', [this]);
if (canOpen === false) {
  return;
}
```

We usually allow this event to be vetoed. Also a functional test would be good :)

tmcastro:
Hi guys, I just updated the test page on this PR but the build failed. I tried reverting it and checking it locally but it's still giving me the same error.

PR #1110, opened 2018-10-26
1011 - undefined issue in dropdown fail [v4.12.x]

### LINKED ISSUE #1011: Dropdown with Icons: Searching in IE causes the icons to disappear and replaced with "undefined"
**Describe the bug**
Dropdown with Icons: Searching in IE causes the icons to disappear and replaced with "undefined".

**To Reproduce**
1. Go to http://localhost:4000/components/dropdown/example-icons.html in **Windows IE**
2. Type "icon" in the dropdown field
3. Notice that the icons will disappear and replaced by "undefined"

**Expected behavior**
Icons should not disappear when searching with the dropdown field

**Screenshots**
screen shot 2018-10-16 at 6 50 03 pm

**Platform**
 - Windows IE

**Additional context**
No additional context.

tjamesallen15:

Just a follow-up fix for the QA Test fail issue.

Closes #1011 

### DIFF
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -1057,7 +1057,15 @@ Dropdown.prototype = {
       const li = $(this);
 
       const text = a.text();
-      const icon = (li.children('a').find('svg').length !== 0) ? li.children('a').find('svg')[0].outerHTML : '';
+      let icon = '';
+
+      if (li.children('a').find('svg').length !== 0) {
+        if (li.children('a').find('svg')[0].outerHTML) {
+          icon = li.children('a').find('svg')[0].outerHTML;
+        } else {
+          icon = new XMLSerializer().serializeToString(li.children('a').find('svg')[0]);
+        }
+      }
 
       if (icon) {
         hasIcons = true;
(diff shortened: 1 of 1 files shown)


src/components/dropdown/dropdown.js
```diff
@@ -1057,7 +1057,15 @@ Dropdown.prototype = {
...
 
       const text = a.text();
-      const icon = (li.children('a').find('svg').length !== 0) ? li.children('a').find('svg')[0].outerHTML : '';
+      let icon = '';
+
+      if (li.children('a').find('svg').length !== 0) {
+        if (li.children('a').find('svg')[0].outerHTML) {
+          icon = li.children('a').find('svg')[0].outerHTML;
+        } else {
+          icon = new XMLSerializer().serializeToString(li.children('a').find('svg')[0]);
```
EdwardCoyle: @tjamesallen15 good solution for the missing `outerHTML` on SVG elements... I think you could safely just grab the HTML this way and not even worry about checking for the existence of outerHTML here.  SVG elements won't have it.

tjamesallen15:
Thanks for those replies @tmcconechy  @EdwardCoyle

PR #1117, opened 2018-10-29
1114 - Fix Editor error after calling updated [v4.12.x]

### LINKED ISSUE #1114: Editor throws exception: Cannot read property 'parentElement' of undefined when text is changed
**Describe the bug**
When the editor is configured to use the track dirty functionality and the editor.updated() method has been executed, changes to the editor text will throw an exception: Uncaught TypeError: Cannot read property 'parentElement'.

**To Reproduce**
1. Change /components/editor/example-dirty-tracking.html to run the editor.updated() method
2. Run /components/editor/example-dirty-tracking.html
3. Make a change to the text in the editor
4. See error in console 
5. Dirty indicator is not set on editor and a change.dirty event is not fired

**Expected behavior**
Expect the dirty indicator to be placed on the editor and a change.dirty event to be fired

**Screenshots**
image

**Additional context**
This is an issue for the soho-editor.component in ids-enterprise-ng as the updated() method is called by  ngAfterViewChecked(). 
The issue stems from the teardown() in editor.js as the textarea property is not initialized to null.
Also attached is an updated \editor\example-dirty-tracking.html that duplicates the issue.

[example-dirty-tracking.zip](https://github.com/infor-design/enterprise/files/2520546/example-dirty-tracking.zip)

tmcconechy:
When using dirty tracking and calling 'updated'. The editor will error out when typed in .

Closes #1114 

http://localhost:4000/components/editor/example-dirty-tracking.html
- type in the editor
- should show dirty tick and no errors in console.

### DIFF
--- a/app/views/components/editor/example-dirty-tracking.html
+++ b/app/views/components/editor/example-dirty-tracking.html
@@ -16,3 +16,8 @@
 
   </div>
 </div>
+<script>
+  $('body').one('initialized', function () {
+    $('#editor1').data('editor').updated();
+  });
+</script>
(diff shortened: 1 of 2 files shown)


fitzorama:
Did this change make it to the 4.12.0-rc.0? And if it did, how would we know?

tmcconechy:
Yes, it made 4.12.0-rc.0 but this had an issue so we immediately made 4.12.0-rc.1. You would only know by the dates of the releases based on the time it was merged onto the branch.

Issue #269, opened 2018-06-25
Complete Searchfield integration with Flex Toolbar


EdwardCoyle:
**Is your feature request related to a problem? Please describe.**

Flex Toolbar has basic support for barebones Searchfields, but it's support isn't currently 1-to-1 with the regular Toolbar.  We should complete the basic Searchfield handling of the Toolbar, and make sure that all alternates/contexts for the regular Toolbar in other components will work properly if a Flex Toolbar is used in its place.

**Describe the solution you'd like**

We'd need to add the following support to Flex Toolbar:

- [x] Responsive view for Searchfields inside the Flex Toolbar (making the `.toolbar-section.searchfield` expand and hide everything else on mobile views)
- [x] Smooth out the display of Collapsible Searchfields:
  - [x] on mobile views, hide Searchfields that are "collapsibleOnMobile" or "collapsible" behind a button if they aren't active.
  - [x] automatically expand the `.toolbar-section.buttonset` on desktop/larger views if the searchfield has focus.

Additionally, will fix some codepaths in these components that are specific to the regular toolbar, so that they can also support the alternate markup in the Flex Toolbar (f.x., `.toolbar-section` is an element that won't do anything in a regular Toolbar, but is integral to the function of Flex Toolbar.  We could use the class to define sections in both)

EdwardCoyle:
Scoped the part of this issue dealing with integration into larger components to #1120.  Going to pick the remaining part of this one up for the November sprint

EdwardCoyle:
Possibly looking at adding a piece of functionality to the Searchfield itself to make the responsive experience better:

- [ ] Create a responsive-only (collapse) button for the Flex Toolbar Searchfield
  - Only shows up on Responsive view (but exists all the time)
  - CSS hides it by default, prevents tabIndex by default.  In responsive view, it's focusable.
  - Clicking/Keying "Enter" on this button collapses the Searchfield and selects the next element on the toolbar.

EdwardCoyle:
Some of the above requires some async code to deal with focus traps... looking at adopting a promise polyfill in #1172 so we can actually use promises instead of jQuery deferreds for this

CindyMercadoReyes:
Hi @EdwardCoyle I'm moving this ticket to the Failed QA folder due to the toast not appearing for Win 10 IE 11 for URL: http://4130-beta0-enterprise.demo.design.infor.com/components/toolbar-flex/example-index.html#
[...]

PR #1148, opened 2018-11-02
269 - Toolbar Flex: Better searchfield integration

EdwardCoyle:
This PR addresses some of the points on #269:

- Adds a responsive view for an active searchfield residing inside a Flex Toolbar.  When active, the searchfield will expand to fill the space provided by the toolbar element and temporarily hide all the other items.  This replicates a feature present in the old toolbar.
- Completes the expand/collapse feature that was present on old Toolbar Searchfields on a Flex Toolbar.  When an "expanded" or "non-collapsible" searchfield is displayed below the tablet breakpoint (about 767px), it will collapse and appear as a search button until activated, which will cause it to expand to the newly added responsive view.

#269 

Pull this branch, build, run the demo app, then...

_To test responsive unopened state:_
- Navigate to http://localhost:4000/components/toolbar-flex/example-index?nofrills=true (don't test the regular view, since it's header still uses an old toolbar)
- Ensure all Searchfields on the page appear fully expanded with the correct background color
- Shrink the width of your browser viewport to less than 767px
- Ensure all Searchfields appear as a Search icon with a transparent background color.

_To test opened (focused) state:_
- With the browser viewport shrunk below 767px, focus one of the Search "icons"
- See that the other toolbar items disappear and the searchfield fills out to 100% of the width of the toolbar.
- Key in a search term.  Ensure that the clear button is in the correct spot.
- Defocus the searchfield.  Ensure that on the responsive view, no text is present on or near the search icon.
- Stretch your browser viewport to beyond 767px.
- Ensure that the searchfield has visible text entry, and it's gray background returns.


tmcconechy:
Some things are a bit funny with the keyboard. Try this.
- desktop size
- focus the link (SOHO-7475)
- tab in to toolbar
- go left arrow a few times
Result: The delete button is skipped and the focus stalls at the search field.

Similar problem at mobile size as you tab in , you can never get out of the search field.

Also for the animation would it look nicer if it animated out both ways when focus/clicked? Now it seems to move itself to the left and slide in one direction. This is just an idea, im not sure if its too hard or would actually look better.

EdwardCoyle:
The keyboard issue is definitely screwy.  The Flex Toolbar is weird with this because it won't let you key to buttons that are overflowed on purpose, and the buttons are all technically "overflowed" when the responsve searchfield is present, just due to being hidden.  

I think I'm going to fix this with a button that only shows up in the responsive searchfield that acts as a "collapse" trigger... we'll see how that works.

I'm not sure how much nicer I can make the animation.  The sections are all controlled by "flex grow" and "flex shrink", which don't animate too well.  I'm also trying to avoid adding Javascript based resizing of any sort, which caused a lot of the problems with the old expand/collapse feature in the toolbar.

tmcconechy:
Ok Dont worry about the animation then. Was just a suggestion.

EdwardCoyle:
There's also this one that I used to build it a bit:  http://localhost:4000/components/toolbar-flex/test-gauntlet.html

PR #1163, opened 2018-11-05
1147 - TextArea autogrow max height setting

### LINKED ISSUE #1147: TextArea autogrow max height setting
**Describe the bug**
The textarea can continue to grow to a size that may be too large to manage, eg a lot of scrolling.

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/components/textarea/example-autogrow.html'
2. Enter a large amount of text in to the autogrow textarea.

**Expected behavior**
That there was a way to limit the autogrow height to a manageable size.

**Screenshots**
The example screenshot shows a datagrid, that show the log for the active row .
image

**Additional context**
I have looked at adding an additional setting to the textarea, autoGrowMaxHeight, that will only apply the max height if defined on the control. This will then limit the height and display the scrollbars if the content is larger than required.

marknolaninfor:
The textarea can continue to grow to a size that may be too large to manage, eg a lot of scrolling.

Closes #1147 

1. http://localhost:4000/components/textarea/example-autogrow.html
2. Enter and remove a large amount of text into both of the text areas

The second textarea has the max height set so it will grow to the max height and then display a scollbar.


app/views/components/textarea/example-autogrow.html
```diff
@@ -6,5 +6,18 @@
       <textarea id="autogrow" class="textarea autogrow" name="autogrow" placeholder="Type your notes here..."></textarea>
     </div>
 
+    <div class="field">
+      <label for="autogrowMax">Notes (autogrow) Max Height</label>
```
tmcconechy: Use kebab case for ids  `autogrow-max`

PR #1223, opened 2018-11-15
1088 - Added hideMenuArrow setting to buttons

### LINKED ISSUE #1088: Provide a configuration option to hide drop-down arrow icon on a menu-button
**Describe the bug**

Using a `menu-button` for the "User" control on the masthead but don't want to see the drop-down arrow next to the User icon

Setting the `showArrow` to false did not have the desired affect.

**To Reproduce**

Add a menu button with the showArrow option set to false, and then display the page.

**Expected behavior**

The icon should be displayed without the arrow.

**Screenshots**

**Platform**
**Additional context**

I discussed this with @tmcconechy , and agreed that showArrow is the correct way to achieve this, however this does not currently work..

A corresponding change has been made in the enterprise-ng control to support showArrow.

jmacaluso711:
Adds option to hide menu arrow on menu buttons. 
Added example to http://localhost:4000/components/menubutton/test-default.html

closes #1088 

* pull branch
* got to http://localhost:4000/components/menubutton/test-default.html
* make sure second button example doesn't have `icon-dropdown`


clepore:
The one test failed here: https://travis-ci.com/infor-design/enterprise/jobs/159146585#L1890

tmcconechy:
The failing test is a fix im preparing on seperate PR that i mentioned in chat

src/components/button/button.js
```diff
@@ -68,6 +69,10 @@ Button.prototype = {
           $(this).find('span').text(a.text());
         });
       }
+
+      if (this.settings.hideMenuArrow) {
+        ddIcon.remove();
```
deep7102: Should add another test to make sure functionally (removing icon) is working. Currently added test with this PR only testing for setting options.

clepore: Would this be an e2e test, @deep7102 ?

deep7102: Yes, you can do e2e test or you can do just functional by checking the icon is on page or not

tmcconechy: Yes, to any and all tests 👍 If it can be a functional test that is nice because they are faster and dont need an extra page added to the page count to QA. Maybe a functional here as its an API method.

Issue #285, opened 2018-06-27
Contextual Action Panel Control High Contrast And Buttons Not Working Issues


CindyMercadoReyes:
**Describe the bug**
Gray spaces at the bottom of the contextual action panel when on High Contrast theme and Cancel and Submit buttons not working

**To Reproduce**
1. Go to url: 480-beta0-sohoxi.demo.design.infor.com/components/contextualactionpanel/example-workspaces.html
2. Switch to High Contrast theme under the (…) Menu
3. Click Show Contextual Action Panel button
4. Notice the gray spaces at the bottom of the contextual action panel
5. In addition, the "Cancel" and "Submit" button is not working. The user needs to click on the escape key to get out of the pop up window

**Expected behavior**
High contrast should appear correctly
The Cancel and Submit button should close the pop up window

**Screenshots**
<img width="1144" alt="contextual panel high contrast" src="https://user-images.githubusercontent.com/39101346/41991668-3fc75d28-7a14-11e8-82cf-e30ce592104d.png">

**Platform**

 - All Browsers 

**Additional context**
None

CindyMercadoReyes:
On WIn 10 IE 11, using URL: http://master-enterprise.demo.design.infor.com/components/contextualactionpanel/example-workspaces.html
The "Show Contextual Action Panel" does not pop up. The screen just turns a dark gray. Please see screen shot attached. 

![bs_win10_ie_11 0 1](https://user-images.githubusercontent.com/39101346/48374325-870a1a00-e692-11e8-8f65-783e922f22e8.jpg)

tmcconechy:
Seeing an error in the console: `Object doesnt support property or method `from`` maybe this is why?

ericangeles:
Agreed. I am seeing this error too in browserstack IE.

EdwardCoyle:
#1244 and #1252 should have this one totally fixed

tmcconechy:
I moved this to retest on the RC build as it does seem to be fixed on that branch but the fix does not seem to be in 4.13. beta 0 yet.

CindyMercadoReyes:
This issue is now resolved.

CindyMercadoReyes:
This issue is now resolved.

PR #1244, opened 2018-11-20
285 - Fix Array.from for IE10 and above

ericangeles:

It seems that `Array.from` method doesn't work and may not be present in other implementations of standards (IE10+, IE11 etc). I added a polyfill script that will allow/emulate `Array.from` method.

Closes https://github.com/infor-design/enterprise/issues/285

- Pull this branch
- Run http://localhost:4000/components/contextualactionpanel/example-workspaces.html in Windows IE10 browser or in browserstack.
- The "Show Contextual Action Panel" should pop up and not turn into dark gray background.

### DIFF
--- a/src/components/toolbar-flex/toolbar-flex.js
+++ b/src/components/toolbar-flex/toolbar-flex.js
@@ -23,6 +23,85 @@ function ToolbarFlex(element, settings) {
   this.element = element;
   this.settings = utils.mergeSettings(this.element, settings, TOOLBAR_FLEX_DEFAULTS);
 
+  // Array.from not supported in  IE6+ standards and Windows 8.1
+  // This code will emulate an ES6's Array.from method.
+  if (!Array.from) {
+    Array.from = (function () {
+      const toStr = Object.prototype.toString;
+      const isCallable = function (fn) {
+        return typeof fn === 'function' || toStr.call(fn) === '[object Function]';
+      };
+      const toInteger = function (value) {
+        const number = Number(value);
+        if (isNaN(number)) { return 0; }
+        if (number === 0 || !isFinite(number)) { return number; }
+        return (number > 0 ? 1 : -1) * Math.floor(Math.abs(number));
+      };
+      const maxSafeInteger = Math.pow(2, 53) - 1;
+      const toLength = function (value) {
+        const len = toInteger(value);
+        return Math.min(Math.max(len, 0), maxSafeInteger);
+      };
+
+      // The length property of the from method is 1.
+      return function from(arrayLike/* , mapFn, thisArg */) {
+        // 1. Let C be the this value.
+        const C = this;
+
+        // 2. Let items be ToObject(arrayLike).
+        const items = Object(arrayLike);
+
+        // 3. ReturnIfAbrupt(items).
+        if (arrayLike == null) {
+          throw new TypeError('Array.from requires an array-like object - not null or undefined');
+        }
+
+        // 4. If mapfn is undefined, then let mapping be false.
+        const mapFn = arguments.length > 1 ? arguments[1] : void undefined; //eslint-disable-line
+        let T;
+        if (typeof mapFn !== 'undefined') {
+          // 5. else
+          // 5. a If IsCallable(mapfn) is false, throw a TypeError exception.
+          if (!isCallable(mapFn)) {
+            throw new TypeError('Array.from: when provided, the second argument must be a function');
+          }
+
+          // 5. b. If thisArg was supplied, let T be thisArg; else let T be undefined.
+          if (arguments.length > 2) {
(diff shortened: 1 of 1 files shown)


src/components/toolbar-flex/toolbar-flex.js
```diff
@@ -23,6 +23,85 @@ function ToolbarFlex(element, settings) {
   this.element = element;
   this.settings = utils.mergeSettings(this.element, settings, TOOLBAR_FLEX_DEFAULTS);
 
+  // Array.from not supported in  IE6+ standards and Windows 8.1
```
tmcconechy: Did you consider just using array.slice to shallow copy rather than add all this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

I know we have a lot of instances in the code of slice or even using the JSON copy method so this is good to perhaps settle on one array copy method. Array.from could be the way but wanted to put it out there that there are a bunch of options.

@EdwardCoyle

src/components/toolbar-flex/toolbar-flex.js
```diff
@@ -23,6 +23,85 @@ function ToolbarFlex(element, settings) {
...
+  // This code will emulate an ES6's Array.from method.
+  if (!Array.from) {
```
EdwardCoyle: If we need to polyfill `Array.from`, we shouldn't be doing it this way.  I can see us wanting to use it elsewhere, so we should modularize it and have it available to any component via ES6 imports.  I would say, take this code and:

- Remove it from Flex Toolbar
- Create a new folder in `/src` called "polyfills"
- Add a file called `array.from.js` in `/src/polyfills`
- Add the below lines 29 thru 102 in that file
- At the top of this file, use the following along with the other imports:

```
import '../../polyfills/array.from';
```

Alternatively, we could import from an external package like [this one](https://www.npmjs.com/package/array-from-polyfill).  I don't mind managing the source internally for something like this though.

@tmcconechy @deep7102 @davidcarlsonberg what do you guys think?

src/components/toolbar-flex/toolbar-flex.js
```diff
@@ -99,12 +178,6 @@ ToolbarFlex.prototype = {
     $(this.element).on(`selected.${COMPONENT_NAME}`, (e, ...args) => {
       log('dir', args);
     });
-
-    // Inlined Searchfields can cause navigation requiring a focus change to occur on collapse.
-    $(this.element).on(`collapsed-responsive.${COMPONENT_NAME}`, (e, direction) => {
```
EdwardCoyle: This was just added in #1148... maybe you need a merge from `master`?

EDIT: Actually pretty much any of the code being removed in this PR is stuff that was just added.  I think you definitely want to merge master before this gets any further.

src/components/toolbar-flex/toolbar-flex.js
EdwardCoyle: @tmcconechy I'm not opposed to the polyfill personally, but we've gotta separate the polyfill from the Flex Toolbar source code, I think

EdwardCoyle: @tmcconechy also the original Flex Toolbar code uses `Array.from()` because it's converting a [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) to an Array, so if you tried using `Array.slice()` on the result instead, it wouldn't work.  It needs to be converted first.

tmcconechy:
I think maybe we should just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice for this case and save a few hundred lines of code and a poly fill. If that works.

EdwardCoyle:
@ericangeles we realized after some testing today that the Flex Toolbar was completely unusable in IE in the current version of IDS because of the usage of `Array.from()`.  Rather than wait to have it fixed in `master`, #1252 that will fix in the Q/A branch (`4.13.x`).  We will probably have to skip merging this PR... Sorry for the confusion on this

ericangeles:
@tmcconechy @EdwardCoyle thanks for all the suggestions! To conclude, are we gonna hold this at the moment?

EdwardCoyle:
Yes, for now let's leave this one alone.  It may no longer be an issue once we get #1252 merged

tmcconechy:
More specifically.

- close this
- wait and see about https://github.com/infor-design/enterprise/pull/1252 merged
- make a new clean branch off 4.13.x
- redo the fix by changing array.from to the new method Ed added `utils.getArrayFromList` which will work fine on IE.

EdwardCoyle:
@ericangeles #1252 actually fixed the whole issue with Flex Toolbar so there should be no need to redo your issue.  We can close that as well

PR #1266, opened 2018-11-22
1056 - Fix for X button Alignment Issue 2 [v4.13.x]

### LINKED ISSUE #1056: Tabs: On mobile, alignment issue on X button of dismissible tab with dropdown
**Describe the bug**
Tabs: On mobile, alignment issue on X button of dismissible tab with dropdown .

**To Reproduce**
1. Go to http://localhost:4000/components/tabs/test-dismissibles-with-dropdowns.html on Android or iOS devices
2. Click the arrow down on the right-most part of tab
3. Notice that the X button is out of position in "Plastic (Dismissible & Dropdown)" tab

**Expected behavior**
X button in the dismissible tab with dropdown should be in the right position.

**Screenshots**
screen shot 2018-10-23 at 6 53 24 pm

**Platform**
 - Android and iOS devices

**Additional context**
No additional context.

tjamesallen15:

Tweaking the position of "X" Button.

Closes #1056 

- open ios simulator
- go to http://localhost:4000/components/tabs/test-dismissibles-with-dropdowns.html

### DIFF
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -27,8 +27,19 @@
   }
 
   &.bottom {
-    #tab-container-popupmenu {
+    .popupmenu {
       .dismissible {
+        .wrapper {
+          li {
+            a {
+              .icon {
+                &.close {
+                  right: 6px;
+                }
+              }
+            }
+          }
+        }
         .icon.close {
           height: 15px;
           padding-left: 6px;
@@ -36,15 +47,6 @@
           position: absolute;
           right: 17px;
           top: 2px;
-        }
-      }
-    }
-
-    .wrapper {
-      .dismissible {
-        .icon.close {
-          padding: !important;
-          right: 6px !important;
         }
       }
     }
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -48,7 +48,7 @@
         }
       }
     }
-    
+
     .arrow {
       border-bottom-color: $popupmenu-border-color;
       //left: auto;
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -40,7 +40,7 @@
             }
           }
         }
-        
+
         .icon.close {
           height: 15px;
           padding-left: 6px;
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -27,6 +27,28 @@
   }
 
   &.bottom {
+    #tab-container-popupmenu {
+      .dismissible {
+        .icon.close {
+          height: 15px;
+          padding-left: 6px;
+          padding-right: 6px;
+          position: absolute;
+          right: 17px;
+          top: 2px;
+        }
+      }
+    }
+
+    .wrapper {
+      .dismissible {
+        .icon.close {
+          padding: !important;
+          right: 6px !important;
+        }
+      }
+    }
+    
     .arrow {
       border-bottom-color: $popupmenu-border-color;
       //left: auto;
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -40,6 +40,7 @@
             }
           }
         }
+        
         .icon.close {
           height: 15px;
           padding-left: 6px;


src/components/popupmenu/_popupmenu.scss
```diff
@@ -27,6 +27,28 @@
...
+          padding-left: 6px;
+          padding-right: 6px;
+          position: absolute;
+          right: 17px;
+          top: 2px;
+        }
+      }
+    }
+
+    .wrapper {
```
ericangeles: Instead of using `!important`, you might probably do is,

```
.wrapper {
      .dismissible {
        .wrapper {
          #popupmenu-2 {
            li {
              a {
                .icon {
                  &.close {
                    right: 6px;
                  }
                }
              }
            }
          }
        }
      }
    }
```

src/components/popupmenu/_popupmenu.scss
```diff
@@ -27,6 +27,28 @@
   }
 
   &.bottom {
+    #tab-container-popupmenu {
+      .dismissible {
+        .icon.close {
```
ericangeles: Since we're using scss, i would recommend to stick with their structure.

instead `.icon.close`, make it,

```
.icon {
  &.close {
  }
}
```

src/components/popupmenu/_popupmenu.scss
```diff
@@ -27,6 +27,28 @@
...
+          right: 17px;
+          top: 2px;
+        }
+      }
+    }
+
+    .wrapper {
+      .dismissible {
+        .icon.close {
+          padding: !important;
```
ericangeles: Not sure what's with this padding. You should probably remove this.

src/components/popupmenu/_popupmenu.scss
EdwardCoyle: The other thing I'd add here is that we should avoid targeting HTML ID attributes if we can.  The ID being targeted here isn't guaranteed to exist in someone's application.

clepore: Some context - targeting the `id` attribute should be avoided in css just like `!important`. It throws off selector specification, by prioritizing out of cascading order. Also nesting `id` selectors should be avoided entirely as good practice due to the way browsers interpret the selector chain.

EdwardCoyle: I'd like to add:  We should avoid using HTML IDs in general (even though this one is generated internally).  Our JS code that generates these IDs is subject to change, and really shouldn't be in the business of determining those IDs at all, leaving it to the application developers that need to implement IDS.  Any new CSS written going forward should do targeting with CSS classes, and in some rare cases, specific attributes.

More importantly, this type of icon alignment should be setup against all Popupmenus, and shouldn't be scoped to just the tabs spillover menu, since this could potentially be used outside of the tabs.  

For example, these Dropdown Tab menus now have a different size/padding/alignment for the normal "close" icon when the `#tab-container-popupmenu` ID isn't present:

screen shot 2018-11-26 at 10 04 52 am

... versus...

screen shot 2018-11-26 at 10 04 57 am

src/components/popupmenu/_popupmenu.scss
```diff
@@ -27,6 +27,28 @@
...
+          right: 6px !important;
+        }
+      }
+    }
+
     .arrow {
       border-bottom-color: $popupmenu-border-color;
       //left: auto;
```
tmcconechy: Can remove this as well..

src/components/popupmenu/_popupmenu.scss
tmcconechy: Instead of the !important sometimes all you need to do is make sure the entire rule is matched all the way.

tjamesallen15:
Hi @tmcconechy,

I'm having issues on debugging the browserstack. jQuery also is not working when trying in the developer's console.

See the picture below.

screen shot 2018-11-28 at 1 44 21 pm

I input the "debugger;" inside the showSubmenu(li) method.

screen shot 2018-11-28 at 1 50 25 pm

Note: I wasn't the only one experiencing this. Terrence also encountered this issue. I also tried to research on the web about this. :(

tjamesallen15:
I will try your suggestion, @tmcconechy! Thanks!!!

tmcconechy:
@tjamesallen15 im going to push this to next sprint so that you can get this fixed up. You can redo a PR on master for this once your ready with a fix.

PR #1271, opened 2018-11-26
1238 - Fixed issues regarding lookup being hidden on tabs [v4.13.x]

tmcastro:
- Fixed issues regarding lookup being hidden on tabs
- Added same fix on date picker
- Added test page
- http://localhost:4000/components/lookup/test-tabs.html

Closes #1238 

- Click on first tab to display content with lookup
- Lookup icon should now be visible and working correctly

### DIFF
--- a/src/components/datepicker/datepicker.js
+++ b/src/components/datepicker/datepicker.js
@@ -154,7 +154,7 @@ DatePicker.prototype = {
     this.trigger = $.createIconElement('calendar').insertAfter(this.element);
 
     // Hide icon if datepicker input is hidden
-    const isHiddenComponent = this.element.parents('.modal').length > 0 && this.element.parents('.tab').length > 0;
+    const isHiddenComponent = this.element.parents('.modal').length > 0 && this.element.parents('.tab-panel').length > 0;
     if ((this.element.is(':hidden') || this.element.css('visibility') === 'hidden') && !isHiddenComponent) {
       this.trigger.hide();
     }
--- a/src/components/datepicker/datepicker.js
+++ b/src/components/datepicker/datepicker.js
@@ -154,9 +154,8 @@ DatePicker.prototype = {
     this.trigger = $.createIconElement('calendar').insertAfter(this.element);
 
     // Hide icon if datepicker input is hidden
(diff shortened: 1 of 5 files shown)


src/components/lookup/lookup.js
```diff
@@ -131,7 +131,8 @@ Lookup.prototype = {
     lookup.after(this.icon);
 
     // Hide icon if lookup input is hidden
-    if ((lookup.is(':hidden') || lookup.css('visibility') === 'hidden') && lookup.parents('.modal').length === 0) {
+    const isHiddenComponent = lookup.parents('.modal').length > 0 || lookup.parents('.tab').length > 0;
+    if ((lookup.is(':hidden') || lookup.css('visibility') === 'hidden') && !isHiddenComponent) {
```
EdwardCoyle: I'm seeing this get hidden when the page loads, but I'm never seeing the icon re-appear, which causes the Lookup not to work on these tabs at all.

nov 26 2018 10_36 am

tmcconechy:
An alternative here as noted https://jira.infor.com/browse/SOHO-6549?focusedCommentId=4199498&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4199498 Andy is open to using a class to determine hidden.

If thats needed we can change this logic to only initialize when the hidden class is there.

src/components/datepicker/datepicker.js
```diff
@@ -154,7 +154,8 @@ DatePicker.prototype = {
     this.trigger = $.createIconElement('calendar').insertAfter(this.element);
 
     // Hide icon if datepicker input is hidden
-    if ((this.element.is(':hidden') || this.element.css('visibility') === 'hidden') && this.element.parents('.modal').length === 0) {
+    const isHiddenComponent = this.element.parents('.modal').length > 0 && this.element.parents('.tab-panel').length > 0;
```
tmcconechy: Im wondering about this now. What if they want to init a field on a tab as hidden.
I think we should change this to ONLY work with the hidden class on the input.

fitzorama:
@tim We are hoping this can be reviewed and merged today as this has become a high priority item for Landmark. We have a lot of lookup and date fields on tabs that are missing their icons. Right now it is only affecting the developers, but the QAs could pick up a build without this change.

PR #1275, opened 2018-11-26
379 - Add sanitizeHTML option to message

### LINKED ISSUE #379: SOHO-8111: Provide the ability to do some formatting in the Message component
Would like to be able to maybe bold something and also do a line feed. 

For example perhaps leave it up to the dev whether html is stripped or not, and that way I can use html in the message and do my own encoding/sanatize.
 
message: "This is my own text <b><br>" + htmlEncode(userprovidedtext) + "</b>"
 
Or have the bold tag and line break tab not be stripped.

So and modifiecation that will allos for some basic formating like bold/emaphsize and line feeds/breaks

Use Case:

Use case would be a Confirm Delete as title. Then... 

Are you sure you want to delete the <b>Role Name</b> security role?<br>You must delete this role in Mingle as well.

as the message.

davidcarlsonberg:
Allows a dev to turn on and off sanitization of text in Message title and Message message.

Closes #379 

- Pull branch.
- Run app.
- Browse to http://localhost:4000/components/message/example-index
- Ensure text is sanitized or not based on inclusion of option in Message.
- Maybe try a new one for yourself

### DIFF
--- a/app/views/components/message/example-index.html
+++ b/app/views/components/message/example-index.html
@@ -19,7 +19,8 @@
         title: '<span>Application Error</span>',
         status: 'error',
         returnFocus: $(this),
-        message: 'This application has experienced a system error due to the lack of internet access. Please restart the application in order to proceed.',
+        message: 'This application has experienced a <b>system error</b> due to the lack of internet access.<br><em>Please</em> restart the application in order to proceed.',
+        sanitizeHTML: false,
         buttons: [{
             text: 'Restart Now',
             click: function() {
@@ -74,7 +75,7 @@
     $('#show-fileupload-confirmation').on('click', function() {
       $('body').message({
         title: 'File Upload Complete',
-        message: 'Your file "<b>photo.png</b>" was successfully uploaded to your personal folder and is now public for viewing.',
+        message: 'Your file "<b>photo.png</b>" was successfully uploaded to your personal folder and is now <em>public for viewing</em>.',
         returnFocus: $(this),
         buttons: [{
           text: 'Done',
@@ -90,8 +91,9 @@
     $('#show-delete-confirmation').on('click', function() {
       $('body').message({
         title: 'Delete this Application?',
-        message: 'You are about to delete this application permanently. Would you like to proceed?',
+        message: 'You are about to delete this application permanently.<br>Would you like to proceed?',
         returnFocus: $(this),
+        sanitizeHTML: false,
         buttons: [{
           text: 'Yes',
           click: function(e, modal) {
--- a/src/components/message/message.js
+++ b/src/components/message/message.js
@@ -28,7 +28,8 @@ const MESSAGE_DEFAULTS = {
   width: 'auto',
   buttons: null,
   cssClass: null,
-  returnFocus: null
+  returnFocus: null,
+  sanitizeHTML: true
 };
 
 function Message(element, settings) {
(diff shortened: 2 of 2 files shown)


app/views/components/message/example-index.html
```diff
@@ -19,7 +19,8 @@
         title: '<span>Application Error</span>',
         status: 'error',
         returnFocus: $(this),
-        message: 'This application has experienced a system error due to the lack of internet access. Please restart the application in order to proceed.',
+        message: 'This application has experienced a <b>system error</b> due to the lack of internet access.<br><em>Please</em> restart the application in order to proceed.',
+        sanitizeHTML: false,
```
EdwardCoyle: Just wondering if this is something we want to be optional?  Seems like we'd always want this content run through the sanitizer as a best practice?

Also, @tmcconechy @clepore one of the goals listed for this ticket was to choose tags to be whitelisted.  Are we still doing that?

clepore: I'd vote always sanitize, and I believe we mentioned something somewhere about choosing tags, but I do not know in which medium.

davidcarlsonberg: So we should be stripTags instead of stripHTML with some options. I see Tim included a list from W3Schools in the issue...whitelist all those?

src/components/message/message.js
```diff
@@ -28,7 +28,8 @@ const MESSAGE_DEFAULTS = {
   width: 'auto',
   buttons: null,
   cssClass: null,
-  returnFocus: null
+  returnFocus: null,
+  allowHTMLTags: false
```
tmcconechy: Not against making this a setting. But i think it is maybe unnecessary. What do you think folks? Without this it can just "work" without knowing about a setting and the tags are that harmful.

src/components/message/message.js
```diff
@@ -48,8 +49,8 @@ Message.prototype = {
     // Create the Markup
     this.message = $('<div class="modal message"></div>');
     this.messageContent = $('<div class="modal-content"></div>');
-    this.title = $(`<h1 class="modal-title" id="message-title">${xssUtils.stripHTML(this.settings.title)}</h1>`).appendTo(this.messageContent).wrap('<div class="modal-header"></div>');
-    this.content = $(`<div class="modal-body"><p class="message" id="message-text">${xssUtils.stripHTML(this.settings.message)}</p></div>`).appendTo(this.messageContent);
+    this.title = $(`<h1 class="modal-title" id="message-title">${this.settings.allowHTMLTags ? xssUtils.stripTags(this.settings.title, '<b><strong><i><em><mark><small><del><ins><sub><sup>') : xssUtils.stripHTML(this.settings.title)}</h1>`).appendTo(this.messageContent).wrap('<div class="modal-header"></div>');
+    this.content = $(`<div class="modal-body"><p class="message" id="message-text">${this.settings.allowHTMLTags ? xssUtils.stripTags(this.settings.message, '<b><strong><i><em><mark><small><del><ins><sub><sup>') : xssUtils.stripHTML(this.settings.message)}</p></div>`).appendTo(this.messageContent);
```
tmcconechy: Can we also allow <br> as that was mentioned on the ticket.

src/components/message/message.js
davidcarlsonberg: I think it would just allow stricter control for the adopting dev. It defaults to stripping all, so no harm in not knowing about the setting. It’s a bit of a toss up. Happy to go either way.

tmcconechy: yeah i could go either way too. just add a comment to the top section so it appears in the docs above and lets leave it then. We will have to add an angular option too later but thats fine.

tmcconechy: I mean here for the docs https://github.com/infor-design/enterprise/blob/7dce65f12a590d54ca78bdf061158d989ee47f9e/src/components/message/message.js#L22

PR #1285, opened 2018-11-26
1269 -  Datagrid applyfilter does not work with related datafields

### LINKED ISSUE #1269: Datagrid applyfilter does not work with related datafields
**Describe the bug**
The logic to set the rowValue within apply filter is not correct and will not failback to the fieldValue function when it need to evaluate any related data.

The test is only if rowData exists, and not that also rowData[columnDef.field] does not return an undefined value. Our example has the columnDef.field set as 'csysmpt.descr'
```
let rowValue = rowData ? rowData[columnDef.field] :
          self.fieldValue(rowData, columnDef.field);
```
Correct check will then fail to the function, as rowData['csysmpt.descr'] will return undefined.
```
let rowValue = rowData && rowData[columnDef.field] !== undefined ? rowData[columnDef.field] :
          self.fieldValue(rowData, columnDef.field)
```
**To Reproduce**

**Expected behavior**
The rowValue is set correctly so that the datagrid can be filtered.

**Version**
v4.14

**Screenshots**
Example of rowData to filter the csysmpt.descr field
image

marknolaninfor:
Corrected the logic on applyfilter to return the correct rowValue.

Closes #1269 

Goto 'http://localhost:4000/components/datagrid/test-filter-editable.html'
Filter on Price, Quantity, or Order Date.

### DIFF
--- a/app/views/components/datagrid/test-filter-editable.html
+++ b/app/views/components/datagrid/test-filter-editable.html
@@ -13,14 +13,14 @@
         data = [];
 
       // Define Some Sample Data
-      data.push({ id: 1, productId: 2142201, productName: 'Compressor', inStock: true, activity:  'Assemble Paint', quantity: 1, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action', status: 'Error'});
-      data.push({ id: 2, productId: 2241202, productName: '1 Different Compressor', inStock: true, activity:  'Inspect and Repair', quantity: 2, price: 210.99, status: '', orderDate: new Date(2015, 7, 3), action: 'On Hold', status: 'Error'});
-      data.push({ id: 3, productId: 2342203, productName: 'Compressor', inStock: true, activity:  'Inspect and Repair', quantity: 1, price: 120.99, status: null, orderDate: new Date(2014, 6, 3), action: 'Action', status: 'Error'});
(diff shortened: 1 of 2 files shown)


app/views/components/datagrid/test-filter-editable.html
```diff
@@ -33,7 +33,13 @@
       columns.push({ id: 'productId', name: 'Product Id', field: 'productId', formatter: Formatters.Readonly, filterType: 'text'});
       columns.push({ id: 'productId2', name: 'Product Id 2', field: 'productId', formatter: Formatters.Readonly, filterType: 'text'});
       columns.push({ id: 'productName', name: 'Product Name', field: 'productName', required: true, validate: 'required', formatter: Formatters.Hyperlink, filterType: 'text', editor: Editors.Input});
-	  columns.push({ id: 'orderDate', name: 'Order Date', field: 'orderDate', width: 130, formatter: Formatters.Date, dateFormat: 'MM/dd/yyyy', filterType: 'date', editor: Editors.Date});
+	    columns.push({ id: 'quantity', name: 'Quantity', field: 'order.quantity', align: 'right', editor: Editors.Input, filterType: 'number', mask: '###', isEditable: function (row, cell, value, col, item) {
```
tmcconechy: I think filterType should be `'integer'  or  'decimal'` this is causing an error when you open the filter button on this field. And why the menu doesnt open as Ed noted.

PR #1291, opened 2018-11-27
379 - Refactor message HTML tag allowance

davidcarlsonberg:
Creates a setting to allow customization of allowed HTML tags in a message. Default tags are `<a><b><br><br/><del><em><i><ins><mark><small><strong><sub><sup>`. Empty string results in all tags being stripped. Adds a test to ensure all tags are stripped if empty string is used.

References #379 .
Supersedes #1275 .

- Pull branch
- Run app
- http://localhost:4000/components/message/example-index
- Click Allow Tags button. Tags should format message (looks like a ransom note).
- Dismiss message.
- Click Disallow Tags button. Tags should be stripped and no formatting appear in message.


davidcarlsonberg:
Adding another test to check if stripping properly works.

PR #1350, opened 2018-12-05
1327 - Allow the demo app to use more than the nofrills custom layout

### LINKED ISSUE #1327: Display theme tabs for embedded demos
People want to visually see dark and contrast examples of the embedded demos on Enterprise components.

Use tabs to provide options to switch between different themes of that same demo.

https://design.infor.com/code/ids-enterprise/latest/demo/components/tabs/example-index?font=source-sans

Maybe a dropdown instead? But I think tabs would be better.

clepore:
We wanted a way to switch themes for the demos in the IDS website.

Closes #1327 

1. Start / Restart your local server for the new routing
1. Goto http://localhost:4000/components/alerts/example-index.html?layout=nofrills
    - make sure the nofrills layout works as it always as
1. Goto http://localhost:4000/components/alerts/example-index.html?layout=embedded
    - You should be able to switch themes by clicking the theme nav buttons and see:
        screen shot 2018-12-05 at 1 47 23 pm

**Also** there are probably things that I do not know about that could have broke as the `nofrills` part is embedded pretty deep. _Please_ spend some time and test anything you can think of that I may have missed (i.e. patterns, tests, etc).


clepore:
@tmcconechy I change the app to use a `layout=` param now so that you can do `nofrills` or `embedded` (the new one I added for the IDS website). I think it is a bit more clean, but please run it through the gauntlet.

app/views/layouts/layout.html
```diff
@@ -1,8 +0,0 @@
-{{> includes/head}}
```
tmcconechy: I think we should keep https://github.com/infor-design/enterprise/blob/master/app/views/layouts/layout.html 

Because all the pages on here http://localhost:4000/layouts/list are messed up / have double header ect now. Otherwise all these should be updated to put that same layout in each example.

FYI nofrills didnt work here before - as was not needed but maybe we want embedded too? Or its fine to wait until we make this a component.

tmcconechy: Examples also have double headers fx http://localhost:4000/examples/lawson/hotkeys.html
See some of http://localhost:4000/examples

clepore: I delete this file completely, so not sure what you mean here...?

EdwardCoyle:
We should probably find a way to detect if the layout exists or not, and default to the standard one that it would otherwise load if it doesn't.

For example, running http://localhost:4000/components/dropdown/example-index?layout=notalayout currently errors out.  Should probably just ignore the param in that case.  We don't have anything that scans the filesystem for a matching layout at the moment so we'll need to write something.

clepore:
Due to the nature of the timing, we are going to PR this into `4.13.x`

EdwardCoyle:
Raised #1357 for the bugfix for later

PR #1362, opened 2018-12-10
1070 - Fix for page skips and zoom issue on Spinbox component

### LINKED ISSUE #1070: Form Spinbox Component: The page skips and the text gets zoomed in each time you click on the "-" or "+" button
**Describe the bug**
Form Spinbox Component: The page skips and the text gets zoomed in each time you click on the "-" or "+" button

**To Reproduce**
1. On iOS Iphone X
2. Go to url: http://4120-beta0-enterprise.demo.design.infor.com/components/form/test-spinbox-compound-field.html
2. Click on the "-" or "+" button for the Age 2 spinbox
3. The page skips and the text gets zoomed in each time you click on the "-" or "+" button

**Expected behavior**
The page should not skip nor should it zoom in each time the user clicks on minus or plus symbol 

**Screenshots**
None
**Platform**
 - Device: iOS iPhone X 

**Additional context**
None

tjamesallen15:
Page skips and zoom issues on Spin Box Component

Closes #1070 

Go to http://localhost:4000/components/form/test-spinbox-compound-field.html on iphone X
Try clicking the Age 2 Spin Box Increase Value Button


src/components/spinbox/spinbox.js
```diff
@@ -510,7 +510,6 @@ Spinbox.prototype = {
   updateVal(newVal) {
     this.element.val(newVal).trigger('change');
     this.updateAria(newVal);
-    this.element.focus();
```
tmcconechy: I think it would be ok if we did this on the click event to focus but i dont thinnk this should be here. This will cause the side effect of if updating the value with the API it will focus the input.

I dont think people would expect that, maybe see if it works in the click event. https://github.com/infor-design/enterprise/blob/cc220758932cf661e98b2cc9390db403ba5de6ef/src/components/spinbox/spinbox.js#L295

Or alternative  maybe something can be preventDefault on in the touch / click events.

EdwardCoyle: Agreed on needing a different fix... @tjamesallen15 are we sure the issue isn't what I described in [my comment on the original issue](https://github.com/infor-design/enterprise/issues/1070#issuecomment-435968490)?  If it's not that, maybe try what Tim suggested?

tjamesallen15: @EdwardCoyle yes. I already checked the zoom component that's why I tried that way. I remove that focus line because it always been called each and every time the button is clicked. I'll try first @tmcconechy suggestion.

tjamesallen15:
@tmcconechy  Try testing it again. I just added preventDefault. The zooming will not be triggered anymore.

tmcconechy:
This fix works well with the zoom. 

1. But didnt it focus the field before on click? I think it shoul focus when you click the + or -
2. And very minor but i think the inputs should be width: 100% on this breakpoint so that the fields match up widths. Can we see why?

screen shot 2018-12-11 at 11 32 58 am

src/components/spinbox/spinbox.js
```diff
@@ -83,6 +84,11 @@ Spinbox.prototype = {
     const self = this;
     const val = self.checkForNumeric(self.element.val());
 
+    // Prevents zoom feature in spinbox to prevent zooming every click on spinbox
+    if (env.os.name === 'ios' || env.os.name === 'android') {
+      $('head').triggerHandler('disable-zoom');
```
EdwardCoyle: This is on the right track.  However, using this here will permanently disable zooming.  We don't necessarily want to do that, because the iOS zoom is an accessibility feature, and we want the user to be able to do things like double-tap an empty spot on the page if they really want to zoom in.

The goal of this fix is to only disable the zoom before (and re-enable after) the point where the buttons and the input field are focused, and re-enable the zoom immediately after.  I think [your fix here](https://github.com/infor-design/enterprise/pull/1362/files#diff-0cb1774867e736cb897a054fbeaa6bcfR314) to force the input field focus is probably correct (along with the preventDefault for the click), but you also want to ensure that the field focus isn't also going to zoom.

As an example, if you look at Dropdown's `openList()` function, between the points where we use this same pattern, all the potential field "focusing" on the list happens:
- [disable point](https://github.com/infor-design/enterprise/blob/4.13.x/src/components/dropdown/dropdown.js#L1565-L1567)
- [enable point](https://github.com/infor-design/enterprise/blob/4.13.x/src/components/dropdown/dropdown.js#L1803-L1805) 

If you were to remove the event triggers from `openList()`, you'd see the same problem you're seeing here.  The trick that fixes the problem is all based on timing:
- the quickly-timed disable of zoom
- then the field focus
- then the re-enable of zoom

Feel free to ping me for more info on how this all works.  Admittedly it's not straightforward.

EdwardCoyle:
@tmcconechy I raised #1384 for the field alignment issue for another sprint.

Issue #1239, opened 2018-11-19
Enter key opens dropdown list, when it should only be used to select items within an open list


sanx72:
According to the documentation, the following key behaviours apply to drop-down lists...

![image](https://user-images.githubusercontent.com/44396995/48700603-4f482880-ebe4-11e8-887b-d10ccc27ad19.png)

...so pressing space or down arrow is documented as opening the list dropdown (so is alt+down arrow but doesn't work in FireFox/Google), but pressing the Enter key also opens the list, and this causes a big problem for us as out applications use enter as input/advance, so if the user has focus on a drop-down they cannot advance to the next screen without moving to a non-list field.

Previously, in 4.4 version, I was able to override this behaviour by putting an addtional key-down event on the "pseudoElem", but now it seems that there is another key event firing before this so the list is already open.

Ideally, the drop-down should conform to the documented behaviour, but if that is not possible, I need to be able to intercept the very first key-down event and block it before the "list open" event is fired (FYI - tried using the onKeyDown option and that only appears to be called when the list is already open) , so any recommended example of doing that would be useful.

tmcconechy:
I think your right, enter should only work when the list is open or you cant submit the form. We should be following these guidelines https://www.w3.org/TR/wai-aria-practices-1.1/#combobox and as noted in the docs. i also tested the onKeyDown and it is in fact only working when open.

Changes
- Make onKeyDown work when closed
- fix alt+down arrow
- make enter not block and not open the list
- test other keys.

tmcconechy:
@ericangeles make sure to test https://github.com/infor-design/enterprise/issues/1262 as well but i think thats a dup, just give it a test and we should be good on that too.

brianjuan:
http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html
Moving to failed QA. Enter key doesn't work when the list is open.

tmcconechy:
@ericangeles - For that case you could just update the example so the enter key works if the list is open. Fx.. 

```
if ('#dropdown-list').is(':visible')) {
   return;
}
```

janahintal:
QA Passed. Moving to Done.
Tested in https://4140-beta0-enterprise.demo.design.infor.com/components/dropdown/test-allow-custom-keystroke.html

CindyMercadoReyes:
QA BDD Test Script:
[story_github#1239.docx](https://github.com/infor-design/enterprise/files/2691043/story_github.1239.docx)

PR #1363, opened 2018-12-10
1239 - Fixes enter key stroke to be disabled

ericangeles:

Adding enter key stroke to disabled via overriding it on `onKeyDown` function. I also added a sample template for this. 

Closes https://github.com/infor-design/enterprise/issues/1239

- Pull this branch
- Run http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html
- Select the dropdown via tab key. 
- Press enter key
- It should not open the list


src/components/dropdown/dropdown.js
```diff
@@ -1306,6 +1306,15 @@ Dropdown.prototype = {
       return false;
     }
 
+    if (this.settings.onKeyDown) {
```
tmcconechy: Since this repeats on  https://github.com/infor-design/enterprise/blob/57346de44658fa40694a699469b72c9ebf0b9a3a/src/components/dropdown/dropdown.js#L1364 perhaps just merge these two into a new @private function (handleOnKeyDown maybe)

app/views/components/dropdown/test-disabled-enter-key.html
```diff
@@ -0,0 +1,38 @@
+<div class="row">
```
tmcconechy: We are trying to reduce the number of sample pages. This seems to be covered by http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html so lets just remove this file and update the test description to point to http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html

But its working there!

src/components/dropdown/dropdown.js
ericangeles: Hi @tmcconechy, made some several tests and I believe the handleKeyDown will only use once. So I did not create a function to reuse the script.

Issue #931, opened 2018-10-04
Hierarchy Context Menu With Details Component: Profile text is cut off


CindyMercadoReyes:
**Describe the bug**
Windows 10 IE 11: Hierarchy Context Menu With Details Component: Profile text is cut off

**To Reproduce**
1. Go to URL: http://4110-beta1-enterprise.demo.design.infor.com/components/hierarchy/example-context-menu-with-details.html
2. Click on the three dot eclipse button for Kaylee Edwards on WIn 10 IE 11
3. Her profile text is cut off.
4. Please see screen shot attached. 

**Expected behavior**
The text for her profile should not be cut off

**Screenshots**
![hierarchy component](https://user-images.githubusercontent.com/39101346/46495471-7fc63580-c7e3-11e8-9c2b-44ad76b2666f.png)

**Platform**
Windows 10 IE 11
**Additional context**
None

tmcconechy:
@krishollenbeck

tmcconechy:
Added to this sprint so it can be QA'd. This was pulled out of up for grabs while we change sprints.

CindyMercadoReyes:
This issue is still occurring on Mac FF 63.

tmcconechy:
@CindyMercadoReyes The issue was for windows. What do you see wrong in Mac FF?
The only thing i can see if is if you make the page shorter the menu wont fit, but you can scroll it. Thats all we can do.

ericangeles:
Yeah I see it also in Mac FF. It seems when the scrollbar appears, it tends to overlap the detail fields same as in IE11. We can handle this one @tmcconechy. This would be a quick fix, then I'll create a new PR.

janahintal:
QA Passed.  Moving to Done. 
Tested in: http://4140-rc0-enterprise.demo.design.infor.com/components/hierarchy/example-context-menu-with-details.html for Windows 10 IE 11

PR #1392, opened 2018-12-13
931 - Added fix for hierarchy profile text cut off

ericangeles:

Windows 10 IE 11: Hierarchy Context Menu With Details Component: Profile text is cut off. Fixing it by adding some padding specific only in IE and IE11.

Closes https://github.com/infor-design/enterprise/issues/931

- Pull this branch
- Run http://localhost:4000/components/hierarchy/example-context-menu-with-details.html
- Click on the three dot eclipse button for Kaylee Edwards on Windows 10 IE 11 (Test it via Browserstack)
- Her profile should not cut off

### DIFF
--- a/src/components/hierarchy/_hierarchy.scss
+++ b/src/components/hierarchy/_hierarchy.scss
@@ -847,3 +847,13 @@ html[dir='rtl'] {
     }
   }
 }
+
+// For IE layout issues
+.ie,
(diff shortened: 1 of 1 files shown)


tmcconechy:
Added to this sprint so it can be QA'd. This was pulled out of up for grabs while we change sprints.

PR #1397, opened 2018-12-13
406 - Datagrid: use Mask component with Lookup Editors

### LINKED ISSUE #406: Datagrid: `Soho.Editors.Lookup` editor doesn't support mask
Original Issue: [SOHO-7928](https://jira.infor.com/browse/SOHO-7928)
----

Datagrid columns using the `Soho.Editors.Lookup` do not support Mask.  For example, we would like to restrict the input to 10 numeric characters.  This is currently not supported.

So the following does not work (even when adding `maskMode` and `maxLength`)

```
SohoDataGridColumn)
    columns.push({
      id: 'lookup-click',
      name: 'Lookup Click',
      field: 'productId',
      editor: Soho.Editors.Lookup,
      editorOptions: {editable: true, click: LMLookupClickFunction},
      maskMode: 'number',
      mask: '##########',
      maxLength: 10
    });
```

EdwardCoyle:
This PR enables the ability to use Masks on Lookup Editor fields in a Datagrid, the same way you'd use a mask in a normal input field.

Closes #406

- Navigate to http://localhost:4000/components/datagrid/test-editable-lookup-mask
- Select any cell in the "Product ID" column
- Attempt to key letters as the value.  See that no letters are allowed to be used.  Using numbers will work.

There's also a new e2e test for this page/case that should pass

### DIFF
--- a/test/protractor.local.debug.conf.js
+++ b/test/protractor.local.debug.conf.js
@@ -2,7 +2,6 @@
 const basePath = __dirname;
 const { SpecReporter } = require('jasmine-spec-reporter');
 const protractorImageComparison = require('protractor-image-comparison');
-const customSpecs = require('./helpers/detect-custom-spec-list')('e2e');
 const specs = require('./helpers/detect-custom-spec-list')('e2e', process.env.PROTRACTOR_SPECS);
 
 exports.config = {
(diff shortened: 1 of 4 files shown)


test/protractor.local.debug.conf.js
```diff
@@ -15,8 +14,7 @@ exports.config = {
   SELENIUM_PROMISE_MANAGER: false,
   capabilities: {
     browserName: 'chrome' || 'firefox',
-    shardTestFiles: true,
-    maxInstances: 2,
+    shardTestFiles: false
```
tmcconechy: Made this change on my PR too 👍 One of us may need to fix a conflict...

brianjuan:
http://localhost:4000/components/datagrid/test-editable-lookup-mask
Passed QA testing on localhost and master, will conduct further testing when 4.15.0 build is ready.

PR #1416, opened 2018-12-17
1248 - disable client sort and filter and retain settings

### LINKED ISSUE #1248: disableClientSort/disableClientFilter visual for the filter bar and sort indicator are lost after each loading of data.
**Describe the bug**
When loading data with datagrid.loadData any filter field and sort visual state is lost when disableClientFilter and disableClientSort are used, respectively.

Suggest not resetting the visual state when these settings are being used. The justification for that would be that the implementing code should manage all those when these settings are used.

**To Reproduce**
1. Change this in components/datagrid/example-filter.html 
```
* Make sure one column has a sortable: true
      columns.push({ id: 'productName', name: 'Product Name', field: 'productName', sortable: true, formatter: Formatters.Hyperlink, filterType: 'text', click: function(e, args){ console.log(e, args);}});

* use disableClientfilter and sort settings and loadData when filtered adn sorted events are fired. 

      // Init and get the api for the grid
      var datagrid = $('#datagrid').datagrid({
        columns: columns,
        dataset: data,
        filterable: true,
        selectable: 'single',
        disableClientFilter: true,
        disableClientSort: true,
        editable: true,
        emptyMessage: {title: Locale.translate('NoData'), info: Locale.translate('NoDataFilter'), icon: 'icon-empty-no-data'},
        toolbar: {title: 'Filterable Datagrid', filterRow: true, results: true, dateFilter: false ,keywordFilter: false, actions: true, views: false, rowHeight: true, collapsibleFilter: false}
      }).on('filtered', function (e, args) {
        console.log('filtered', args);
        datagrid.loadData(data);
      }).on('sorted', function (e, args) {
        console.log('sorted', args);
        datagrid.loadData(data);
      }).data('datagrid');
```
2. Open the sample 
3. Click on sort column and note the sort indicator doesn't seem to get set. 
4. Enter in a filter value and type enter. notice the value is cleared

**Expected behavior**
When using the disableClientFilter and disableClientSort calling loadData should leave the visual state of these as is and not clear them.

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version 70.0.3538.102 (Official Build) (64-bit)

**Additional context**

### LINKED ISSUE #1414: About Component display IDS Version in device specs area with apps version
**Describe the bug**
If include the device specs, the version being passed in is displaying for the IDS Version as well in the About component. Should display the IDS Version instead.

**To Reproduce**

1. edit app/views/components/about/example-index.html to 
```
    //Show About Screen
    $('body').about({
      appName: 'IDS Enterprise',
      productName: 'Controls',
      version: 'Example Application Version No. XX',
      content: '<p>Fashionable components for fashionable applications.</p>'
    });
```
2. open http://localhost:4000/components/about/example-index.html
3. should show bothe the app version on the top and the current ids version on the bottom

If include the device specs, the version being passed in is displaying for the IDS Version as well in the About component.  Should display the IDS Version instead.
**Expected behavior**
Should be able to show both a custom app version AND the Ids version.

**Version**
- 4.14-rc0

**Platform**
- Any

**Additional context**
Mentioned on MS Teams.

Overlooked on https://github.com/infor-design/enterprise/pull/913

tmcconechy:

Fixes two unrelated issues as mentioned in the test notes. And additionally
- removed extra spaces in the markup class names and undefined classes being added
- Add a couple awaits i spotted as missing

Fixes #1248
Fixes #1414

For 1248
http://localhost:4000/components/datagrid/test-disable-client-filter-and-sort.html
- sort on product id
- should stay looking like its sorted
- type 22 on product id filter
- should stay looking like its sorted and the values in the filter should stay

NOTE: Because of what disableClientFilter and disableClientSort do. The data wont actually appear sorted or filtered at all. Idea is that the server would return this but in this example its just refreshing the same data. So its correct that sort and filter do not do anything to the data except for the visual indicators.

For 1414
http://localhost:4000/components/about/example-index.html
- open it
- version at the top and the IDS version should be able to be different


brianjuan:
http://localhost:4000/components/datagrid/test-disable-client-filter-and-sort.html
Passed QA testing on localhost and master, will do a retest once 4.15.0 build is provided.

PR #1418, opened 2018-12-17
1357 - Fixed Demoapp's user-preferred layout settings

### LINKED ISSUE #1357: Demoapp: Detect user-defined layout before usage
This is a follow-up issue to #1327.  It's currently possible to use query params to set the desired alternate layout for a test/example page, but the existence of the layout is not validated before the demoapp attempts to load it.

**Describe the bug**
For example, running http://localhost:4000/components/dropdown/example-index?layout=notalayout currently errors out with a 404.  The error happens on the attempt to load the layout from the filesystem.

**Expected behavior**
We should find a way to detect if the layout exists or not, defaulting to the standard one that it would otherwise load if no match exists.  In this case, the query parameter will simply be ignored if it doesn't match a valid layout. We don't have anything that scans the filesystem for a matching layout at the moment so we'll need to write a small detector for the file.

**Version**
- Starting in v4.13.x
- n/a in prior versions

EdwardCoyle:
This PR makes some adjustments to the way the demoapp works with layout files, and how users can set a URL query parameter to define preferred layout, such as `nofrills` or `embedded`:
- The existence of a layout file is now pre-checked before the system attempts to use it (fixes a bug where adding a query param of `?layout=notalayout` would cause the demoapp to hang)
- In the demoapp code, setting a layout by hitting a specific route or via an option is now more explicit (a method `setLayout` handles the checking of the file, and setting the file path back to default if needed).
- There were some issues where `app/views/components/[componentName]/layout.html` files existing inside of a test/example page folder were being ignored in favor of the base layout.  Now these files are preferred over the base layout if they exist.

Closes #1357 

Pull this branch and run the demoapp with `npx grunt demo && node app/server.js --verbose`.  After, check the following URLs for the requirements below:

- http://localhost:4000/components/header/test-header-gauntlet.html
  - Ensure that there is only one blue header
- http://localhost:4000/components/header/test-header-gauntlet.html?layout=embedded
  - Ensure that the embedded layout links properly load, and that the blue header is displayed.
- http://localhost:4000/?layout=nofrills
  - Should show a scrollable pane with the no-frills layout
- http://localhost:4000/?layout=notalayout
  - In the console, there should be a warning about not being able to find the specified layout file from the query params.
  - When the page loads, it should be using the default layout.

You can also just browse the various demoapp directories and example pages, testing their normal layouts and adding `layout=nofrills` or `layout=embeded` to make sure they work.


EdwardCoyle:
@davidcarlsonberg you'll only see the warning in your console if you run the demoapp in verbose mode:

```sh
node app/server.js --verbose
```
Should look like this when it happens:

screen shot 2018-12-19 at 11 46 00 am

davidcarlsonberg:
Weird, I am in `verbose` mode and it still isn't in console...

Issue #1314, opened 2018-11-29
double click on open accordion will close it but set the icon to open


darinw:
**Describe the bug**
When you double click an open accordion-header the it will close on the first click, but will not reopen on the second. the arrow will change to open on the second click. On the third click the arrow will flip back with the content remaining closed. with the fourth click it will be back to normal

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/accordion
2. Click on accordion-header
3. Once open double click accordion-header

**Expected behavior**
either open back up or not allow the second click from going through

**Version**
current release 4.12.1

janahintal:
QA Passed. Moving to Done.
Tested in: http://localhost:4000/components/accordion/example-index.html for all browser, OS and devices.

PR #1428, opened 2018-12-20
1314 - Fixed double click on open accordion

ericangeles:

When you double click an open accordion-header the it will close on the first click, but will not reopen on the second. the arrow will change to open on the second click. On the third click the arrow will flip back with the content remaining closed.

Closes https://github.com/infor-design/enterprise/issues/1314

- Checkout this branch
- Run http://localhost:4000/components/accordion/example-index.html
- Click on accordion header
- Once open double click accordion header
- Should expect to not flicker and open even doing double click


src/components/accordion/accordion.js
```diff
@@ -775,12 +777,16 @@ Accordion.prototype = {
 
       if (pane.hasClass('no-transition')) {
         for (let i = 0; i < pane.length; i++) {
-          pane[i].style.display = 'block';
-          pane[i].style.height = 'auto';
+          setTimeout(() => {
```
EdwardCoyle: This has an adverse effect of adding a delay to the expansion of any headers that are clicked a single time.  It also seems like the expand/collapse calls still stack up on each other:

dec 21 2018 11_06 am

ericangeles: @EdwardCoyle thanks for the suggestions! Anyway, I don't see that error occurring on my end though. I tried to stopPropagation when doing a double click in the event but it seems the error is still there.

tmcconechy:
Oh also can merge master now to fix tests.

ericangeles:
Thanks for all the inputs @EdwardCoyle & @tmcconechy. Will do the fix now.

PR #1433, opened 2018-12-20
1432- Allowed pager to have a custom pageSizeButton text

### LINKED ISSUE #1432: Add ability to override pageSizeButtonText for a stand-alone-pager
Want to be able to show the page size chooser at narrow widths. But once it gets too narrow the standard pageSizeButton text is too wide.

**Additional context**
image

pwpatton:
Add ability to change pageSizeButton text on a stand alone pager bar.

Closes #1432

1. edit `app/views/components/pager/example-standalone.html`
2. Add pageSizeLabelText: 'Recs {0}', // must be translated text.
3. Run the example and see text change

### DIFF
--- a/app/views/components/pager/example-standalone.html
+++ b/app/views/components/pager/example-standalone.html
@@ -49,6 +49,7 @@
       type: 'standalone',
       pagesize: 10,
       showPageSizeSelector: false,
+      pageSizeLabelText: 'Recs {0}', // must be translated text.
       showFirstButton: true,
       enableFirstButton: true,
       showPreviousButton: true,
@@ -57,6 +58,7 @@
       enableNextButton: true,
       showLastButton: true,
       enableLastButton: true,
+
       onFirstPage: function (elem, args) {
         console.log(elem, args);
       },
--- a/src/components/pager/pager.js
+++ b/src/components/pager/pager.js
@@ -12,6 +12,8 @@ import '../tooltip/tooltip.jquery';
 // The name of this component.
 const COMPONENT_NAME = 'pager';
 
+const RECORD_PER_PAGE_TEXT = 'RecordsPerPage';
+
 /**
 * The Pager Component supports paging on lists.
 * @class Pager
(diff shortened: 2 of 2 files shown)


src/components/pager/pager.js
```diff
@@ -731,8 +736,10 @@ Pager.prototype = {
         });
 
         // Update the number of records per page
+        const buttonText2 = self.settings.pageSizeLabelText ||
```
EdwardCoyle: @pwpatton as a code style in EP, we try to avoid adding arbitrary numbers to variable names that look similar to others.  I think this set of lines is basically doing the same replacement as the [ones here](https://github.com/infor-design/enterprise/pull/1433/files#diff-5ec3aeece02e6ca8b60f7f7080ef5362R827).  Could we pull these out into a new method, maybe `updateRecordsPerPage()`?  That would avoid the variable name conflict and make it a bit cleaner

app/views/components/pager/example-standalone.html
```diff
@@ -49,6 +49,7 @@
       type: 'standalone',
       pagesize: 10,
       showPageSizeSelector: false,
+      pageSizeLabelText: 'Recs {0}', // must be translated text.
```
tmcconechy: We can translate this in https://github.com/infor-design/enterprise/blob/master/src/components/locale/cultures/en-US.js and i usually add one to https://github.com/infor-design/enterprise/blob/master/src/components/locale/cultures/es-ES.js for testing

pwpatton: @tmcconechy do you think it would be better to, in EP, come up with standard shorter text when the screen is narrowed instead of optional pagesize label text?

tmcconechy: Yes i kind of thought thats what you were going for for smaller size. I think thats a good idea.
I dont think "Recs" will translate well. But maybe just "10 Records" or even just "10"

pwpatton: Ok, so cancel this PR and create a new issues to have a pageSizeChooser show at a smaller (phone) width.

pwpatton:
Intend to make pageSizeChooser responsive to display at narrower screen widths

Issue #1435, opened 2018-12-20
SASS - Migrate all original color palette variables to token color palette variables


clepore:
**Is your feature request related to a problem? Please describe.**
We would like to switch from the existing sass color palette variables to solely using the token's.

**Describe the solution you'd like**
Replace existing sass color palette variables with the token color palette variables

**Describe alternatives you've considered**
n/a

**Additional context**
This will also help to isolate colors that aren't in the color the color palette so we can remove and deprecate them in the future.

brianjuan:
Passed QA testing on all browsers using localhost, will do testing again once 4.15.0 build is ready.

PR #1440, opened 2018-12-21
1435 - Migrate color palette variables to be those of the tokens

clepore:
We want to deprecate the original color palette scss variables in exchange for the ids-identity token ones.

#1435

1. `npm i`
1. `npm run start`
    - Verify the app colors looks the same as on production for each theme
        - [ ] Light
        - [ ] Dark
        - [ ] Contrast

Note that the high contrast yellow now being revealed from tokens [in this PR](http://localhost:4000/components/badges/example-index.html?theme=high-contrast) looks very dark as opposed to [4.13.0 (current)](https://master-enterprise.demo.design.infor.com/components/badges/example-index.html)

screen shot 2018-12-21 at 12 49 38 pm

@kayiuho Can you comment as to whether or not this is correct as you recently fixed the high contrast colors in this PR: https://github.com/infor-design/design-system/pull/255.


clepore:
Per @kayiuho , the "caution" color is correct, as the previous yellow shown is not AAA compliant. I should update the high contrast styles to use a white font for `caution`.

Also I need to update the package once https://github.com/infor-design/design-system/pull/259 is merged.

PR #1455, opened 2019-01-02
Adding support for lookup on a datagrid filter.

### LINKED ISSUE #653: Lost support for Lookup filter type in datagrid
**Describe the bug**
In the 3.x controls, we could specify a lookup filter type. This is missing in 4.0.

**Expected behavior**
Should be able to specify a lookup filter type for datagrid.

**Screenshots**
In 3.x, this is what we passed in as the column definition:
image

and the lookup filter looked like this:
image

**Platform**
 - All

**Additional context**

Sovia:
Adding the ability to have a lookup in a datagrid filter.

"Closes #653"

After pulling in the change, navigate to /components/datagrid/example-filter

The Product Id column's filter should now be a lookup:
image

And when you press the trigger button in the Product Id's filter, the lookup should popup:
image

You can select a row, and the value is returned:
image

And the value can be successfully filtered on (and cleared):
image


src/components/datagrid/_datagrid.scss
```diff
@@ -2286,6 +2286,43 @@ $datagrid-short-row-height: 23px;
...
       }
 
+      .lookup-wrapper {
+        background-color: transparent;
+        display: inline;
+        left: 2px;
+        vertical-align: top;
+        width: calc(100% - 44px);
+
+        > input.lookup {
```
davidcarlsonberg: Is the `>` combinator necessary in any of these selector chains?

src/components/datagrid/_datagrid.scss
```diff
@@ -2286,6 +2286,43 @@ $datagrid-short-row-height: 23px;
...
+            fill: $datagrid-filter-icon-color;
+            height: 30px;
+            right: -6px;
+            top: -3px;
+            width: 16px;
+          }
+
+          &:not(:disabled) > .icon:hover,
+          &:hover:not(:disabled) > .icon {
+            fill: $datagrid-filter-hover-color !important;
```
davidcarlsonberg: Is `!important` necessary here or for any of these values?

src/components/datagrid/_datagrid.scss
```diff
@@ -2286,6 +2286,43 @@ $datagrid-short-row-height: 23px;
...
+          > .icon {
+            fill: $datagrid-filter-icon-color;
+            height: 30px;
+            right: -6px;
+            top: -3px;
+            width: 16px;
+          }
+
+          &:not(:disabled) > .icon:hover,
+          &:hover:not(:disabled) > .icon {
```
davidcarlsonberg: Hover state should activate when user hovers over the sibling `input`. Similar to `datepicker` in the same example page.

src/components/datagrid/_datagrid.scss
Sovia: Nope. Took them out

src/components/datagrid/_datagrid.scss
Sovia: It is not. Took it out

src/components/datagrid/_datagrid.scss
Sovia: Fixed it so the icon highlights like date picker icon

PR #1480, opened 2019-01-11
1422 - Added support to allowChildExpandOnMatch with datagrid

### LINKED ISSUE #1422: Datagrid Tree - Make Filtered child rows a setting
**Describe the bug**
As a continuation on #1281 

**To Reproduce**
Behavior number 2 still happens when you apply filter that match the parent node. Expand/collapse button is displayed but doesn't do anything.
Steps:

- Go to https://4140-beta0-enterprise.demo.design.infor.com/components/datagrid/test-tree-filter.html
- Apply Contains filter on Task column with the word "HMM"
- Try to Expand/Collapse the first row, nothing happens.

**Expected behavior**
Make this behavior a setting.

Add this as api settings allowChildExpandOnMatch: boolean

allowChildExpandOnMatch === true

- if only parent got match then add children nodes too
- if one or more child node got match then add parent node and all the children nodes
allowChildExpandOnMatch === false

- if only parent got match then (disable|hide|anything-else) +/- button and no add children nodes
- if one or more child node got match then add parent node and only matching children nodes (as it doing now)

**Additional context**
#1281

deep7102:
Make filtered child rows a setting `allowChildExpandOnMatch` to display the children nodes with tree datagrid.

Closes #1422

http://localhost:4000/components/datagrid/test-tree-filter-child-on-match.html
- Open above link
- Use switch "allowChildExpandOnMatch" to toggle the settings
- Apply "Contains" filter on "Task" column with the word "HMM"
- if "allowChildExpandOnMatch" is "true"
- there should be 21 rows, 18-visible, 3-hidden(2,3,4 -in column "id")
- and in first row "+" button should be enabled
- if "allowChildExpandOnMatch" is "false"
- there should be 6 rows, all 6-should be visible
- and in first row "+" button should be disabled

### DIFF
--- a/app/views/components/datagrid/test-tree-filter-child-on-match.html
+++ b/app/views/components/datagrid/test-tree-filter-child-on-match.html
@@ -1,0 +1,88 @@
+
+<div class="row">
+  <div class="twelve columns">
+    <h3>Api setting: <strong>allowChildExpandOnMatch</strong></h3>
+    <p>
+      if true
+      <ul>
+        <li>&mdash; and if only parent got match then add all children nodes too</li>
+        <li>&mdash; or if one or more child node got match then add parent node and all the children nodes</li>
+      </ul>
+    </p>
+    <p>
+      if false
+      <ul>
+        <li>&mdash; and if only parent got match then make expand/collapse button to be collapsed, disabled and do not add any children nodes</li>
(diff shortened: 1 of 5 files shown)


deep7102:
Thanks! @davidcarlsonberg and sorry for my notes above not fully clear
- hidden means, not expanded (5 is filtered out, so it should not be in the list)

PR #1492, opened 2019-01-16
257 - Fix for Color Picker not initially disabled using editable false

tjamesallen15:

Added editable false option and example pages.

Fix for 257 of enterprise-ng
https://github.com/infor-design/enterprise-ng/issues/257


tjamesallen15:
@tmcconechy I just aligned everything what is needed in the NG. Since NG has a editable option in the setting, I also put the disable option there too. 

I used the existing disable option in the NG and I just added the option lines inside of it.
https://github.com/infor-design/enterprise-ng/pull/323
You can see it 

soho-colorpicker.component.ts under line 84 aligned in 107.

tmcconechy:
Ok, Sounds good.

PR #1493, opened 2019-01-16
1267 - Fix for initialized event is not called if a non-standard locale

### LINKED ISSUE #1267: The "initialized" event is not called if a non-standard locale is used
The "initialized" event is not called if a non-standard locale is used.

Our application allows the customer to define and configure their own locales. In the pre-SoHo Xi versions of the controls (3.x), if the customer defined a locale that the controls did not recognise/support, it defaulted to en-US so at least the program continued to work.

In SoHo Xi, if an invalid locale is passed to the initialize() method, it seems to prevent the "initialized" event from firing, which, in our application, is where all the page load and configuration functionality is located, so, to the end user, the application appears to freeze.

Here is a simple example...

```
		var initialLocale = 'xx-XX';
		//var initialLocale = 'en-UK';

		$('body').initialize({
			locale : initialLocale
		}).on('initialized', function() {
			alert("Initialized fired");
		});

```

Add this script to any standard SoHo Xi page template and the alert box will not be fired. Comment the "xx-XX" locale variable and uncomment the "en-UK" one and it works as expected. 

Looking in the sohoxi.js, it looks like it should be dropping down to the "en-US" locale as a default, but in the init function, if the "Locale.set(locale)" fails, it will not call the self.initAll() method, which then calls the "initialized" event.

tjamesallen15:

Fix for initialized event is not called if a non-standard locale.

Closes #1267 


tmcconechy:
As mentioned on the call - looks good. Lets just:

- make sure simpler locales like 'de' vs 'de-DE' still work
- make sure simpler locales like 'de' vs 'de-DE' fire events
- make sure the test checks both locales like 'xx' vs 'xx-XX' fire events as well as reverting to 'en-US'

test/components/locale/locale.func-spec.js
```diff
@@ -71,6 +71,20 @@ describe('Locale API', () => {
...
+    Locale.set('xx-XX');
+
+    expect(Locale.currentLocale.name).toEqual('en-US');
+
+    Locale.set('xx');
+
+    expect(Locale.currentLocale.name).toEqual('en-US');
   });
 
   it('Should be possible to after set locale', () => {
```
tmcconechy: I feel like this tests is covered in the previous test. So maybe this can be removed. Or at least correct the description of the test.

PR #1511, opened 2019-01-22
1284 - Fix for QA Failed on Circle Pager Not Rendering Properly [v4.15.x]

### LINKED ISSUE #1284: Circle Pager Tabs Component: Similar Products tab is not displayed correctly on responsive view
**Describe the bug**
Circle Pager Tabs Component: Similar Products tab is not displayed correctly on responsive view

**To Reproduce**
1. Go to url: http://4130-beta0-enterprise.demo.design.infor.com/components/circlepager/example-tabs.html
2. Make the page responsive
3. Click the down arrow key and select "Similar Products
4. The circle images are not being displayed properly. Please see screen shot attached. 

**Expected behavior**
The images should not overlap each other. 

**Version**

**Screenshots**
circle 1

**Platform**
 - OS Version: [iOS 8]
 - Browser Name [Chrome]

**Additional context**
None

tjamesallen15:

Tabs not in the middle.

Closes #1284

Check it on http://localhost:4000/components/circlepager/example-tabs.html
And follow the QA steps on https://github.com/infor-design/enterprise/issues/1284#issuecomment-455044321


src/components/circlepager/circlepager.js
```diff
@@ -544,11 +545,16 @@ CirclePager.prototype = {
       self.responsiveSlidesToShow();
     });
 
-    if (this.element.parent().attr('id')) {
-      $(`a[href="#${this.element.parent().attr('id')}"]`).on('focus.tabparent', () => {
-        self.responsiveSlidesToShow();
-      });
-    }
+    setInterval(() => {
```
deep7102: Not much fan of method `setInterval()` and did not see here anywhere we stop this soon it trigger `clearInterval()`
If the issue causing only in tab changing, is there way we could listen event before changing tabs or any other way to tackle this issue.

tmcconechy: Good suggestion i think you could look on the page for tabs and attach to the tabs afteractivated event. We are trying to remove as many setTimeouts as we can

src/components/circlepager/circlepager.js
```diff
@@ -544,11 +545,9 @@ CirclePager.prototype = {
       self.responsiveSlidesToShow();
     });
 
-    if (this.element.parent().attr('id')) {
-      $(`a[href="#${this.element.parent().attr('id')}"]`).on('focus.tabparent', () => {
-        self.responsiveSlidesToShow();
-      });
-    }
+    $('#tabs-normal').on('activated', () => {
```
tmcconechy: You probably need to make this less specific for example `$('.tab')` because that ID is specific to that example. Try that

deep7102: @tjamesallen15 few things here:
1) the target selector `#tabs-normal` is not good here, this will not work with any other id. So instead of strict the user to `#tabs-normal` id, we should change our selector for more general as class selector `.tab-container`
2) When bind any event please use the namespace, so instead of `activated` use  `activated.circlepager`
3) Then turn off the bonded event `activated.circlepager` in method `unbind()`

Issue #1462, opened 2019-01-04
Tabs: The "add tab" api treats the name as html


jake-harrington:
**Describe the bug**
HTML can be passed as the 'name'  to the add tab api and it will get treated as actual html. The rename api seems to be immune to this in our testing.

**To Reproduce**
1. [example-add-tab-button.zip](https://github.com/infor-design/enterprise/files/2728026/example-add-tab-button.zip)
2. Fill out the fields
3. click the second radio button to use the callback
4. See that an image is used for the tab. There's some JS that would otherwise run if the security settings weren't as strict on the test server (was just trying other possible vulnerabilities), but a script element also would get appended as-is and have the same vulnerability.

**Expected behavior**
Passing a name with html as a tab name shouldn't be allowed to be treated as html.

**Version**
4.11.x

**Screenshots**
![image](https://user-images.githubusercontent.com/39064853/50703507-f872a300-1019-11e9-975e-fb4fb474a7b4.png)

awbuboltz:
Hey @tmcconechy , can we get some type of high priority on this, it's a security issue for us and we need to get it out as soon as possible.

tmcconechy:
Ok, Whoever picks this up try to use `xssUtils.stripTags('1 <br/> 1', '*')` around the message when its added to the DOM.

brianjuan:
Passed QA testing, the fix passed on functional test. Moving to Done.

jake-harrington:
@tmcconechy, One of our devs found that you can still exploit this issue with a name like:
<<a/>script<a>>alert("testing");</<a/>script<a>>

Using stripTags doesn't seem like a very complete solution. Why not just set the text of the element directly so this can't be exploited?

tmcconechy:
It doesnt have a test case for that. But handles basic cases. What do you mean by setting the test of the element directly?

jake-harrington:
Instead of:
const anchorMarkup = $(`<a href="#${tabId}" role="tab" aria-expanded="false" aria-selected="false" tabindex="-1">${xssUtils.stripTags(options.name)}</a>`);

const anchorMarkup = $(`<a href="#${tabId}" role="tab" aria-expanded="false" aria-selected="false" tabindex="-1"></a>`).text(options.name);

tmcconechy:
Ok, I thought thats what you meant 👍 Use the DOM.
That would work but we use stripTags all over so maybe we need to expand it (as well?) to handle that case.

tmcconechy:
[...]

PR #1519, opened 2019-01-22
1462 - Added fix for addTab API treats the name as html

ericangeles:

Issue:  It allows HTML to passed in the `name` tab api and it will get treated as an actual html.

Solution: Adding `xssUtils.stripTags()` into the tab header will remove any markup (xss injection).

Closes https://github.com/infor-design/enterprise/issues/1462

- Checkout this branch
- run the tests added to tabs and confirm they work


tmcconechy:
I'm a little confused about the example, what im seeing is no matter which method i use and i add nothing is being added on the Tab.

screen shot 2019-01-22 at 11 04 18 am

I think this is one we can just do with a unit test only? Also wondering if the add method is now not adding text anything because of what im seeing on the test.

I did merge master to fix your tests

tmcconechy:
I typed in the name field but still didnt show for me?

ericangeles:
Manually change the value of name in the template page so it would show up. It's kinda inconvenient. 😞

PR #1533, opened 2019-01-23
1532 - Ensure that the passed in editorOptions are used when the datepicker gets recreated with a change in filter type

### LINKED ISSUE #1532: Datagrid filter's datepicker doesn't honor editorOptions
**Describe the bug**
The datepicker filter when set to 'in range' filter type doesn't obey passed in editorOptions. For example, passing in 'showMonthYearPicker: true' has the month and year show up as drop downs in the date picker. When you change to 'in range' filter type, the month and year just show up as plain text.

**To Reproduce**
1. Pull in the attached [example-filter.zip](https://github.com/infor-design/enterprise/files/2788817/example-filter.zip) and put it in your datagrid/example-filter directory.
2. Navgiate to /components/datagrid/example-filter
3. Open the datepicker for Order Date, without changing the filter type.
4. The month and year should be dropdowns.
5. Close the datepicker and change it's filter type to 'in range'. Then reopen the datepicker
4. See that the month and years aren't dropdowns

**Expected behavior**
Month and Year should still be dropdowns, so the use can easily set a large date range.

**Version**

- ids-enterprise: v4.15.0-dev

**Screenshots**
On load:
image

After change to 'in range':
image

**Platform**
 - All

Sovia:
Passing in the editorOptions for a datepicker when the filter type is changed.

"Closes #1532 "

To test:
1. Pull in the change
2. Navgiate to http://localhost:4000/components/datagrid/example-filter.html
3. Open the datepicker for Order Date, without changing the filter type.
4. The month and year should be dropdowns.
5. Close the datepicker and change it's filter type to 'in range'. Then reopen the datepicker
6 . The month and year should be dropdowns.

### DIFF
--- a/app/views/components/datagrid/example-filter.html
+++ b/app/views/components/datagrid/example-filter.html
@@ -66,7 +66,7 @@
       columns.push({ id: 'productName', name: 'Product Name', field: 'productName',  formatter: Formatters.Hyperlink, filterType: 'text', click: function(e, args){ console.log(e, args);} , mask: '*****************'});
       columns.push({ id: 'activity', name: 'Activity', field: 'activity', formatter: Formatters.Dropdown, filterType: 'multiselect', options: activities, editorOptions: {showSelectAll: true}});
       columns.push({ id: 'quantity', name: 'Qty', field: 'quantity', align: 'right', filterType: 'integer'});
-      columns.push({ id: 'orderDate', name: 'Order Date', field: 'orderDate', formatter: Formatters.Date, dateFormat: 'MM/dd/yyyy', filterType: 'date'});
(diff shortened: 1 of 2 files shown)


tmcconechy:
I'm not clear on testing this. Do i just added useRange: true on one?
Maybe you can just add a second column next to it so both can be seen without modifying things?

Sovia:
@tmcconechy It's changing the filter type in the column. When you load up the page, by default it's set to equals for a date filter:
image

Having showMonthYearPicker: true on this means that you see the dropdowns:
image

To test, just change from equals to in range:
image

which is the same as setting useRange on the date picker. Without my change, those two dropdowns just appear as text.

In the issue write up I attached an example-filter.html page which set the showMonthYearPicker: true for the order date column. I could add a date column with showMonthYearPicker: false if you want it though?

tmcconechy:
Thanks i glanced over that part

Sovia:
I've raised that with https://github.com/infor-design/enterprise/issues/1537. I don't think it's too important for our case at this moment, and would like to get this fix merged in soon.

Issue #971, opened 2018-10-10
Tab content cut off in mobile 


davidcarlsonberg:
**Describe the bug**
The content of the tab is cut off in smaller screens and also doesn't display in dropdown menu of tab options.

**To Reproduce**
- http://localhost:4000/components/tabs-multi/example-index
- Open mobile Dev Tools in Chrome
- Notice Tab content is cut off on smaller device screens such as iPhone 5
- Notice content of first Tab also does not appear in dropdown of Tab options

**Expected behavior**
As a user I would expect to be able to read the content of the first tab either as a tab or as an option in the dropdown of tab options.

**Screenshots**
<img width="418" alt="screenshot" src="https://user-images.githubusercontent.com/7952168/46739989-bda5dc80-cc6f-11e8-894a-fc6bf4542b8e.png">

**Platform**
 - Mobile, specifically iPhone 5 (or Dev Tools Mobile view for iPhone 5 in Chrome for macOS)
 - Most recent iOS or most recent macOS
 - Chrome
 - Version 69.0.3497.100 (Official Build) (64-bit)

**Additional context**
Spawned from PR #619 and corresponding Issue #960

tmcconechy:
Im not sure how we could fix this, its not totally incorrect. So may be one we just leave as an awkward edge case.

davidcarlsonberg:
Happy to take your lead on it. We can consider this issue documentation of the edge case.

tmcconechy:
One thought is maybe on this very low break point can hide the "4 More items" text so its just an arrow and there is a bit more room. Still a long title wont show, but perhaps better.

Further, im not sure a side by side pattern like this is going to be of much benefit on a smaller devices as you are comparing two things side by side.

janahintal:
QA Failed. Issue still exist in iOS Mobile for both Safari and Chrome browser. 
Tested in: https://4150-beta0-enterprise.demo.design.infor.com/components/tabs-multi/example-index

Steps to replicate:
1. in portrait mode, open safari or chrome 
2. goto > https://4150-beta0-enterprise.demo.design.infor.com/components/tabs-multi/example-index
3. notice that there is no ellipsis for the 1st and 3rd column
4. turn the phone to Landscape mode
5. notice that the ellipsis is already visible
6. turn the phone again to Portrait mode
7.  refer to the image below

Using iphone XS: doesn't have ellipsis and tab is cut off
[...]

PR #1536, opened 2019-01-24
971 - Added solution so pages can run on safari <=9 [v4.15.x]

tmcconechy:

We found that CSP is not support for nonce values on Safari <=9 (iphone 6s or lower).
This fixes this by allowing the unsafe-inline for just those cases.

#971 

- pull branch
- restart server
- open any page at all like http://localhost:4000/components/datagrid/example-index.html
- do that on Iphone 6s and then on any newer browser plus chrome desktop

### DIFF
--- a/app/src/js/middleware/csp-handler.js
+++ b/app/src/js/middleware/csp-handler.js
@@ -1,3 +1,5 @@
+const browser = require('browser-detect');
+
 // Adds a stored "nonce" attribute to all script tags to conform with security policy.
 function addNonceToScript(html, nonce) {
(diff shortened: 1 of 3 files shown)


clepore:
FYI We currently support Safari `v11` and `v12` (aka release minus 1). Remember to verify browser version validity before spending time and adding unnecessary code, please.

clepore:
FYI We currently support Safari v11 and v12 (aka release minus 1). Remember to verify browser version validity before spending time and adding unnecessary code, please.

PR #1544, opened 2019-01-25
Changes z-index on dropdown popupmenu

guoliang:
Infor Mongoose has a setting which turns open tabs into floating elements which what I could see sets the z-index into 19000. Because of that the dropdown popup is hidden.

https://jira.infor.com/browse/MUPS-2117

Put dropdown component in a positioned element and give that element a high z-index above 4501

### DIFF
--- a/src/components/dropdown/_dropdown.scss
+++ b/src/components/dropdown/_dropdown.scss
@@ -210,7 +210,7 @@ div.multiselect {
   position: absolute;
   white-space: nowrap;
   width: auto;
-  z-index: 4501;
+  z-index: 24501;
 
   &.text-align-reverse {
     text-align: right;


tmcconechy:
So just going to close this one. Appolgies. As we have the system in place i mentioned that best way is to update your indexes to follow the system we made or make a custom override on your end.

This change would break things for us.

clepore:
I agree with that @tmcconechy has said.

If we enabled a setting for z-index on all of our components, there is a good chance that a consumer setting one incorrectly would compromise library's component interaction in the consumer's system. We would like to avoid that kind of negative scenario.

I would highly suggest customizing your CSS to override our css `z-index` properties (perhaps by way of applying a class dynamically or using plain css on page load). This will guarantee the exact results you want, working exactly as you want to control it. 

Note that our markup/html structure is fairly concrete and we only release breaking changes in major versions, so this would be a fairly safe customization (as those are done few and far between).

Issue #464, opened 2018-07-17
SOHO-1564: Datagrid Frozen Columns


clepore:
The inforDataGrid allows you to specify a "frozenColumn" option, which causes some columns to become fixed so that when the grid is too wide to fit on the page, the scroll bar that appears only applies to the unfrozen columns.  Under the frozen columns, there is a scrollbar, but it is always disabled.
  It would be nice if there was some way to reduce the width of the region containing the frozen columns, and then using the scroll bar under the frozen columns to see them.

Landmark: Needs ability to freeze/lock the at least first column of a datagrid, which may be a possible workaround.

Sales Hub: Needs to freeze/lock the last column (which is an action button) that should always stay in view.

brianjuan:
Noticed some issues that might be related to this enhancement.
**Issue 1: There is alignment issue on Button Column Header**
<img width="1671" alt="screen shot 2019-02-26 at 11 56 42 am" src="https://user-images.githubusercontent.com/39508864/53386994-aae23900-39bf-11e9-9017-033a510bd6da.png">

**Issue 2: Alignment issues occur on the Column Headers and rows after you scroll to the rightmost of the grid**
<img width="1214" alt="screen shot 2019-02-26 at 12 01 37 pm" src="https://user-images.githubusercontent.com/39508864/53387078-217f3680-39c0-11e9-9694-2e0cb99063fc.png">

**Issue 3: Sorting a column makes the page longer**
Steps:
1. Sort the Product Name column several times
2. Noticed that the page is now longer, vertical scrollbar will be displayed on the right of the page
![frozengrid1](https://user-images.githubusercontent.com/39508864/53387361-8ab37980-39c1-11e9-9951-bbd7f3473b9d.gif)

Please advise if we need to create separate tickets for these issues, thanks.

tmcconechy:
@brianjuan I think these are worth looking at. Are these windows? I suggest making one ticket with all three points as they are somewhat related and I’ll investigate further.

brianjuan:
@tmcconechy The first two issues on column alignment only happen on Mac Chrome, Issue 3 happens on all browsers both in Mac and Windows. I will just create a ticket for issue 1 and 2, and a separate ticket for issue 3. Thanks!

brianjuan:
Moving to Done since issue 1 and 2 are intermittent, can't replicate it now. And we already have a separate ticket for issue 3.

tmcconechy:
[...]

PR #1598, opened 2019-02-05
922 - Listview Paging Performance / Pager Refactoring

### LINKED ISSUE #922: List View: Improve paging performance on larger datasets
**Describe the bug**
Extending on #430 we want to improve the performance on paging.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/listview/test-listview-performance.html from the PR on #430 once merged
2. Notice on render that initially all nodes are added to the page, then they are paged
3. Change page size to 100 so many pages are added (numbers are in a weird sequence)

**Expected behavior**
Only the first page is ever added to the DOM and the numbers are in a better sequence.

**Platform**
 - Worst on IE11

**Additional context**
#430 
#250

EdwardCoyle:
This PR fixes performance issues that were previously causing long page reloads when attempting to navigate pages on the Listview component.

The changes involve re-working the Pager component to be less depended-on by components that integrate it, and separate the concepts of "paging" from "rendering" completely, leaving the actual components to manage rendering.  These changes make a dataset the "source of truth" as opposed to DOM elements in the following components:
- Listview
- Blockgrid
- Datagrid (already primarily data-based)

Fixes #922

Pull this branch, run the demoapp.
_To check changes to Listview performance, play around with paging/filtering on these pages in IE11 (and other browsers):_
- http://localhost:4000/components/listview/test-listview-performance.html (note that this one still loads all the records, so it will still be a bit slow)
- http://localhost:4000/components/listview/test-listview-performance-paging.html (This one should be significantly faster)

_To check pager rendering management changes, check that paging/filtering all works as expected on this list (note that a backend is not necessarily tied to all grid columns, but generally filtering on `Product ID`, `Product Name`, or `Quantity` works:_
- http://localhost:4000/components/listview/example-paging-clientside.html
- http://localhost:4000/components/listview/test-paging-indeterminate.html
- http://localhost:4000/components/listview/test-paging-pagesize.html
- http://localhost:4000/components/listview/test-reload-paging.html
- http://localhost:4000/components/datagrid/example-paging
- http://localhost:4000/components/datagrid/example-paging-indeterminate
- http://localhost:4000/components/datagrid/test-paging-page-size-selector.html
- http://localhost:4000/components/datagrid/test-paging-multiselect-select-across-page.html

_Other things to test:_
- Standalone pager example should simply work when all checkmarks are toggled on/off: (http://localhost:4000/components/pager/example-standalone.html)
- Filtering (Listview/Datagrid)
- Sorting (Datagrid)
- Make sure all existing e2e/functional tests pass as expected


app/views/components/listview/test-listview-set-page.html
```diff
@@ -44,14 +44,15 @@ <h2 class="card-title">Pagable Listview</h2>
...
+          activePage: 2,
           dataset: testData,
           paging: true,
           pagesize: 3,
           template: $('#project-tmpl').html()
         }).data('listview');
 
         //Set initial page
-        listview.pager.setActivePage(2, true);
+        //listview.pager.setActivePage(2, true);
```
tmcconechy: Do both ways work? Maybe add a functional test for each.

tmcconechy:
Some things i noticed.

1. http://localhost:4000/components/listview/example-paging-clientside.html -> this mode should not have first and last button. (Compare to master) if numbers then no first and last.

List View
1. http://localhost:4000/components/listview/example-paging.html-> this mode should not have first and last button. (Compare to master) if numbers then no first and last.
1. http://localhost:4000/components/listview/test-listview-performance-paging.html -> same
1. http://localhost:4000/components/listview/test-listview-set-page.html -> is not working (maybe because of the commented out line in the file i added, not sure if both ways will work). But here the page is no longer set to 2.
1. http://localhost:4000/components/listview/test-paging-indeterminate.html -> is not loading anything and also has the problem with the extra next/prev buttons
1. http://localhost:4000/components/listview/test-paging-pagesize.html -> also same problem with the extra next/prev butt
1. http://localhost:4000/components/listview/test-reload-paging.html -> also same problem with the extra next/prev button

Block Grid
- maybe add more items to the view for better testing (see https://github.com/infor-design/enterprise/pull/1603)

Datagrid
1. http://localhost:4000/components/datagrid/example-paging-page-size-selector.html -> is not working when you press toggle page selector its not toggling
1. http://localhost:4000/components/datagrid/example-paging.html -> filtering is not working try to filter on Product Id - contains - 14 and it only shows one page.
. http://localhost:4000/components/datagrid/test-datagrid-paging-indeterminate-initial.html -> paging is not working here
1. http://localhost:4000/components/datagrid/test-filter-paging-client-side.html -> filter on quantity : 1 and its also only showing the first page.
1. components/datagrid/test-paging-deferred.html -> this example is not showing pages after reloading 
1. http://localhost:4000/components/datagrid/test-paging-disable-lastpage.html -> this example is no longer disabling the last and next 
1. http://localhost:4000/components/datagrid/test-paging-empty-dataset.html _. set page size to 10 and press add 15 or so times. Its not showing pager in this scenario now. This is a bit like some of the other issues i feel.
1. components/datagrid/test-paging-force-disabled.html -> the disable/enable is no longer working here
1. http://localhost:4000/components/datagrid/test-paging-indeterminate-results.html -> this example is no longer showing results.
1. http://localhost:4000/components/datagrid/test-paging-multiselect-select-across-page.html as well as http://localhost:4000/components/datagrid/test-paging-multiselect.html -> these example is working but the results in the title show as undefined.
1. http://localhost:4000/components/datagrid/test-paging-select-indeterminate-multiple.html -> results also undefined
1. http://localhost:4000/components/datagrid/test-paging-select-serverside-multiple.html -> again results undefined
1. http://localhost:4000/components/datagrid/test-paging-select-serverside-single.html -> also undefined
1. http://localhost:4000/components/datagrid/test-paging-specific-page.html _> i dont know what this example is for can maybe be removed. But it is giving an error. Go last then first and then next (Cannot read property 'activePage' of undefined)

Other
1. paging no longer working on this example: http://localhost:4000/examples/businessinnovation/landing-page.html (Important as this team is very complainy)
1. paging not working on this pattern: http://localhost:4000/patterns/list-detail-paging.html check all the ones in http://localhost:4000/patterns/list* as they all seem to have the same issue (also important)

clepore:
Note: Removed review requests until its no longer `WIP`, then we can request again.

app/views/components/listview/test-listview-set-page.html
EdwardCoyle: I actually removed the comment I placed here because it was incorrect

EdwardCoyle:
Even though all the tests on this PR are currently passing, there are still a few things off.  Reserving this section for a list of those:

- [ ] Basically revisit each of [Tim's concerns](https://github.com/infor-design/enterprise/pull/1598#issuecomment-461148872) above.  After going back and forth on various ways to fix `indeterminate`/`preserveSelected`, I've re-broken some of these examples.
- [ ] If we really want to button this up, the items above should be added as e2e/functional tests.
- [ ] http://localhost:4000/components/pager/example-index - There is something to fix regarding DOM-defined Listviews.  I think we may need to add detection for these legacy Listviews and convert them to datasets to be compatible with the rendering changes.

EdwardCoyle:
Just a note for reviewers:  The big list of links with checkboxes above may or may not be accurate any more due to the merge from #464, which altered many Datagrid tests.

app/views/components/datagrid/test-paging-force-disabled.html
```diff
@@ -23,7 +23,7 @@ <h2>Datagrid Test: Force Disable the Pager</h2>
   <div class="twelve columns">
     <div class="toolbar">
       <div class="title">
-        This datagrid will refresh in <span id="reset-counter-time" style="font-family: Courier;"></span>ms.
+        &nbsp;
```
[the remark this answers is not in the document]
tmcconechy: No probably not - can you do an edit?

tmcconechy:
@davidcarlsonberg With regards to paging and selection we have some rules.

http://localhost:4000/components/datagrid/example-paging -> Will not save pages as its serverside paging 
http://localhost:4000/components/datagrid/example-paging-indeterminate -> this example shows the activation saved across page but not the selection (click checkbox vs row)

http://localhost:4000/components/datagrid/test-paging-multiselect-select-across-page.html -> this example shows an option to actually save across pages

So its "all the options".

app/views/components/datagrid/test-paging-force-disabled.html
davidcarlsonberg:
```suggestion
```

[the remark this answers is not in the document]
EdwardCoyle: Not sure exactly where that came in... part of the updates from `master` I believe

EdwardCoyle:
@davidcarlsonberg I addressed your point about filtering and incorrect pager buttons on http://localhost:4000/components/listview/test-listview-performance-paging.html.  Now, when you filter on something like "ri", you should see the pager's buttons actually conform to the result set... It was that way before but we may have borked it when getting the tests to pass.

screen shot 2019-02-22 at 2 31 09 pm

Would you and @deep7102 @tmcconechy mind a quick re-review?

PR #1602, opened 2019-02-06
1020 - Fix Date Editor blank on '00000000' date

### LINKED ISSUE #1020: Date editor shows '00000000' when data is blank
**Describe the bug**
When the dataGrid makes a cell editable, if  there is no cellText it fetches the field value.  The field value in our application has zeros for blank date values.  The date editor opens with zeros.

Below is the code that is being referenced.
```
    makeCellEditable: function(row, cell, event) {

      if (!this.isCellEditable(row, cell)) {
        return;
      }

      //Locate the Editor
      var col = this.columnSettings(cell);
      if (!col.editor) {
        if (event.keyCode === 32) {
          this.toggleRowSelection(this.activeCell.node.closest('tr'));
        }
        return;
      }

      // Put the Cell into Focus Mode
      this.setActiveCell(row, cell);

      var cellNode = this.activeCell.node.find('.datagrid-cell-wrapper'),
        cellParent = cellNode.parent('td'),
        cellValue = (cellNode.text() ? cellNode.text() : this.fieldValue(this.settings.dataset[row], col.field));

      if (cellParent.hasClass('is-editing')) {
        //Already in edit mode
        //Editor.focus
        cellNode.find('input').focus();
        return false;
      }
```

**To Reproduce**
1. Edit datagrid/example-editable so that the date or orderdate is  '00000000'
2. open  http://localhost:4000/components/datagrid/example-editable

**Expected behavior**
Expect if there is no cell text then  either an empty string is passed or the field value is formatted.

davidcarlsonberg:
Markup is not added for date editor when conditions are met for '0000', '000000', or '00000000'. Cell is not properly editable.

Closes #1020.

- Pull branch
- Run app
- Open app/views/components/datagrid/example-editable in an editor
  - Notice string in HTML is '00000000' in `orderDate` in for the first `data` entry
- Browse to http://localhost:4000/components/datagrid/example-editable
  - Cal icon should appear as per normal on hover and should be able to select a date. If no date is selected, empty string inhabits the cell. Cell is editable as well.


tmcconechy:
From the description of the issue it doesn't seem to work as expected.

1. Click in row 2 to edit
2. it shows 00000000 in the field. From what they are saying this should be treated as a blank string (no date). 

Also for the issue i noticed that they can intercept the value if they use this event: https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L7242 then they can get at the cell node and change the value if they want (was something similar to 3.5 they can use).

davidcarlsonberg:
It looks like there are actually multiple issues with the editor's handling of a string of zeroes. The proper markup is also missing to generate the `datepicker`.

tmcconechy:
When i was thinking about this issue i thought maybe we could handle it in locale.js so if you pass a string with all zeros into parseDate it could return `''`. Not sure if that helps but its a thought.

davidcarlsonberg:
`datagrid.formatters.js` did not have adequate condition logic to deal with an empty string. This has been added along with moving the logic to address `0000`, `000000`, `00000000` in `Locale.parseDate` and `Locale.formatDate`.

src/components/locale/locale.js
```diff
@@ -475,6 +480,11 @@ const Locale = {  // eslint-disable-line
       return undefined;
     }
 
+    if (dateString === '0000' || dateString === '000000' || dateString === '00000000') {
+      // Means no date in some applications
+      return '';
```
tmcconechy: Probably return undefined here https://github.com/infor-design/enterprise/blob/master/src/components/locale/locale.js#L468 as this is expected to be a date or undefined (or array for arabic)

src/components/datagrid/datagrid.editors.js
```diff
@@ -486,6 +486,7 @@ const editors = {
     };
 
     this.val = function (v) {
+      v = Locale.parseDate(v);
```
tmcconechy: I dont think we can do this as its reformatting the date. In this example http://localhost:4000/components/datagrid/example-editable click into the third row and its changing the format of the date.

Also parseDate only works on date objects. So if this is really needed maybe have to check if the type is a date object and make sure to use the same format as the column is using on render some logic around it based on whats set: https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.formatters.js#L63-L86

So i presume this might not be needed?

davidcarlsonberg: I've altered this to mutate a string of zeroes to an empty string, not using `parseDate`. If we don't handle that case, when you click on the cell to edit, the string of zeroes shows up.

davidcarlsonberg:
Will update CHANGELOG for this issue once #1611 is merged.

test/components/locale/locale.func-spec.js
```diff
@@ -877,4 +877,16 @@ describe('Locale API', () => {
     expect(Locale.parseDate('ذو الحجة 1439', 'MMMM yyyy')[2]).toEqual(1);
     Locale.set('en-US');
   });
+
+  it('Should parse or format a string of four, six, or eight zeroes', () => {
```
tmcconechy: 👍

PR #1623, opened 2019-02-12
457 - Handle Application Menu Accordion long text truncation

### LINKED ISSUE #457: Application Menu: Truncating of Text / Long Language Text Handling
**Older Details**
Original Discussion can be seen here:  https://jira.infor.com/browse/SOHO-5332

The function panel list on the left hand side has a fixed width, which can't be adjusted, and ellipses are added if the text extends beyond the width of the panel. This causes truncation issues when the function names are translated into different languages.  Many translations are longer than the English text and often the crucial bit of information regarding the function you want to select comes at the end and gets therefore cut off.

The function names are held in the SunSystems Data Dictionary (DOMN HIERARCHY NODES) which allow for a maximum length of 50 characters, but the panel is not wide enough to accommodate this.  In the pervious SunSystems version the old menu catered for this by displaying the menu panel based on text width, so this a regression of sorts introduced by UX4.

I have added some example screenshot for German, Italian, French, Spanish and Russian. Portuguese is similarly affected.  Chinese and Japanese not as much, as the translations tend to be shorter. A particularly bad example is the list of the different version of Corporate Allocation in German.  The function name starts with the German equivalent of ‘Corporate Allocations’ and the latter part denominates the type of the function you want to select.  In the screenshot attached you can’t actually read much further than ‘Corporate Allocations’, and it is ‘hit and miss’ as to which function you select. This is just one example. There are many others, for groups of functions and for single function names, where it is not possible to read the translated name sufficiently to make the correct function selection.

davidcarlsonberg:
Proposes a fix for the truncation of long text in an accordion in an applicationmenu. Tooltips are applied to `.accordion-header a span` based on logic determining the text is truncated.

**Related github/jira issue (required)**
Closes #457.

**Steps necessary to review your pull request (required)**
- Pull branch
- Run app
- Browse to http://localhost:4000/components/applicationmenu/test-tooltips.html
- Hover over nested accordion elements in the applicationmenu to cause a tooltip to appear if the text is truncated.


app/views/components/applicationmenu/test-long-content-string.html
```diff
@@ -88,7 +88,7 @@
         </div>
         <div class="accordion-pane">
           <div class="accordion-header list-item">
-            <a href="#"><span>Role #2 Task #1</span></a>
+            <a href="#" title="Role #2 Task #1"><span>Role #2 Task #1</span></a>
```
tmcconechy: Do we need a title here?

src/components/accordion/_accordion.scss
```diff
@@ -522,6 +522,10 @@
     padding: 13px 0 13px 20px;
     white-space: nowrap;
     width: calc(100% - 1px);
+
+    > span {
+      display: inline-block; // to allow js width calculation
+    }
   }
 
   > [class^='btn'] {
```
tmcconechy: Is this an empty rule?

app/views/components/applicationmenu/test-tooltips.html
```diff
@@ -0,0 +1,179 @@
+{{> includes/head}}
+
+<style>
```
tmcconechy: Do we still need these? I thought that the styles where now in the core scss?

EdwardCoyle: @davidcarlsonberg there's a folder of SASS files in the demoapp's `src` folder that gets pre-compiled whenever it runs:  https://github.com/infor-design/enterprise/tree/master/app/src/sass.  We started doing this instead of using `style` tags in the tests because those tags get flagged by the Veracode scans sometimes.

app/views/components/applicationmenu/test-long-content-string.html
davidcarlsonberg:
```suggestion
            <a href="#"><span>Role #2 Task #1</span></a>
```

app/views/components/applicationmenu/test-tooltips.html
davidcarlsonberg: Nope. Will remove as it's in core scss now.

docs/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - `[Datepicker]` Fixed disabled dates example to validate again on disabled dates. ([#1445](https://github.com/infor-design/enterprise/issues/1445))
 - `[Dategrid]` Fixed a Date Editor bug when passing a series of zeroes to a datagrid cell with an editable date. ([#1020](https://github.com/infor-design/enterprise/issues/1020))
 - `[Icons]` Established missing icon sourcing and sizing consistency from ids-identity icon/svg assets. ([PR#1628](https://github.com/infor-design/enterprise/pull/1628))
+- `[Stepprocess]` Fixed a bug where the step folder was still selected when it was collapsed or expanded. ([#1633](https://github.com/infor-design/enterprise/issues/1633))
```
davidcarlsonberg: Fixes a merge conflict.

src/components/accordion/accordion.js
```diff
@@ -1488,6 +1491,85 @@ Accordion.prototype = {
...
     return this;
+  },
+
+  /**
+   * Hide the visible tooltip.
+   * @private
+   * @returns {void}
+   */
+  hideTooltip() {
+    if (self.tooltip) {
```
deep7102:
```suggestion
    const self = this;
    if (self.tooltip) {
```

PR #1624, opened 2019-02-13
## - Updating Datagrid column hardcoded 0 value issue

spikeeedge:

The ColumnsFromString function has hadcoded column value "0" instead dynamic value.
This causes the issue when we revert the datagrid changes

When you call the function resetColumns() of the dataGrid, the function is using the variable "originalColumns" which reads the columns from Settings of the datagrid.

 ```this.originalColumns = this.columnsFromString(JSON.stringify(this.settings.columns));```

columnsFromString function loops through the columns present in given string and saves as an object value. this process saves oth Column Id in ith Column object.

```$.extend(columns[i], orgColumn[0]);```

1. Set a default row into the grid on page load
1. Add a new row into the grid by adding "New" button which will call "addRow()" function
1. Call Undo function to remove added row by calling resetColumns() function
    - Result:
    The grid will be restored with first(0th) Column value in all columns
    - Expected Result:
    The grid should be restored with default column values (not by first column values)

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -4096,7 +4096,7 @@ Datagrid.prototype = {
       if (orgColumn) {
         isHidden = columns[i].hidden;
 
-        $.extend(columns[i], orgColumn[0]);
+        $.extend(columns[i], orgColumn[i]);
 
         if (isHidden !== undefined) {
           columns[i].hidden = isHidden;


spikeeedge:
Hi Tim,

Thanks for quickly looking into this.

Unfortunately, after doing really deep dive into the code I found the our
project was slightly managing the grid functionality wrong - my bad.
So sorry about this. you can reject the commit. The soho code is working as
expected.

Thank you

tmcconechy:
OK Thanks @spikeeedge will just close then.
Looking forward to more future contributions :thumbs

PR #1585, opened 2019-01-31
Create SVG icon HTML files from ids-identity assets

clepore:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

*Use Icon assets from ids-identity@2.1*
- Create the svg icon files (`svg.html`, `svg-extended.html`, `svg-empty.html`) dynamically from the svg files in the ids-identity node module.
- **Most of the file changes are from this**->  Icons will use the `color` css property to change their stroke color, as `fill` will actually refer to the non-stroke color (background, etc). This will be nice as we can then use css color inheritance and probably remove some instances where we specify icon `color` (previously `fill`).

*Icon Examples*
- Icon example files are now dynamically created from the assets in ids-dentity. We no longer have manually update these files.

**Related github/jira issue (required)**:
https://github.com/infor-design/design-system/issues/247

**Steps necessary to review your pull request (required)**:
1. `npm i && npm run start`
2. Test the following pages and make sure all icons show up and compare to the current release:
    - [index.html - 195 icons](http://localhost:4000/components/icons/example-index.html)
    - [extended.html - 194 icons](http://localhost:4000/components/icons/example-extended.html)
    - [empty-widgets.html - 12 icons](http://localhost:4000/components/icons/example-empty-widgets.html)
3. Examine other icon heavy examples to make sure they colorize and hover as they should.
    - [kitchen sink](http://localhost:4000/)
    - [calendar](http://localhost:4000/components/calendar/example-index.html)
    - [ratings](http://localhost:4000/components/rating/example-index.html)

PR #1628, opened 2019-02-13
N/A - Create SVG icon HTML files from ids-identity assets (cont...)

clepore:

*Use Icon assets from ids-identity@2.1*
- Create the svg icon files (`svg.html`, `svg-extended.html`, `svg-empty.html`) dynamically from the svg files in the ids-identity node module.
- **Most of the file changes are from this**->  Icons will use the `color` css property to change their stroke color, as `fill` will actually refer to the non-stroke color (background, etc). This will be nice as we can then use css color inheritance and probably remove some instances where we specify icon `color` (previously `fill`).

*Icon Examples*
- Icon example files are now dynamically created from the assets in ids-dentity. We no longer have manually update these files.

https://github.com/infor-design/design-system/issues/247

1. `npm i && npm run start`
2. Test the following pages and make sure all icons show up and compare to the current release:
    - [index.html - 195 icons](http://localhost:4000/components/icons/example-index.html)
    - [extended.html - 194 icons](http://localhost:4000/components/icons/example-extended.html)
    - [empty-widgets.html - 12 icons](http://localhost:4000/components/icons/example-empty-widgets.html)
3. Examine other icon heavy examples to make sure they colorize and hover as they should.
    - [kitchen sink](http://localhost:4000/)
    - [calendar](http://localhost:4000/components/calendar/example-index.html)
    - [ratings](http://localhost:4000/components/rating/example-index.html)

---

*Associated PR at https://github.com/infor-design/enterprise/pull/1585*


app/src/js/routes/custom-icons.js
```diff
@@ -0,0 +1,57 @@
+// Icons Route
+//= ====================================================
+const fs = require('fs');
+const path = require('path');
+const handlebars = require('handlebars');
```
tmcconechy: Did'nt catch this before - but we are using both handlebars and mustache templates. (Module mmm on both server and client side)
Something to maybe adjust later to make it more consistent

nickwynja:
@tmcconechy Thanks for taking such a close look. All of those changes are intentional with the exception of the background color icon size and the pager icon alignment. I'll push a fix for these. The intended changes were to standardize the icon sizes and be using the "real" sizes of the icons as designed.

I just noticed this as well: 

image

but it seems to be in the current version: https://4150-enterprise.demo.design.infor.com/components/calendar/example-index

If that's accurate, can you take a look after this is merged in?

tmcconechy:
OK, What about the rating focus state alignment? Its a bit off.
I also think that the chevron and arrow size is a bit large though. It looks way different.

nickwynja:
@tmcconechy Based on the way we redid the current icons out of Sketch, we couldn't keep the icon having a path with a stroke and a fill, so we're using the outlined icon for the empty star. Because we lost the path in this instance, we're using `border` and `box-shadow`. It's maybe not quite as pretty but uses the icons more consistently and satisfies having a focused state.

Looking back at the previous icon size of the chevrons, they seemed a little small. Again we favored consistency in size rather than keeping it as is. We can get @kayiuho's input here:

## Current:

screen shot 2019-02-14 at 10 09 55 am
screen shot 2019-02-14 at 10 12 38 am


## New:

screen shot 2019-02-14 at 10 09 44 am
screen shot 2019-02-14 at 10 12 26 am

kayiuho:
Yea, I agree, having a consistent sizing is important for the icon system as a whole. The chevrons in the previous design do seem a little small. (And is it just me or the down chevron darker?)

nickwynja:
> And is it just me or the down chevron darker?

You're right, it does look darker here. It appears to just be the screenshot though. Must have been a weird hover state captured in the screenshot.

tmcconechy:
Ok sounds good - i guess i'll get used to it. Was just very obvious.
Looked again and the rating and its fine too. So let me know when your all done and i'll approve.
Maybe @deep7102 and @davidcarlsonberg can also have a quick test

PR #1632, opened 2019-02-14
N/A - E2E change autosave option to true for debug configs

clepore:
When doing e2e testing and resetting the baseline images, the cli returns: `Failed: Image not found, saving current image as new baseline.` but it never creates the baseline.

n/a

none


test/protractor.local.debug.conf.js
```diff
@@ -29,7 +29,7 @@ exports.config = {
     browser.protractorImageComparison = new protractorImageComparison({
       baselineFolder: `${basePath}/baseline`,
       screenshotPath: `${basePath}/.tmp/`,
-      autoSaveBaseline: false,
+      autoSaveBaseline: true,
```
tmcconechy: Will have no effect since the tests all have `if utils.isCi` so doesnt matter either way.

PR #1637, opened 2019-02-14
1413 - Added support to preview mode with editor

### LINKED ISSUE #1413: Rich Text Editor: Disabled state shouldn't interfere with RTE internal content styling
**Describe the bug**
When the editor control is set to disabled, a border is displayed and the height of the control is maintained regardless of content.

**To Reproduce**
1. Extract and run the attached test-editor-states.html

**Expected behavior**
Would like to see no border and the control vertically size to the content.

**Version**

v4.13.0

**Screenshots**

Currently:
image

Expectation:
image

[test-editor-states.zip](https://github.com/infor-design/enterprise/files/2686997/test-editor-states.zip)

**Platform**

**Additional context**
An additional setting to indicate to not set the border and vertically size the content would be acceptable so as to maintain backward compatibility.

deep7102:
Added new state called “preview” a non editable mode to editor. Where it only shows the HTML with no toolbar, borders etc.

Closes #1413

http://localhost:4000/components/editor/example-preview.html
- Open above link
- See the editor it should be not editable
- Click top button "Change to Editable Mode"
- Apply any changes to editor content (text color, bold, block quote etc.)
- Click top button "Change to Preview Mode"

### DIFF
--- a/src/components/editor/editor.js
+++ b/src/components/editor/editor.js
@@ -33,6 +33,7 @@ const COMPONENT_NAME = 'editor';
 * @param {string} [settings.image = { url: 'https://imgplaceholder.com/250x250/368AC0/ffffff/fa-image' }] Info object to populate the image dialog defaulting to ` {url: 'http://lorempixel.com/output/cats-q-c-300-200-3.jpg'}`
 * @param {function} [settings.onLinkClick = null] Call back for clicking on links to control link behavior.
 * @param {function} [settings.showHtmlView = false] If set to true, editor should be displayed in HTML view initialy.
+* @param {function} [settings.preview = false] If set to true, editor should be displayed in preview mode non editable content.
 */
 const EDITOR_DEFAULTS = {
   buttons: {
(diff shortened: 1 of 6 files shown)


deep7102:
@tmcconechy @EdwardCoyle 
Updated now, example page with extra content and toggle mode button

src/components/editor/editor.js
```diff
@@ -33,6 +33,7 @@ const COMPONENT_NAME = 'editor';
 * @param {string} [settings.image = { url: 'https://imgplaceholder.com/250x250/368AC0/ffffff/fa-image' }] Info object to populate the image dialog defaulting to ` {url: 'http://lorempixel.com/output/cats-q-c-300-200-3.jpg'}`
 * @param {function} [settings.onLinkClick = null] Call back for clicking on links to control link behavior.
 * @param {function} [settings.showHtmlView = false] If set to true, editor should be displayed in HTML view initialy.
+* @param {function} [settings.preview = false] If set to true, editor should be displayed in preview mode non editable content.
```
davidcarlsonberg:
```suggestion
* @param {function} [settings.preview = false] If set to true, editor should be displayed in preview mode with noneditable content.
```

PR #1635, opened 2019-02-14
464 - Add Datagrid Frozen Columns

tmcconechy:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

This PR adds a really cool frozen column feature to the datagrid. This has an impact on columns/column rendering since new containers are added. While fixing this i fixed other column rendering issues as listed. Also cleaned up the test pages a lot removing old ones or dups or one that did not make sense and added more tests for datagrid example pages.

Main impact of frozen columns is things like

- column rendering
- column widths
- column resize
- sort
- export
- plus integration with various other features

**Frozen Columns Works with**
* Alternate row shading
* Drilldown (can be frozen on the left)
* Export
* Selection (can freeze on the left)
* Row Status
* Data Grouping and Totals
* Empty Message (will stay centered)
* Summary Row (you can freeze columns on the left)
* Editing
* Expandable Row -> But the expandable row is only in the middle section
* Filtering - Filtered columns can be frozen
* Row Activation

**Frozen Columns Limitations**
* Wont work with columns with colspans
* Wont work with dynamically height rows (example-comments.html)
* You can’t resize frozen columns
* Frozen columns cant be moved or hidden from the personalize dialog
* Ids are required for the frozen columns and should be used in general to avoid bugs
* If freezing the group column all the group by columns will be frozen
* If the total sum of the frozen columns is wider than the device - then it's not possible to scroll. Will want a future solution
* Percent Columns - Percent column widths will not work on frozen sections at this time.
* Will not work with expandOnActivate setting
* Grouped Headers will not work with frozen columns at this time
* Row Reorder will not work with frozen columns at this time
* Will not work with nested grids
* Will not work with clickToSelect setting
* Will not yet work with tree grid

NOTE: If changing the mac setting "show scrollbars: always vs when scrolling". You need to reload the page to rerender properly.

**Related github/jira issue (required)**:
Closes #464 
Closes #1246
Closes #1560
Closes #1411
Closes #1191
Closes #1481
Closes #1456

**Steps necessary to review your pull request (required)**:

**Reviewing #464** 
http://localhost:4000/components/datagrid/example-frozen-columns
- scrolll up and down and side to side in the middle section
- try things like arrowing around, sorting, paging, export, personalize column (anything in the page)
- you may need to resize the browser less wide to see scrolling

Other pages showing some integration with other features
http://localhost:4000/components/datagrid/example-editable
http://localhost:4000/components/datagrid/example-empty-message
[...]

PR #1642, opened 2019-02-18
1641 - selectRow and unselectRow can alter a checkbox that is not the…

### LINKED ISSUE #1641: selectRow and unselectRow can alter a checkbox that is not the selectionCheckbox
**Describe the bug**
The method to find the selectionCheckbox can result in a cellnode that is not the selection checkbox from being selected and then altered.

If there is no selectionCheckbox columns then, self.columnIdxById('selectionCheckbox') will return -1, but the cellNode method will return the last column node for the row. If this is a checkbox then the checkbox will change.

self.cellNode(elem, self.columnIdxById('selectionCheckbox'));

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/example-singleselect.html'
2. Click on any row to select the row

**Expected behavior**
Only the selectionCheckbox should change and not any other checkbox on the grid

**Version**
4.16.0

**Screenshots**
image

marknolaninfor:
… selectionCheckbox

The method to find the selectionCheckbox can result in a cellnode that is not the selection checkbox from being selected and then altered.

If there is no selectionCheckbox columns then, self.columnIdxById('selectionCheckbox') will return -1, but the cellNode method will return the last column node for the row. If this is a checkbox then the checkbox will change.

self.cellNode(elem, self.columnIdxById('selectionCheckbox'));

Closes #1641 

* Go to 'http://localhost:4000/components/datagrid/example-singleselect.html'
* Click on any row to select the row

### DIFF
--- a/app/views/components/datagrid/example-singleselect.html
+++ b/app/views/components/datagrid/example-singleselect.html
@@ -15,13 +15,13 @@
         data = [];
 
       // Some Sample Data
-      data.push({ id: 1, productId: 2142201, productName: 'Compressor', activity:  'Assemble Paint', quantity: 1, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action'});
-      data.push({ id: 2, productId: 2241202, productName: '1 Different Compressor', activity:  'Inspect and Repair', quantity: 2, price: 210.99, status: '', orderDate: new Date(2015, 7, 3), action: 'On Hold'});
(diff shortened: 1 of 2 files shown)


tmcconechy:
We will need to test this after https://github.com/infor-design/enterprise/pull/1635 and https://github.com/infor-design/enterprise/pull/1598 are merged for any merge conflicts needing resolution

tmcconechy:
@davidcarlsonberg that checkbox is "readonly" i think thats the point. It is not editable so you cant interact with it. Before this fixed it would get changed when you select since there are no selection checkboxes at the front of the grid

PR #1648, opened 2019-02-20
1629 - Landmark/krishollenbeck redesigned hierarchy.js

clepore:
See original pull request https://github.com/infor-design/enterprise/pull/1630 for details.
---
⭐️ Credit and code go to @krishollenbeck and @pwpatton ⭐️


nickwynja:
@kayiuho @beccakroll Can you please take a look at the new stacked layout for the Org Chart control and let us know if you think anything is missing or weird? https://demo-krishollenbeck-master-enterprise.demo.design.infor.com/components/hierarchy/example-stacked.html

nickwynja:
Based on reviewing with Becca and the rest of the design team, here are some UX issues (and possibly bugs) we're seeing. Some might be specific to this coded example:

1. The down/up carets are confusing. In general, we don't think it's the right visual metaphor. The example in Teams/Outlook has a more obvious focus state and a [clear call-out (sizewise)](https://user-images.githubusercontent.com/1012017/53109016-09ae3b00-3506-11e9-9eb8-e77dcdc155cd.png) when you're focused on a manager. We'll take a stab at an improved design today.
1. The ellipsis actions should be on every card and persistent. Currently, they seem to come and go as they please
1. Click in to `Greg Ford > Nicolas Grant > Nicole Cargill`; the caret indicates that Nicole would have reports but when you expend, there aren't any. Feature or bug? Either way, it's a confusing experience right now when you expand Nicole but it only just focuses her.
    * It would be ideal to show some sort of indicator that an employee has reports, even if you're not able to count how many reports they have
    * A more consistent approach could be to Center the "focused" person _always_. So if you search and someone has no reports, they'd be larger and focused

Maybe we just get on a call early this afternoon to review some of this?

pwpatton:
The content of the org chart sample is completely implementation specific. we also made a bar-bones sample to just show the boxes and lines using the same data but not really showing any content. So the idea is that if the blocks and lines are good then its' really just what we want to present to the implementing groups what we'd consider a "best practice" through the samples.

https://demo-krishollenbeck-master-enterprise.demo.design.infor.com/components/hierarchy/example-minimal.html

PR #1652, opened 2019-02-21
Update hierarchy sample

pwpatton:
moved landmark specific stacked hierarchy demo to examples/landmark
added a simplified stacked hierarchy sample

- run http://localhost:4000/components/hierarchy/example-stacked.html
it should no longer show the ... menu button and should navigate using a single click anywhere on the card.


krishollenbeck:
This https://github.com/infor-design/enterprise/pull/1653/ should fix that failing test.

Issue #1629, opened 2019-02-13
Hierarchy redesign. New "stacked" design implementation


krishollenbeck:
**Describe the solution you'd like**
- Creating a stacked hierarchy view that is similar in functionality and flow to ms teams.
- Add deprecation notifications and notes for future deprecation of older hierarchy layouts.

New control will support existing features and continue to support older versions of hierarchy layout until a future TBD date.

Should still provide and support the following functionality:

1. Responsive behavior
2. A visual indicator for paging throw nodes (expand/collapse)
3. Menu button to display more details and a context menu

**Additional context**
![hierarchy-stacked](https://user-images.githubusercontent.com/1056684/52739784-42df2c00-2f97-11e9-8815-51710cc7cf53.png)

tmcconechy:
Moving this into the sprint for QA

CindyMercadoReyes:
This issue is now resolved.

PR #1656, opened 2019-02-22
1629 - Updating hierarchy control to make sure dataset is an array and not an object

krishollenbeck:
There are inconsistencies between the new stacked layout vs the current existing layouts. The stacked layout examples were passing in data as an object vs an array. Allowing an object causes issues with the Angular wrapper which expects the data set to be an array. To resolve this I am updating the control to make sure only an array is passed in as data.

https://github.com/infor-design/enterprise/issues/1629

Navigate through existing examples. http://localhost:4000/components/hierarchy/example-stacked.html Everything should still work as expected. All tests should pass.

### DIFF
--- a/src/components/hierarchy/hierarchy.js
+++ b/src/components/hierarchy/hierarchy.js
@@ -99,20 +99,12 @@ Hierarchy.prototype = {
 
     // Safety check, check for data
 
-    if (s.dataset === undefined) {
-      console.error('Hierarchy dataset is undefined.');
-      return;
-    }
-
-    if (!Array.isArray(s.dataset)) {
-      console.error('Hierarchy dataset must be an Array');
-      return;
-    }
-
-    if (s.dataset.length === 0) {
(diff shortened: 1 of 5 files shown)


src/components/hierarchy/hierarchy.js
```diff
@@ -101,22 +101,22 @@ Hierarchy.prototype = {
...
-      if (s.dataset.length === 0) {
-        this.element.emptymessage(s.emptyMessage);
-        return;
-      } else if (s.dataset[0] && s.dataset[0].children.length > 0) {
-        this.render(s.dataset[0]);
-      } else if (s.dataset && s.dataset.children.length > 0) {
-        this.render(s.dataset);
-      }
+    if (!Array.isArray(s.dataset)) {
+      console.error('Hierarchy dataset must be an Array');
```
tmcconechy: I suggest we just return here with no error as the console.errors dont pass the linter or if you really want them can use console.warn as we will allow console.warn.

PR #1664, opened 2019-02-25
Toolbar Flex Component: support ajax callback on more menu

### LINKED ISSUE #1663: Toolbar Flex Component: Support Ajax callbacks for the more menu
**Describe the bug**
Passing in a ajax callback for the more menu on a flex toolbar doesn't work. The callback isn't ever called.

**To Reproduce**
1. Copy the attached [example-more-actions-ajax.zip](https://github.com/infor-design/enterprise/files/2901280/example-more-actions-ajax.zip) to view/components/toolbar-flex directory and navigate to it in the brower
2. Open the more menu.
3. See that the the html menu is displayed, not the ajax callback one

**Expected behavior**
The ajax callback should be called.

**Version**

- ids-enterprise: v4.16.0-beta.0

**Screenshots**
Problem: 
image

What should be seen: 
image

**Platform**
 - All

Sovia:
Adding the ability to pass in a user-defined `beforeOpen()` callback method on the more menu.

"Closes #1663"

1. Navigate to http://localhost:4000/components/toolbar-flex/example-more-actions-ajax
2. Open the more menu and see that the entries have 'AJAX Option' in the labels
image
3. Shrink the window so the 'Menu Button' is overflowed into the more menu
4. Ensure that the Menu Button in the overflowed menu matches the entries from the Menu Button in the button set
image
image

Also added a e2e test which uses image comparison for the above test case, toolbar-flex.e2e-spec.js


docs/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 - `[Lookup]` Fixed a bug with settings: async, server-side, and single select modes.  The grid was not deselecting the previously selected value when a new row was clicked.  If the value is preselected in the markup, the lookup modal will no longer close prematurely. ([PR#1654](https://github.com/infor-design/enterprise/issues/1654))
 - `[Tree]` Fixed a bug when calling the disable or enable methods of the tree. This was not working with ie11. ([PR#1600](https://github.com/infor-design/enterprise/issues/1600))
 - `[Stepprocess]` Fixed a bug where the step folder was still selected when it was collapsed or expanded. ([#1633](https://github.com/infor-design/enterprise/issues/1633))
+- `[Toolbar Flex]` Fixed a bug where an ajax callback couldn't be configured for the more menu
```
EdwardCoyle:
```suggestion
- `[Toolbar Flex]` Added the ability to pass in a `beforeOpen` callback to the More Actions menu (fixes a bug where it wasn't possible to dynamically add content to the More Actions menu in same way that was possible on the original Toolbar component)
```

src/components/toolbar-flex/toolbar-flex.item.js
```diff
@@ -52,7 +52,8 @@ const TOOLBAR_FLEX_ITEM_DEFAULTS = {
   disabled: false,
   readOnly: false,
   hidden: false,
-  componentSettings: undefined
+  componentSettings: undefined,
+  beforeOpen: undefined
```
EdwardCoyle: Not sure this is used anywhere now, since `componentSettings` handles the callback AND other settings.

docs/CHANGELOG.md
Sovia: I'll update this!

src/components/toolbar-flex/toolbar-flex.item.js
Sovia: You are right, it's not needed. Removing it!

Issue #1688, opened 2019-02-27
Datagrid serverside filtering regression


Mincol:
**Describe the bug**
While server side filtering is configured, changing multi select and single select combo box filters did not called server for new Datagrid content

**To Reproduce**
1. Go to http://localhost/demo/components/datagrid/example-paging
2. Click on multi select filter (currently not in place, I intend to enrich example as part of pull request)
3. The data was not filtered and no call to server was made

**Expected behavior**
The function Datagrid.settings.source should be called, which will issue ajax call and grid is refreshed with new data.
The filter is still visible.

**Version**
On master since Dec 17, 2018
Version  4.14.0 (?) till current latest 4.15.0

**Additional context**
Regression caused by https://github.com/infor-design/enterprise/pull/1416.
Change prevents calling Datagrid.settings.source function, but intended change was just to prevent calling: 
```
 /**
         * Fires after a filter action ocurs
         * @event filtered
         * @memberof Datagrid
         * @property {object} event The jquery event object
         * @property {object} args Object with the arguments
         * @property {number} args.op The filter operation, this can be 'apply', 'clear'
         * @property {object} args.conditions An object with all the condition data.
         * @property {string} args.trigger Info on what was the triggering action. May be render, select or key
         */
        this.element.trigger('filtered', {op: 'apply', conditions: conditions, trigger: trigger});
```
Why it worked for test fields I do not understand. Maybe fix should be in different place? Or refactoring is needed?

tmcconechy:
Added to the sprint for later QA.

brianjuan:
https://4170-beta0-enterprise.demo.design.infor.com/components/datagrid/test-disable-client-filter-and-sort.html
Passed QA testing on all browsers following steps in #1745 , moving to Done.

PR #1689, opened 2019-02-27
1688 - Ensure that fn Datagrid.settings.source is called

Mincol:
I want to ensure that function Datagrid.settings.source is called in all cases, so server side code can act on any change done by user.

#1688

- Check that hhttp://localhost:4000/components/datagrid/example-paging was enriched by all types of column filters (multi select, single select)
- Check why text filters work even without this fix + Check why sorting is not experiencing same problem,

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -2076,20 +2076,20 @@ Datagrid.prototype = {
 
     if (this.restoreFilterClientSide) {
       this.restoreFilterClientSide = false;
-      return;
+    } else {
+      /**
+      * Fires after a filter action ocurs
+      * @event filtered
+      * @memberof Datagrid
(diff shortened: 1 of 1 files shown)


tmcconechy:
This looks like the fix we discussed but i dont think the steps to reproduce and test are right on the PR. I think what you mean is to test:

http://localhost:4000/components/datagrid/test-disable-client-filter-and-sort.html
- make sure that the sort indicator stays
- filter and make sure the filter stays.

So the fix looks good. Just update the steps to test and you can remove the draft on the PR.

tmcconechy:
Im closing this issue but i will create a new pull request based on what we discussed and the fix.

tmcconechy:
Moved this to https://github.com/infor-design/enterprise/pull/1745 for you

PR #1715, opened 2019-03-02
1608 - Datagrid tabbing into editable cells

### LINKED ISSUE #1608: Datagrid - tabbing into editable cells of different rows with inlineEditor=true not working correctly
We have columns in Datagrids with inlineEditor = true and when we tab through the rows, even though the focus changes, the row is not selected correctly.

This is the SOHO example that has it: 
https://latest-enterprise.demo.design.infor.com/examples/saleshub/product-search.html

**To Reproduce**
1)	Select the qty cell of say, row1
2)	Tab through the cells/rows to say, qty cell of row3
3)	Change qty to 8 
4)	You will notice that it has changed the qty of row1

**Expected behavior**
Qty of row3 should change to 8

**Platform**
- Browser Name [e.g. chrome, IE, safari, stock browser]
- currently using SOHO version 4.12.1 but seems to be an issue in the latest one too

davidcarlsonberg:
Tabbing into editable cells with `inlineEditor=true`does not allow editable of newly focused cell.

Closes #1608.

- Pull branch
- Run app
- Browse to http://localhost:4000/examples/saleshub/product-search.html
- Click into Row 0, Cell 2 (Quantity)
- Change value
- Tab through to Row 1, Cell 2 (Quantity)
- Change value (value should change appropriately)


davidcarlsonberg:
I don't believe there are any tests for these pages (app/views/examples)...happy to write one if desired?

marknolaninfor:
This change breaks tabbing through the actionable grid items. 
1. goto '/components/datagrid/test-editable-actionable-mode.html"
2. Click on the date cell
3. use Shift-Tab and Tab to move through the cells within the datagrid

The cells are not inaccessible 
image

There needs to be a check to see if the column has an inline editor, so that it is not applied to other columns.

{code}
if (key === 9 && self.editor && self.editor.name === 'input' && col.inlineEditor === true) {
{code}

PR #1720, opened 2019-03-04
1718 - Datagrid is-not-empty filter on null

### LINKED ISSUE #1718: Datagrid date filter does not filter 'null' values on 'is not empty'
**Describe the bug**
When filtering 'is not empty' dates on datagrid, "null" values are still being shown. Whereas empty string values are being filtered out.

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/components/datagrid/test-filter-editable.html'
2. On Row 6, remove the 'Order Date' value
image
3. Row 6 should have an empty string value ("") and Row 7 has a null value.
image
4. Filter our the Order Date using the "Is Not Empty" option.
image

**Expected behavior**
Both Row 6 and Row 7 should not be shown anymore since they are "empty" and we are filtering out values that are "not empty".

**Actual Behavior**
Row 7 is still shown.
image

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome

davidcarlsonberg:
Datagrid was not filtering properly on a `null` value in Order Date.

Closes #1718.

- Pull branch
- Run app
- Goto: http://localhost:4000/components/datagrid/example-filter.html
- Filter on Order Date with `is-not-empty` filter (will show 7 results with id: 1, 2, 3, 4, 5, 5, 6)
- Should NOT show two results: one `''` and one `null`.


deep7102:
@davidcarlsonberg 🍌 Looks good! Updated the steps in description to test.

Issue #1699, opened 2019-02-27
Form-compact: New form-compact initial foundation


clepore:
**Is your feature request related to a problem? Please describe.**
This is the foundation for the form-group visual updates.

**Describe the solution you'd like**
Build the foundation for the new form-compact component per the design skeleton attached. The code should be production worthy and be considered the modular base infrastructure for future enhancements (like dropdowns, masking, dirty states, searches, lookups - some of which can be referenced in infor-design/design-system#248).

Please feel free to reach out to anyone with questions and leave comments on this issue as necessary.

Architecture:
- [x] Build it as a new component `form-compact`
    - sass, js, readme.md, etc
- [ ] Make sure the component works with the custom builder
- [x] Provide a full width `example-index.html` demo page
- [x] Add basic documentation to `/readme.md` as necessary
- [x] Ignore writing any tests for now

Code/Development:
- [x] Be sure to use a 12-col grid without gutters for responsiveness
- [x] Be conscious of focus state & tabIndex
- [x] Note the ellipsis for test and labels
- [x] All section headers should have the gray background to match "section three" 
    - Note this is *not* represented in the design. That's okay.

**Describe alternatives you've considered**
n/a

**Additional context**
![mock compact v4 white bg](https://user-images.githubusercontent.com/742319/53523551-1ead4000-3aab-11e9-9491-dc46225ff174.png)

EdwardCoyle:
@clepore @nickwynja - Branch available at https://github.com/infor-design/enterprise/compare/1699-form-compact with some preliminary work.  If you check this branch out you can see the basic form pretty much working:

- http://localhost:4000/components/form-compact/example-index.html?font=source-sans (the exact same layout as the PNG)
- http://localhost:4000/components/form-compact/example-responsive-flow?font=source-sans (slightly different HTML structure but still uses the 12-column grid.

I laid out the second one a different way because the first attempt does not flow the section headers correctly on a responsive layout.  As of writing this, I'm still trying to figure out if we can eliminate the extra space at the bottom of Section 1, but I'd probably have to do some funny things with the columns... not sure about it.

nickwynja:
@kayiuho Should there be any placeholders in this design?

CindyMercadoReyes:
All Form-compact components needed to be tested.

URL: https://4170-beta0-enterprise.demo.design.infor.com/components/form-compact/list

janahintal:
[...]

PR #1728, opened 2019-03-05
Demoapp: Remove extra `head` templates

EdwardCoyle:
While working on #1699 I found a problem in the demoapp where an outdated `app/views/includes/head-patterns.html` was being used, and was trying to pull in CSS files that no longer exist.  This file was pretty much the same as `app/views/includes/head.html` and was missing some previous changes we made to the headers, so I took the duplicate ones out, and re-worked existing examples and router code to use the same one.

Additionally, I made a change to the demoapp that prevents a layout set through the `custom-route-options.js` middleware from being overwritten by other middleware.  This fixes issues on some test pages where the layouts were being loaded multiple times

Related to #1699

Pull this branch and run the demoapp. Then:
- Open http://localhost:4000/patterns/list.  Everything should render properly
- Open any test page in that list.  There should be only one blue header present on the page (no double rendering)
- Open other pages with custom-set layouts, such as http://localhost:4000/components/searchfield/example-header-compact.html to ensure their layouts are not being overridden.  See the [full list of these here](https://github.com/infor-design/enterprise/blob/master/app/src/js/custom-route-options.js#L19-L76).


deep7102:
Found some broken

Double Rendering (looks like all list-detail pages)
http://localhost:4000/patterns/list-detail-datepicker.html
http://localhost:4000/patterns/list-detail-mixed-selection.html
http://localhost:4000/patterns/list-detail-no-split-header-toolbar.html
http://localhost:4000/patterns/list-detail-paging-indeterminate.html
http://localhost:4000/patterns/list-detail-paging-pagesize-selector.html
http://localhost:4000/patterns/list-detail-paging.html
http://localhost:4000/patterns/list-detail-popup-search-no-split-header.html
http://localhost:4000/patterns/list-detail-search-no-split-header.html
http://localhost:4000/patterns/list-detail-search.html
http://localhost:4000/patterns/list-detail.html

No Header, not sure these should be without header or it's just missing (same in master)
http://localhost:4000/patterns/master-detail.html
http://localhost:4000/patterns/tree-detail.html

Other
http://localhost:4000/patterns/full-width.html
http://localhost:4000/patterns/iframe-with-module-tabs.html

EdwardCoyle:
In this last push, I fixed all on both @deep7102 and @tmcconechy 's lists, except for these:
- http://localhost:4000/patterns/navigation-tabs-outbound-page2.html
- http://localhost:4000/patterns/navigation-tabs-outbound-page1.html

I think there's actually a bug in the "Style-only" case on Header Tabs... will explore further

EdwardCoyle:
Raised #1746 for the Tabs navigation test issues

PR #1733, opened 2019-03-05
Fixes #1731 - Refactor for escape and tab key handling with noSearch …

### LINKED ISSUE #1731: Dropdowns do not close on tab press when using noSearch
**Describe the bug**
When using `noSearch: true` on a dropdown control, tab does not close the list. Strangely, when tabbing BACK through and getting to that element, then pressing tab, it does close. I think it has to do with the active element at that point, which triggers the `handleKeyDown()` handler. When using the mouse to open the menu, only `handlePseudoElemKeydown()` fires.

I noticed in `handlePseudoElemKeydown()`, there is this code: `(key === 'Tab' && this.isOpen())`.
The issue is that it's within an if statement that requires `['ArrowDown', 'ArrowUp', 'Down', 'Up', 'Enter', 'Spacebar', ' ']`, so that code will never properly fire. Lower in the function, this code exists:

`    // Allow some keys to pass through with no changes in functionality
    if (isEscapeKey || key === 'Tab') {
      return true;
    }
`

That doesn't seem correct to me, pressing tab should always close the open menu when using noSearch I would think.

**To Reproduce**
1. Go to http://localhost:4000/components/dropdown/example-no-search-lsf.html
2. Use the mouse to open the menu, then press tab. The next dropdown is focused, but the menu doesn't close.

If you keep tabbing so that you cycle back through the page, the menu will close, which is coming from `handleKeyDown`, `case 9:`.

**Expected behavior**
When a dropdown is open, and tab is pressed, the menu should close and focus the next element available.

**Version**
- ids-enterprise: master

awbuboltz:
…option to close an open dropdown menu

When using noSearch: true on a dropdown control, tab does not close the list.

https://github.com/infor-design/enterprise/issues/1731

Opening a standard dropdown (noSearch: false) and a noSearch:true dropdown and pressing tab should close the menu and focus the next element. Escape, arrow keys, and normal searching keys should all behave the same.

### DIFF
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -1348,10 +1348,6 @@ Dropdown.prototype = {
         this.open();
       }
 
-      if (key === 'Tab' && this.isOpen()) {
-        this.closeList('tab');
-      }
-
       // TODO: refactor this out so that `handleKeyDown` is no longer necessary.
       // This is necessary here because in `noSearch` mode, there is no actionable searchInput.
       if (this.settings.noSearch && !e.ctrlKey) {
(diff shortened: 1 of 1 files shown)


tmcconechy:
Would you edit the pr template to add the mentioned details please. 
Seems like just the template was left in by mistake.
Fix look good I’ll test it out

Issue #1747, opened 2019-03-06
Demo App: Remove "font" query parameter functionality


clepore:
**Is your feature request related to a problem? Please describe.**
Since we are slowly creating the uplift theme, we do not need `?font=` url query parameter for the demo app.

**Describe the solution you'd like**
Remove the`?font=source-sans` capability from the demo app.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

EdwardCoyle:
Is the idea here that the font style will come through the tokens, per theme?

tmcconechy:
I think so. I think we can just remove the url param, but also the personalize functions to set theme.

EdwardCoyle:
@clepore @nickwynja

Are we ok with removing the query param from the app?  One concern we had was the fact that design.infor.com shows all the current samples using the new font, so we would technically be breaking the website (even though the font would just default to the standard).

For example, the embedded design here:  https://design.infor.com/code/ids-enterprise/latest/grid

tmcconechy:
Things we could remove:
1. Do we want to say that this font is part of the theme now?
1. Change the site to not show that font for this theme when embedded
1. Change the site to not send a url param for this theme when embedded
1. Remove the option on personalize.js to change the font for a theme.
1. Remove the URL Param?

Sound like the goal?

nickwynja:
We never really adequately rolled out the new font in the default "soho" theme, so I'm OK removing the font query param from the app and standing behind the new font for the uplift theme. We can also remove the query param from the embeds on the website.

I'm not sure we should touch `personalize()` though. If there's a chance that someone is depending on it there, we'd be breaking things for them.

tmcconechy:
Sounds like the best idea. Maybe we can deprecate the setting in personalize.js. It mostly works but we’ve discovered a couple line height issues that we probably don’t want to continue to support going forward for old themes. Not sure that many would be using that in prod.

EdwardCoyle:
I think I may also try to pick up #1658 in this sprint, in which case I will punt on the deprecation of the font setting in personalize.js and maybe try to work that into the larger deprecation notice framework.

nickwynja:
I've removed the font query param from the website.

Issue #1738, opened 2019-03-05
Listview: Updated() passing "undefined" for setting no longer refreshes the component


pwpatton:
**Describe the bug**
The listview.js `updated()` function was changed to only refresh the list view when setting were not undefined. This breaks EP/NG content projection where we project new elements into the `div.listview` element. It manifests itself as not displaying checkboxes for mixed and multiple selection modes.

See the change here that caused the issue.
https://github.com/infor-design/enterprise/commit/211804d13fa0ed7d2fb43bf4ab9fde846da020b9#r32531426

**Expected behavior**
Expect `update()` with undefined args to call the `refresh()` function of listview.

**Version**
4.16.0-beta.0 and 4.16.0-rc.0

tmcconechy:
I saw this comment but wasnt 100% clear on how to reproduce this. What do you think @pwpatton ?

pwpatton:
To duplicate the issue in EP I think we’d need a test sample to do these  steps 

- initialize a list view w/o a template and instead initialize from html with settings.selection set to ‘mixed’
- once displayed change the list items in the dom and call updated() w/o any args. 
Before the regression they updated() function would call the refresh() function and it would add the checkboxes to the items. 
After the regression the refresh isn’t called and the checkboxes don’t get rendered. 

This essentially mimics what we do in

pwpatton:
a work around to this issue could be to call the updated function with an empty object: ‘listview.updated({})’. 

That would cause the refresh function to be called that adds the checkboxes.

CindyMercadoReyes:
This issue is now resolved.

PR #1748, opened 2019-03-06
new sample that initializes and updates the listview from html

pwpatton:
Added a sample that mimics the way list view is used in enterprise-ng

#1747

- http://localhost:4000/components/listview/example-custom.html
- click on mulitple selection button
- click on show dataset 2
verify checkboxes are still displayed.


EdwardCoyle:
Related to the changes in #1738 - A similar sample in NG was used to diagnose that issue, so it's good to have this as an EP test as well.

PR #1763, opened 2019-03-07
1604 - Fixed filter dropdown popup overflow with datagrid

### LINKED ISSUE #1604: Datagrid filter dropdown overflows page when column is at the edge of the page and a value is long
**Describe the bug**
When the column is next to the edge of the browser, the filter dropdown's popup can overflow the page.  when there is a long value in it. This means that the scrollbar doesn't appear and some values could be cut off.

**To Reproduce**
1. Copy the attached
[example-filter.zip](https://github.com/infor-design/enterprise/files/2837803/example-filter.zip)
 to components/datagrid and navigate to it.
3. Open the dropdown filter for the status column.
4. See that the popup is overflowed off the page.

**Expected behavior**
The popup doesn't get overflowed off the page.

**Version**

- ids-enterprise: 4.16.0-dev

**Screenshots**
image

**Platform**
 - All

**Additional context**
Happening in mobile and desktop

deep7102:
When the column is next to the edge of the browser, the filter dropdown's popup was overflow the page. Where they're long value in it the scrollbar was not appear and some values cut off.

Closes #1604

http://localhost:4000/components/datagrid/test-filter-dropdown-overflows.html
- Open above link
- Make window size smaller till reach the right edge of grid
- Open the dropdown filter for the status column
- See that the dropdown should not overflowed off the page


EdwardCoyle:
I tested this one pretty thoroughly and it's a nice fix for the problem that leaves the other place logic in tact.

@deepak, I think the [visual regression test](https://travis-ci.com/infor-design/enterprise/jobs/183205316#L1344) that keeps failing may be ever-so-slightly affected by the change... It's consistently failing on each run, so we may need to update the baseline image for the test before they pass.  For the record, it looks fine to me at a glance:

screen shot 2019-03-07 at 4 59 37 pm

deep7102:
Fixed now, filter dropdown trigger icon was not aligning

tmcconechy:
Not sure if it’s related but I see your headers are screwy. Maybe it’s the tokens. Try doing ‘npm install’ and then ‘npm run start’. See if that makes a difference but it may be that’s a different “problem” caused by just there being too many columns so they won’t always auto fit.

davidcarlsonberg:
Blew away `node_modules` and reinstalled and started. Issue persists.

tmcconechy:
The issue noted by @davidcarlsonberg can be fixed by removing the fixed with `width: 170, ` in the example. It doesnt seem like thats needed to reproduce the issue, but its also not totally incorrect to use fixed width calculations but its not really a bug in my opinion. Maybe just remove it @deep7102 ?

deep7102:
Removed now, the fixed `width: 170` from example page

EdwardCoyle:
I still see the overflow if the column bleeds off:

Screen Shot 2019-03-12 at 10 26 06 AM

davidcarlsonberg:
I also see what Ed mentions.

davidcarlsonberg:
> I still see the overflow if the column bleeds off:
> 
> Screen Shot 2019-03-12 at 10 26 06 AM

I also see it.

tmcconechy:
This was fixed, im sure i saw it. Maybe removing the width causes it not to work right?

deep7102:
Updated now, with another fix for bleeds off overflow

EdwardCoyle:
I'm still seeing an edge bleed, and I found something else that may/may not be related:

- Open the test page at a breakpoint that forces column scrolling
- Scroll the datagrid so the dropdown is half-bled off the right edge
- Open the dropdown
- Click around the text/sort buttons in the Dropdown column's header.  If the dropdown closes, sometimes the column alignment gets out of whack compared to the cells.

See the gif below (happens around 0:11 seconds):

Mar-12-2019-11_58-AM

Issue #404, opened 2018-07-17
Accessibility: Failed Dropdown Accessibility Tests


clepore:
Axe is reporting the following WCAG2AA errors as critical errors

- [x] Ensures ARIA attributes are allowed for an element\'s role (remove aria-readonly)
- [x] Ensures elements with an ARIA role that require child roles contain them (doesn't like the div and would prefer using an input)
- [x] Ensures all ARIA attributes have valid values (remove aria-controls and add only when the list is actually open.
- [ ] Test and document use with Jaws/nvda

Follow debugging instructions on docs/TESTING.md to investigate the issue in the devTools console, and run test in high-contrast which has disables zero rules, These errors can also be double checked with Chrome Audit in devTools, or use the Chrome or Firefox extension, https://www.axe-core.org/

This also effects the dependent on dropdown; multiselect component as well

Formerly: SOHO-7937

tmcconechy:
Approach here will be to try to change from a combobox pattern to https://w3c.github.io/aria-practices/examples/listbox/listbox-collapsible.html

janahintal:
QA Passed. Will Moved to Done. 
Tested in: http://master-enterprise.demo.design.infor.com/components/dropdown/example-index.html using Chrome and ChromeVox.

PR #1767, opened 2019-03-08
1552 - Allow seperate locale and language

### LINKED ISSUE #402: SOHO-7947: Add the ability to add locales to the defaultLocale function in sohoxi.js
**Use Case 1**

Landmark uses some specific languages called Manglish and Maribic to test translations.
Landmark passes in "mx-LW" and "mx-LW-RTL" to Soho.Locale.

These are not included in the variables of allLocales or defaults$$1 so they always return blank.
Would like the ability to pass in some additional locales to be appended to these varaibles. 

Something like.
```
defaultLocale: function defaultLocale(locale, addLocales, addDefaults) {
}
```

**Use Case 2**

Show an example of adding custom strings to an existing locale.

```
Soho.Locale.currentLocale.data.messages = extend(true, {},
Soho.Locale.currentLocale.data.messages, newMessages);
```

### LINKED ISSUE #1552: Locale: Add support to set language and locale independently
**Background**
A customer requirement is the ability to separate locale formatting from language. Ming.le provides these values as separate parameters, and we need the IDS components to support the same. The language and locale need to be set on a high level, when initializing the IDS Locale object, and not as options when initializing specific components. 

Example: Use English translations for all IDS components, but Swedish formatting (dates, numbers, etc.)

**Problem description**
There is currently no way to set language and locale to different values when initializing the IDS Locale object, all formatting is done based on the set language. 

**Proposed solution**
Make it possible to set language and locale independently of each other. No code change should be required for Homepages widgets / content that consume the actual IDS components, only for the IDS Locale initialization.

tmcconechy:

Rejigger the locale api so that you can have a different locale and language at the same time. Corrected incorrect html lang value. Also solved #404 at the same time by exposing the arrays for supported locales so you can add your own custom one (add Latin la-IT as a test language).

Fixes #402
Fixes #1552

- everything is covered by the new tests
- peep out the new api functions
- also can check out these examples:

http://localhost:4000/components/locale/test-set-language.html?locale=ar-SA
- first set of text is in whatever initial locale is set
- second set of text is Finnish after calling the api to switch language

http://localhost:4000/components/locale/test-add-custom-locale.html?locale=ar-SA
- first set of text is in whatever initial locale is set
- second set of text is Latin after calling the api to add a "non-supported" locale

http://localhost:4000/components/locale/test-extend-messages.html
- last two text items are added by extending the api


src/components/locale/readme.md
```diff
@@ -32,6 +32,48 @@ Soho.Locale.set('es-ES').done(function () {
 
 By default it will fetch the js files on your server from the cultures folder that is assumed to be in the same localtion as where the sohoxi.js script is. However, this can be adjusted by setting `SohoConfig.culturesPath = 'myserver/path/cultures/`
 
+## Code Example - Using a different Language and Locale
+
+It is possible to use the strings in another language, independently of the locale settings for date and numbers ect. to do this just call the `setLanguage` api function. As with the `set` function if the locale with the strings is not loaded there may be a delay while the script is loaded in. So use the callback `done` function before initializing components or using the strings.
```
davidcarlsonberg:
```suggestion
It is possible to use the strings in another language, independent of the locale settings for date, numbers, etc. To do this, call the `setLanguage` api function. As with the `set` function, if the locale with the strings is not loaded there may be a delay while the script is loaded in. So use the callback `done` function before initializing components or using the strings.
```

src/components/locale/readme.md
```diff
@@ -32,6 +32,48 @@ Soho.Locale.set('es-ES').done(function () {
...
+```javascript
+Soho.Locale.set('en-US');
+Soho.Locale.setLanguage('da').done(function () => {
+  Locale.translate('Actions'); // Returns 'Handlinger'
+});
+```
+
+## Code Example - Extending a Locale/Language with your own strings
+
+It is possible to add your own strings to an existing locale's language. To do this just set the locale or language to the desired language and then when the locale is loaded call `extendTranslations` with a new set of strings. A string is an object with minimum an id and value of the string.
```
davidcarlsonberg:
```suggestion
It is possible to add your own strings to an existing locale's language. To do this set the locale or language to the desired language and then when the locale is loaded call `extendTranslations` with a new set of strings. A string is an object with, at minimum, an id and the value of the string.
```

src/components/locale/readme.md
```diff
@@ -32,6 +32,48 @@ Soho.Locale.set('es-ES').done(function () {
...
+  Locale.extendTranslations(Locale.currentLanguage.name, myStrings);
+  Locale.translate('Comments'); // Returns Commenti
+  Locale.translate('Thanks'); // Returns Grazie
+  Locale.translate('YourWelcome'); // Returns Prego
+});
+```
+
+## Code Example - Adding an entirely new Locale
+
+It is possible to add a new locale that is not supported by Infor (yet). To do this you need to make a new file like <a href="https://github.com/infor-design/enterprise/tree/master/src/components/locale/cultures" target="_blank">any of these locales</a> and place this file in your cultures folder on the server. You also need to add the new locale to the `defaultLocales` and `supportedLocales` sets. Then you can use it. If you do this consider making a Pull Request to get your new locale added to the core code for others.
```
davidcarlsonberg:
```suggestion
It is possible to add a new locale that is not yet supported by Infor. To do this you need to make a new file like <a href="https://github.com/infor-design/enterprise/tree/master/src/components/locale/cultures" target="_blank">any of these locales</a> and place this file in your cultures folder on the server. You also need to add the new locale to the `defaultLocales` and `supportedLocales` sets. Then you can use it. If you do this consider making a Pull Request to get your new locale added to the core code for others.
```

PR #1774, opened 2019-03-11
#1770 - Add absolute position on error icons in field-short dropdowns…

### LINKED ISSUE #1770: Firefox non-inline field errors shrink dropdowns
**Describe the bug**
When adding a non-inline error icon/tooltip to a dropdown in Firefox, the field shrinks and overlapping occurs.

**To Reproduce**
See this code:
https://gist.github.com/awbuboltz/3386b57f9a9d85e134db61460ae39e58

**Expected behavior**
Other browsers leave the size as-is and just shows the error icon and tooltip.

**Version**

- ids-enterprise: master

**Screenshots**
I have a size 35px dropdown here, it shrinks and overlaps to the point of not being able to see it.
image

**Platform**
 - Browser Name: Firefox

**Additional context**
https://gist.github.com/awbuboltz/3386b57f9a9d85e134db61460ae39e58

awbuboltz:
… and add a test file.

Firefox requires that the error icon is absolutely positioned or it will take up space in the parent div. I found that the error icons in non field-short fields are positioned absolutely via _icons.scss, but field-short was not taken into account there. To minimize risk to other controls or scenarios, I added the position: absolute in _dropdown.scss, within the .field-short parent and only for .icon-error.

This is a fix #1770.

View test here: app\views\components\dropdown\test-firefox-error-positioning.html
http://localhost:4000/components/dropdown/firefox-overlap-error.html

### DIFF
--- a/app/views/components/dropdown/test-firefox-error-positioning.html
+++ b/app/views/components/dropdown/test-firefox-error-positioning.html
@@ -1,0 +1,41 @@
+<div class="row">
+  <div class="six columns">
+
+    <h2>Dropdown Test: Overlapping Firefox Error Icon</h2>
+
+  </div>
+</div>
+
+<div class="row">
+  <div class="twelve columns">
+
+    <div class="field-short">
+      <label class="label" for="validated-dd">LSF Dropdown</label>
+      <select id="validated-dd" class="dropdown">
+        <option value=""></option>
+        <option value="1">1</option>
+        <option value="2">2</option>
+        <option value="3">3</option>
+        <option value="4">4</option>
+      </select>
+    </div>
+
+    <br><br><br>
+
+    <button id="showError" class="btn-primary">Show error</button>
+    <button id="removeError" class="btn-primary">Remove error</button>
+  </div>
+</div>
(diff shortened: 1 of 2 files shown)


davidcarlsonberg:
Do we need to create a whole new test page for this? Or can the behavior be referenced in an existing example page?

davidcarlsonberg:
Hi! To follow up @awbuboltz , feel free to remove the test page(s) you added and simply include this page in the description and include necessary steps (browser etc.) for reproduction/testing: http://localhost:4000/components/validation/test-legacy-tooltip-narrow-field.html

awbuboltz:
@davidcarlsonberg, that page you mentioned does not contain a dropdown, nor does it allow adding/removing errors.

Are you wanting to lessen the amount of test pages that exist? If there is a a better place to put the sample I wrote, that's fine but it's a pretty unique use case and needs it's own sample.

davidcarlsonberg:
Sorry to be unclear. What I meant was you can add your example to that page. And, yes, to reduce the amount of test pages.

PR #1782, opened 2019-03-13
1714 - Create grouped headers align setting

### LINKED ISSUE #1714: Datagrid: Allow align setting on the grouped headers
**Describe the bug**
When using grouped headers the older version (3.x) had the grouped header centered while the current version left aligns it. Would be preferable to add an align setting you can feed in.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/example-grouped-headers.html
2. Look at the "'Column Group ...' text
3. Center may be preferable as a default but should make an option either way.

**Expected behavior**
Could set it up like this and make a decision if center should be default for the top row.

```
  columnGroups: [
          { colspan: 3, id: 'group1', name: 'Column Group One', align: 'center' },
          { colspan: 1, id: 'group2', name: 'Column Group Two' }, // Aligns left or center as default
          { colspan: 2, id: 'group3', name: 'Column Group Three', align: 'right' },
          { colspan: 3, id: 'group4', name: 'Column Group four' } // Aligns left or center as default
  ]
```

**Version**
All

**Additional context**
Landmark Email

davidcarlsonberg:
Add an option to align text for grouped headers.

Closes #1714.

- Pull branch
- Run app
- Browse to http://localhost:4000/components/datagrid/example-grouped-headers.html
- Ensure grouped headers (`.datagrid-header-groups th`) align to option as dictated in example page.


davidcarlsonberg:
Is it desired that we allow control for `rtl` to align their own headers or should `left` always default to `right` in an `rtl` situation?

tmcconechy:
@davidcarlsonberg I think it should flip. So whats aligned left aligns right on RTL and vise versa.

PR #1787, opened 2019-03-13
1764 - Datagrid colspan hidden columns

### LINKED ISSUE #1764: Datagrid colspan is not skipping hidden columns
Describe the bug
For a datagrid containing a column with a colspan function, the returned colspan value is the number of visible columns that need to be spanned.  Issue occurs when a column is already hidden via hideColumn()  the 'is-hidden' class is not being added to the correct number of visible columns.  Leads to a column alignment issue. 

To Reproduce
1. Have a column with a colspan function
2. Have a hidden column within the colspan value
3. See alignment issue

Expected behavior
Columns are correctly aligned because 'is-hidden' class is added to visible columns.

Suggested Fix
Within the rowHtml function
...
if (skipColumns > 0 **&& !_col.hidden**) {
  skipColumns -= 1;
  cssClass += ' is-hidden';
} 

**Screenshots**
Without the fix screenshot:
image

With the fix screenshot:
image

davidcarlsonberg:
Hidden columns were not taken into account when calculating colspan. This was causing alignment issues.

Closes #1764.

- Pull branch
- Run app
- Browse to http://localhost:4000/components/datagrid/example-colspan.html
- Notice alignment is not off after adding logic to account for non-hidden columns.
- Can remove added conditional logic in datagrid.js to see how alignment skews incorrect for testing.


EdwardCoyle:
The test failing has to do with visual regressions related to `colspan`, so we might need to rework the baseline images for that one to get it passing:  https://travis-ci.com/infor-design/enterprise/jobs/184680534#L1570

PR #1804, opened 2019-03-15
1553 - Added support to three label sizes with radar

### LINKED ISSUE #1553: Radar: Legend crops off radar chart Is smaller display like when placed in card or mobile
If the radar chart displays it appears media tags try to handle sizing as the area gets smaller but it appears the labels are not of the solution as at times they are cut off. You can see it in the radar examples by making the screen slowly smaller and watch the repaint.

This can be reproduced in the standard enterprise examples
/components/radar/test-axis-formatter.html

radar

deep7102:
Added support to three label sizes (name, abbrName, shortName).

Closes #1553

http://localhost:4000/components/radar/test-long-text.html
- Open above link
- Resize the the browser window
- See label sizes changes according to view port


src/components/radar/radar.js
```diff
@@ -158,13 +158,27 @@ Radar.prototype = {
...
+    };
+
     let tooltipInterval;
     const colors = d3.scaleOrdinal(self.settings.colors);
 
     // If the supplied maxValue is smaller than the actual one, replace by the max in the data
     const maxValue = Math.max(settings.maxValue, d3.max(data, i => d3.max(i.data.map(o => o.value))));  //eslint-disable-line
 
-    const allAxis = data[0].data.map(i => i.name); // Map the names to the axis
+    const allAxis = data[0].data.map(d => getNameText(d)); // Map the names to the axis
```
davidcarlsonberg:
```suggestion
    const allAxes = data[0].data.map(d => getNameText(d)); // Map the names to the axes
```

src/components/radar/radar.js
```diff
@@ -158,13 +158,27 @@ Radar.prototype = {
...
     const total = allAxis.length; // The number of different axes
```
davidcarlsonberg:
```suggestion
    const total = allAxes.length; // The number of different axes
```

PR #1809, opened 2019-03-18
1658 - Improve deprecation process

### LINKED ISSUE #1658: Build process: Build a developer notification module/utility
We need a way to notify developers that things are being deprecated or going away.

1. I would like a common helper utility that we can tie into in the future to get "deprecations." Right now it can simply take some sort of object, or array of objects, and do a `console.warn()`. See [here](https://css-tricks.com/approaches-to-deprecating-code-in-javascript/) for an example of their "Obsolete" function helper.
1. Use the proper [JSDoc @deprecated](http://usejsdoc.org/tags-deprecated.html) syntax
1. Add a "Deprecation" and maybe "Pending Deprecation" sections to the changelog to list what was deprecated in that release
1. Create a `docs/` document about how deprecate a feature/method...etc

n/a

**Additional context**
Informational links:
- https://css-tricks.com/approaches-to-deprecating-code-in-javascript/
- https://www.npmjs.com/search?q=deprecations

EdwardCoyle:
This PR adds a couple utilities to IDS that implement a better deprecation process:
- A browser-based utility that will log warnings to the console if a deprecated method or property is used.
- An optional build-time plugin for Rollup that analyzes the source code for deprecated methods/properties
- Some component documentation has been modified to remove `private` on some public-facing, deprecated methods, to actually allow the build process to identify them.  In some cases, some much older, deprecated methods were completely removed.
- Documentation on our deprecation process, along with recommendations for use.

Additionally, this PR upgrades Rollup to `v1.6.0`, along with upgrades to some of its plugins.

Closes #1658

_before testing:_
- Pull this branch
- run `rm -rf node_modules && npm i` in the terminal
- Review the documentation in (/docs/CODING_STANDARDS.md#Deprecations)

_test the compile-time plugin:_
- Run `npx rollup -c --verbose` to see Rollup run with a list of deprecated API methods/properties.
- Run `npm run build`.  Ensure the entire build completes.  Rollup should not identify any deprecations in the console.
- Run `npm run build -- --verbose`.  Ensure the entire build completes.  This time, Rollup should output a list of deprecated API methods/properties, among other things.

_test the browser notifications:_
- Run the demoapp
- Open a browser to http://localhost:4000/components/blockgrid/example-paging.html
- Open a Dev Tools console.
- Click on any of the blocks in the Blockgrid.  A warning in the dev tools about using the deprecated `selectBlock()` method should appear.  This is because the Blockgrid is using the original method name instead of the new method name internally (this is noted in #1820).
- Replicate this by opening a sample page for any of the Components listed with deprecated API methods/properties in the build output.

Screen Shot 2019-03-19 at 1 01 47 PM


tmcconechy:
@EdwardCoyle Try and pull master on and see if it fixes some of these test fails.

test/helpers/e2e-utils.js
```diff
@@ -1,3 +1,5 @@
+const ACCEPTABLE_LOG_LEVELS = ['SEVERE'];
```
tmcconechy: What are the different log levels? Does this allow for straight `error`? (not just severe)

EdwardCoyle: After some reading it seems like errors are reported to the browser driver as `SEVERE`: https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities#loggingpreferences-json-object

I may actually take this out and just make Protractor set a higher log level in the settings, per that documentation.

tmcconechy: Ok that sounds good. I had it catching everything and it was working pretty well. So didnt want to loose any "errors"

EdwardCoyle: Yeah I didn't want to lose any of that either.  The problem is -- when I added `console.warn` statements in for deprecations, they started getting captured by this check in the test harness, which was grabbing all types of log statements regardless of type.  First attempt might not be the best way though.

EdwardCoyle:
> Should we show warnings in the browser for private methods?

I agree that we probably shouldn't.  That stuff's important for us, not so much for end users, so I think at the very least we document it in the source code somehow without propagating it to browser warnings.

> I noticed that the `parseOptions` example in the browser above, does not appear in the cli warnings. Should it?

That is technically `private` internal API, so I think I'll go in and change it so that message is not output.

> Should we show the "ultimatum" version along with the deprecation version?

My preference for this is to keep the console warnings (both build and browser) as short as possible, mostly for clarity.  We can use multi-line comments with paragraph breaks in the JSDoc comments in source code to document detailed things about the deprecation, without pushing them to the console.  Maybe I can push this to the docs page, but I think we should point people to look at the source for details about deprecations where they exist.

clepore:
@EdwardCoyle :banana: on that

EdwardCoyle:
@clepore one clarification:

> Prefix the console and browser warnings...

I don't think we need to do this in the terminal notifications, since it's part of the build system and is labeled as such.  I think it's a good idea for the browser.

clepore:
I figured it was a standard message that we displayed for all, but if its 2 different strings, that's fine. Not too many people will be building.

EdwardCoyle:
Just noting for re-reviewers, I had to change the browser test because we decided that specific message shouldn't be displayed in the browser window, since it's private API.

Additionally, the test that replaces it is technically demonstrating a bug, which I have promptly raised #1820 for.

PR #1821, opened 2019-03-19
462 - Allow Locale functions and some components to accept a locale

### LINKED ISSUE #462: SOHO-3124: 4.0 Locale unable to indicate locale for parseNumber() or formatNumber()
Please add the ability to indicate the locale for parseNumber() and formatNumber() and translate()

Landmark used Globalize.cultures["default"] (en-US) to parse and format numbers for the Landmark storage formats for numbers. Locale.parseNumber() and .formatNumber() uses the current culture; if the current culture is not en-US, there is not a way to parse or format a number for en-US or for any other culture for that matter. 
This issue is formerly SOHO-3124

- [x] Include option to pass a non current locale to parseNumber() or formatNumber() and translate
- [x] Include option to pass a non current locale to datepicker/timepicker/monthview and calendar
- [x] Additionally, would also like the ability to indicate to not use grouping. This was possible with 
```
Locale.formatNumber(1234567.1234, { group: '' }); // Equals 1234567,123
```

tmcconechy:

#462 has a few asks around change the locale thats used for a function or component. This PR satisfies it all. Cleanup some smaller issues i found like some missing month data and icon alignment.

Closes #462 

Ability to change the groupsize

- this is just in the tests https://github.com/infor-design/enterprise/compare/462-passing-locales?expand=1#diff-274e5585eb6805af7c63c9dcb61798feR40

Ability to pass locale to translate, parseDate, parseNumber, formatNumber and parseNumber

- this is also in the tests https://github.com/infor-design/enterprise/compare/462-passing-locales?expand=1#diff-aa0a79b8248cb1237ef6589d2049198fR1242
- can also review this working example http://localhost:4000/components/locale/test-pass-locale.html

Ability to pass locales into calendar, date, timepicker and monthview
- test the following pages - each shows a component in Danish which is not the current locale
http://localhost:4000/components/monthview/test-specific-locale.html
http://localhost:4000/components/calendar/test-specific-locale.html
http://localhost:4000/components/datepicker/test-two-locales-same-page.html
http://localhost:4000/components/timepicker/test-specific-locale.html
- also has a func test for each
- smoke test some examples on 
http://localhost:4000/components/monthview/list
http://localhost:4000/components/calendar/list
http://localhost:4000/components/datepicker/list
http://localhost:4000/components/timepicker/list


EdwardCoyle:
Ran through the test pages and didn't see anything out of sorts.  I'll :banana: upon tests passing

EdwardCoyle:
This might be something we have to update later in an example, but I just realized that if you have an input field with a pre-populated value, it doesn't get the updated locale and format itself correctly via the Mask.  You can see it on http://localhost:4000/components/datepicker/example-index?locale=nl-NL:

Screen Shot 2019-03-19 at 3 52 29 PM

This could just be a sample issue, or a case of "just populate it correctly when you create the component", but it could also be something we handle in the component?

tmcconechy:
@EdwardCoyle - since the value is just an attribute, we could make this set the values by the API. But we can do that later. Been like this a while.

PR #1830, opened 2019-03-21
1021 - Fixed rowactivated for ie11 with datagrid [v4.17.x]

### LINKED ISSUE #1021: Datagrid - need a mechanism to cancel a rowActivated event
Need a mechanism to cancel a rowActivated event.  Perhaps before a rowActivated event occurs, fire a before event where a return of false means the event should be ignored?  Or make the rowActivated event cancellable?

Landmark business case:
Have a parent list that drives a child list.  The user adds a record to the child list. The user then changes the active row on the parent list (which then changes the child list).  Before performing the rowActivated event, need to query the user as to whether the child list changes should be saved.

deep7102:
Example page for rowactivated was not working with IE11.

Closes #1021

http://localhost:4000/components/datagrid/test-before-row-activated.html
- Open above link with IE11
- Click on any row to make it activated
- See `beforerowactivated` is triggered
- Then `rowactivated` is triggered

### DIFF
--- a/app/views/components/datagrid/test-before-row-activated.html
+++ b/app/views/components/datagrid/test-before-row-activated.html
@@ -80,20 +80,20 @@
       var delay = 3000;
       toast('Begin: Before Row Activated', delay);
 
-      var response = new Promise((resolve, reject) => {
-        setTimeout(function() {
-          var shouldGoAhead = true;
-          if (shouldGoAhead) {
-            resolve();
-          } else {
-            reject();
-            message({ type: 'error', msg: 'Row Not Activated (row: '+ args.row +')' });
-          }
-          console.log('Complete: Before Row Activated');
-        }, delay);
-      });
+      var deferred = $.Deferred();
 
-      return response;
+      setTimeout(function() {
+        var shouldGoAhead = true;
+        if (shouldGoAhead) {
+          deferred.resolve();
+        } else {
+          deferred.reject();
+          message({ type: 'error', msg: 'Row Not Activated (row: '+ args.row +')' });
+        }
+        console.log('Complete: Before Row Activated');
+      }, delay);
+
+      return deferred.promise();
     });
 
   });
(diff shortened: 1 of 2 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -6218,13 +6218,13 @@ Datagrid.prototype = {
     const findInRows = (rowNodes) => {
       let found = false;
       rowNodes.toArray().forEach((row) => {
-        row.querySelectorAll('td').forEach((cell) => {
+        [].slice.call(row.querySelectorAll('td')).forEach((cell) => {
```
tmcconechy: Did you mean to change this? I didnt think this would be the problem

deep7102: Yes, it was breaking in ie11 because `document.querySelectorAll` return a `node-list` its array like object not an array so this needs to convert to array before using the `forEach`

app/views/components/datagrid/test-before-row-activated.html
```diff
@@ -80,20 +80,20 @@
...
+
+      setTimeout(function() {
+        var shouldGoAhead = true;
+        if (shouldGoAhead) {
+          deferred.resolve();
+        } else {
+          deferred.reject();
+          message({ type: 'error', msg: 'Row Not Activated (row: '+ args.row +')' });
+        }
+        console.log('Complete: Before Row Activated');
```
davidcarlsonberg: Maybe take this out as you have the toast notification?

deep7102: I would keep them in example page, so its easy to navigate the `Begin/Complete`

src/components/datagrid/datagrid.js
EdwardCoyle: @deep7102 FYI I had this problem in a handful of components before and added [`utils.getArrayFromList()`](https://github.com/infor-design/enterprise/blob/master/src/utils/utils.js#L1031-L1041) for the same purpose.  You may want to swap this line for that for consistency.

tmcconechy: That seems cleaner i find this syntax confusing too.

PR #1850, opened 2019-03-26
440 - Respect `groupSize` of defined locale in Mask components

### LINKED ISSUE #440: SOHO-6820: Locale/Mask: Support `groupSize` for thousands separator
Original issue:  http://jira.infor.com/browse/SOHO-6820

----

The Mask object needs to support the groupSize property which indicates how the thousands separator is formatted into the value.

While the default value is [ 3 ], there are two known locales supported by Sohoxi that have different values: hi-IN has a groupSize of [ 3, 2 ] and es-US has a groupSize of [ 3, 0 ].

The Mask object and Locale object need to be in sync when formatting numbers; please reference SOHO-6819 for the Locale object.

EdwardCoyle:
This PR enables the passing of different locales into the Mask, to more easily take advantage of the locale's preferences for formatting numbers.  This fixes a problem where some locales with different `groupSize` definitions for thousands separators were not being respected.

Additionally, this PR includes a change to the `ar-EG` locale's thousands separator, making it a decimal instead of the original comma.  This makes the keyboard output on Egyptian Arabic keyboards match what's expected by the Mask for proper formatting (previously on this locale, it was not possible to properly format the Mask fields with decimals)

Closes #440 

Pull this branch and run the demoapp, then:

_Test the groupSize setting:_
- Open http://localhost:4000/components/mask/test-number-mask-gauntlet.html
- Test fields that include thousands separators and decimals to ensure formatting while typing is working correctly.  Numbers should have the standard group size ([3, 0])
- Open http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=nl-NL
- Test the "Way Longer Number" fields with decimals/thousands to make sure they are working correctly (groupSize is [3,3])
- Open http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=hi-IN
- Test the "Way Longer Number" fields with decimals/thousands to make sure they are working correctly (groupSize is [3,2])

_Test the decimal change on `ar-EG`:_
- Open http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=ar-EG
- Type numbers into fields that handle decimal points, then type the decimal point.  It should correctly propagate.

Additionally, all e2e/functional tests should pass.

----
paging @fitzorama


tmcconechy:
@EdwardCoyle Maybe add a change log entry, i made a PR to set it up https://github.com/infor-design/enterprise/pull/1867/files

tmcconechy:
@EdwardCoyle Looks like one new test is failing? Does it need some work maybe?

tmcconechy:
Echoing on @clepore I see the same on both:
http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=ar-EG
and
http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=nl-NL

If you go to the "Longer Number (1000000.00)" and "Longer Number (1,000,000.00)" fields and type the opposite one is working (and the placeholders seem wrong). Maybe instead of brakets you can use just a dash or something "Longer Number eg. 1000000.00" so the brakets dont do the RTL flip.

EdwardCoyle:
I discovered the number mask gauntlet field labels weren't getting the proper number of fraction digits, and a reference to the locale string, which was causing some weirdness.  I fixed that, fixed some incorrect decimal points on the `ar-EG` sample, and localized the placeholders on all samples.

tmcconechy:
The only issue i still see is on http://localhost:4000/components/mask/test-number-mask-gauntlet.html?locale=nl-NL
- field "Way Longer Number: 100.000.000.000,00"
- if i type 1 then a whole bunch of 0s
- then im not getting any group characters. not sure if that label is wrong or something?

EdwardCoyle:
@tmcconechy I think the test is correctly detecting the label contents now.  It was definitely just a test issue, from what I can tell.

PR #1874, opened 2019-03-28
404 - Accessibility Fixes and Documentation

### LINKED ISSUE #404: Accessibility: Failed Dropdown Accessibility Tests

### LINKED ISSUE #1540: Accessibility: add information to docs about our role in achieving standards compliance
We need to set a clear expectation in the role IDS plays in helping an application achieve a certification in accessibility standards compliance, such as Section 508 or WCAG 2.0 AAA.

We need to draw a fine line between us providing a library that is "ready" for these certifications, while not over-reaching and doing things that an application team would be responsible for, such as making content accessible.

A rough list of items we would be responsible for may include:
- we add `aria` attributes automatically where appropriate
- we provide a compliant series of color schemes
- we provide audio-backups for written text where possible

A list of things an app team would be responsible for:
- Providing accessible language and content to be used within our patterns and components
- Appropriate usage of the components in a way that ensures all of our accessibility measures are correctly working

### LINKED ISSUE #1721: Accessibility: Issues on Skip Link and App Menu
**Describe the bugs**
A listing of minor bugs we might need to break out the bigger ones later.

- [x] ~~App Menu - > Suggest to use a UL and LI instead of a series of divs for the tree~~ Wont do, not a problem
- [x] ~~App Menu - > The has tag in the href is pronounced on older browsers, use a title~~ Wont do dont really know details.
- [x] ~~App Menu - > The divs incorrectly have `role="presentation"` (this is actually accordion `<div class="accordion-header hide-focus is-selected" role="presentation"> </div>`)~~ - On folow up this is correct.
- [x] App Menu - > if open then the skip link does not have a high enough z-index
- [x] ~~Dropdown- > Seems to not have labels~~ - Could not reproduce, maybe a Landmark only issue
- [x] ~~Lookup- > The modal has incorrect alertdialog role.~~ this is not true as on http://localhost:4000/components/lookup/example-index.html the role is dialog, so must be something on the landmark lookups
- [x] ~~Listview- > The ul also has `role="presentation"`  ~~ will be done on #403

**Version**
4.16.x

**Additional context**
Landmark Accessibility Call with External Contractors.

### LINKED ISSUE #1722: Accessibility: List Supported Accessibility Standards and Screen Readers and Browsers
We should list and document which browsers and standards we want to support.

- [x]  list tested screen readers and versions docs 
- [x]  list supported guidelines we are going after - aka WCAG 2.1 
- [x]  list supported VPAT versions we recommend -  aka VPAT 2.2
- [x] Document this in README.MD or on a lint to ACCESSIBILITY.MD from the main readme.

**Decision Info on Screen Readers**
https://webaim.org/projects/screenreadersurvey7/#browsercombos is a bit old but provide some context. Initial thoughts but we want to limit this as much as possible.

Jaws - Chrome (I know IE 11 is old and Edge doesn’t work very well and Jaws 2019 is not perfect yet). Maybe 2018 with Firefox or Chrome? 
Voice Over - Mobile (IOS)
NVDA - Chrome or Firefox
Window Eyes - Was used by some of our government agencies so far since its free with office, but might be secondary to Jaws
Chrome Vox/Voice Over - Do these work well enough on Mac at this point for a mac screen reader?

tmcconechy:

Improved the screen reader on dropdown and menubuttons. Also added and cleaned up some docs and adjusted tests. Further fixed some issues in the app menu that it was visible to screen readers and had issues in the sample app. Also fixed the zindex issue on the skip link.

Fixes #1722 
Fixes #1540 
Fixes #1721 (partially)
Fixes #404 
 
1722 & 1540
- Review the MD updates on the PR

1721
http://localhost:4000/components/applicationmenu/example-open-on-large.html
- tab from the url into the page
- when you hit the skip link it should now be on top

404
- instal and open chrome vox
- go to http://localhost:4000/components/dropdown/example-index.html
- tab to the select (it should announce the label and selected value and that its a button) note that in chrome vox the listbox role isnt correctly announced
- enter to open the list and you should hear that you can filter
- type up and down arrow to hear the items announced
- select one
- tab to the .. menu it will announce there is a popup now
- open that up with enter
- scroll the list

Because the dropdown has filtering and it does not select the elements immediately i did not add aria-selected and i made a label to announce the filtering is possible. This seems logical but looking for feedback.


tmcconechy:
cc @508it

Issue #1864, opened 2019-03-27
Releases: Automate the deploy to the AWS CDN in release it


tmcconechy:
**Is your feature request related to a problem? Please describe.**

On each release we have to manually run this command
https://github.com/infor-design/enterprise/blob/master/package.json#L59

In order to get the sohoxi files up in the AWS CDN (link https://cdn.hookandloop.infor.com/sohoxi/4.15.0/js/sohoxi.min.js)

Secondly the https://www.unpkg.com/ids-enterprise@4.16.2/dist/js/ CDN does not have a minimized js. Should see if that can be fixed too or decide one or the other formally.

I have the AWS keys for who takes this

**Describe the solution you'd like**
Update (releaseit?)so that it runs the command to deploy during the release cycle.

**Describe alternatives you've considered**
Do it manually and forget everytime.

clepore:
Have we confirmed that anyone still uses this?

tmcconechy:
Yes, because it came up the other day, they asked why the new version wasnt there.

PR #1875, opened 2019-03-28
1864 - Fix and update releasing [v4.17.x]

clepore:
We wanted to automate publishing to the sohoxi cdn on aws, and fix missing `*.min.js` files in `dist/` packages.

#1864 

1. `npm i`
1. Test the npm commands individually and make sure files appear where they should
    - `npm run clean && npm run build && npm run minify`

Note: Testing the cdn will be on release, unfortunately.


package.json
```diff
@@ -25,12 +25,14 @@
...
   "scripts": {
-    "start": "npx grunt clean && npm run build && npm run minify && npm run quickstart",
+    "start": "npm run clean && npm run build && npm run minify && npm run quickstart",
     "quickstart": "npx grunt demo && node app/server.js",
     "quickstart:livereload": "npx grunt demo && node app/server.js --livereload",
     "build": "node ./scripts/build",
     "build:custom": "node ./scripts/build",
     "build:icons": "node ./scripts/build-icons.js",
+    "clean": "rimraf temp && rimraf dist",
+    "clean:app": "rimraf app/dist",
```
tmcconechy: @EdwardCoyle This will never get called now? So where do we need to call it for the demo app? 
Should build call `clean` and start call both?

package.json
```diff
@@ -25,12 +25,14 @@
     "dist"
   ],
   "scripts": {
-    "start": "npx grunt clean && npm run build && npm run minify && npm run quickstart",
+    "start": "npm run clean && npm run build && npm run minify && npm run quickstart",
```
tmcconechy: Yeah i think this should be `npm run clean && npm run clean:app && npm run build && npm run minify && npm run quickstart` i guess?

package.json
EdwardCoyle: @clepore some scripts in here are calling `npx grunt demo`, which runs the clean task internally.  [We should take this out](https://github.com/infor-design/enterprise/blob/master/gruntfile.js#L82), and include the `clean:app` npm script in any scripts that also call `npx grunt demo`

EdwardCoyle: @tmcconechy to your question, no need to change that at all.  Just call the `clean:app` script anywhere the demoapp is started

package.json
clepore: We shouldn't clean the src build too, just the demo app files?

package.json
clepore: So what needs to happen then?

pukhrajbal:
@clepore @tmcconechy 4.17.1 is not available on CDN.

https://cdn.hookandloop.infor.com/sohoxi/4.17.1/js/sohoxi.min.js

Does this PR automate publishing or will it be a manual request to team for updating latest release?

clepore:
We are working on automating the CDN release, but it should happen naturally, yes. No request needed.

pukhrajbal:
Is there a defect number - so I can track it to know when it is available?

Also if it will take more than a week - pls update 4.17.1 on CDN for now.

clepore:
You can follow this issue for reports, and as of last night, `4.17.1` is available on the CDN.

PR #1877, opened 2019-03-28
421 - Fixed accessibility issues with sign-in pages

### LINKED ISSUE #421: Accessibility: Issue with Screen Reader and Sign In page
When reviewing our sign in page for accessibility using the MS Windows Screen Reader, I've noticed a problem with Chrome where it does not read the "user" field. The same occurs in the standard sign-in example with the email field.

http://usalvlhlpool1.infor.com/4.4.0-rc/components/signin/example-index

If you tab between the email and password fields in IE11, it reads out both labels. If you do the same in Chrome, it reads the password but when you tab back to the user field, it says "hidden".

An additional chrome issue is with icon buttons on the sign in that we have. In IE11 it reads out the audible label, in Chrome it just says button.

Also, in both browsers, it does not say anything about a field being required. In Chrome it reads out the "asterix". I was expecting different verbal feedback for required and not required fields.

Formerly: SOHO-7473

deep7102:
Fixed sign-in pages were not readable as it should with screen reader.

Closes #421

http://localhost:4000/components/signin/example-index.html
- Open above link
- Review with screen reader
- Everything should be read out in the screen reader fine


EdwardCoyle:
Tested with NVDA 2019.1 on Windows 10, against IE11:

- Opening the page reads out: "IDS Enterprise. Main landmark edit required. Password. Edit protected required. Server. Button Server NL. Button sign-in required forgot password?"
- Click to focus the "Email" field.  Read out is "Email star edit required? Blank."
- Click to focus the "Password" field. Read out is "Password star edit protected required. Blank."
- Tab to focus the "Server" dropdown. Read out is "Server NL Button".
- Click the dropdown to open.  Read out is "Type to filter".
- Hover one of the dropdown list items (I picked "NL".  Read out is "NL".
- Choose a list item. Readout is the same as opening the page (iterates through each field).

@deep7102 I'm not 100% sure on how correct everything is.  I think it's probably fine, with the notable exception of reading the entire page out once a Dropdown item is picked, but I'm not sure what else the readout would be.

PR #1902, opened 2019-04-03
1873 - Lookup filter conditions

### LINKED ISSUE #1873: Datagrid Filter: lookup filtertype has incorrect filter conditions
**Describe the bug**
Currently by default, a filterType of 'lookup' only has 'Equals, Does Not Equal, Is Empty, Is Not Empty'. If you specify a filterConditions of ['equals', 'contains'], only Equals will be in the drop down.
Since a lookup could be for an integer field or an alpha field, would have the default for lookup fields to have the same as text and numeric (Contains, Does Not Contain, Equals, Does Not Equal, Is Empty, Is Not Empty, Ends With, Does Not End With, Starts With Does Not Start With, Less Than, Less Than Or Equals, Greater Than and Greater Than Or Equals). Then the filterConditions could be used by the implementer to narrow down the list depending if the lookup is numeric or alpha.

**To Reproduce**
1. Copy
[test-filter-lookup-click-function.zip](https://github.com/infor-design/enterprise/files/3018447/test-filter-lookup-click-function.zip) into /components/datagrid/ in your codebase.
2. Go to './components/datagrid/test-filter-lookup-click-function.html'
2. Click on the filter conditions dropdown for Product Id
4. Only "Equals" is seen in the dropdown.

**Expected behavior**
Equals and Contains should be seen in the dropdown.

**Version**

- ids-enterprise: v4.17.0-rc.0

**Screenshots**
image

**Platform**
All

**Additional context**
I believe an if for filterType === lookup needs to be added around here: https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L1697

jmacaluso711:
The lookup filterType was not respecting the `filterConditions` and only showing `equals`. This PR adds all filter options to lookup filterType by default. Now the `filterConditions` can be implemented.

closes #1873 

- go to http://localhost:4000/components/datagrid/test-filter-lookup-options.html
- check that the `filterConditions` are rendering properly.


clepore:
@tmcconechy When you're back from vacation, can you verify this is the correct implementation?

tmcconechy:
Yes looks perfect idea is here if you pass `filterConditions: ['equals', 'contains', 'start-with']` it should only show those three. Otherwise it shows that set added. Its using the string ones by default so that should be good.

Issue #441, opened 2018-07-17
SOHO-6819: 4.0 Locale needs to support groupSize for thousands separator


clepore:
The Locale object needs to support the groupSize property which indicates where the thousands separator is placed.

The 3.5 controls were using Globalize which had the groupSize property and when formatting a number would properly format the thousand separator. Two cultures in 3.5 had a groupSize other than [ 3 ] defined: hi-IN with [ 3, 2 ] and es-US with [ 3, 0 ].

As noted below, the examples from 4.0 are not being formatted properly while the 3.5 examples are formatted properly.

Examples from 4.0:
```
Locale.currentLocale
{name: "hi-in", data: {…}}data: __proto__: Objectname: "hi-in"__proto__: Object

Locale.formatNumber(123456789.12, {style: 'decimal', round: false, minimumFractionDigits: 2, maximumFractionDigits: 2})
"123,456,789.12"
```

```
Locale.currentLocale
{name: "es-us", data: {…}}

Locale.formatNumber(123456789.12, {style: 'decimal', round: false, minimumFractionDigits: 2, maximumFractionDigits: 2})
"123,456,789.12"
```

Examples from 3.5:
```
Globalize.culture("hi-IN")
{name: "hi-IN", englishName: "Hindi (India)", nativeName: "????? (????)", isRTL: false, language: "hi", …}

Globalize.culture().numberFormat.groupSizes
(2) [3, 2]

Globalize.format(123456789.12, "n2")
"12,34,56,789.12"
```
```
Globalize.culture("es-US")
{name: "es-US", englishName: "Spanish (United States)", nativeName: "Español (Estados Unidos)", isRTL: false, language: "es", …}

Globalize.culture().numberFormat.groupSizes
(2) [3, 0]

Globalize.format(123456789.12, "n2")
"123456,789.12"
```

CindyMercadoReyes:
Adding URL to test with: Test this page manually using the three different group sizes
http://localhost:4000/components/locale/test-groupsize.html?locale=hi-IN (3, 2)
http://localhost:4000/components/locale/test-groupsize.html?locale=nl-NL (3, 3)
http://localhost:4000/components/locale/test-groupsize.html?locale=en-US (3, 0)

CindyMercadoReyes:
This issue is now resolved.

PR #1911, opened 2019-04-04
1907 - Fix for incorrect groupSize definition on US Locale [v4.17.x]

### LINKED ISSUE #1907: Locale: en-US locale has incorrect groupsize
**Describe the bug**
The locale for en-US has an incorrect groupsize of [3, 0]. It should be [3, 3]. I believe this happened due to a typo between issue #441 where it referenced es-US and the pr #1779 where it referenced en-US.

**To Reproduce**
1. Using the browser console
    ```
    var num = "123456789012.123456"
    Soho.Locale.formatNumber(num, { style: "decimal", round: false, minimumFractionDigits: 6, 
    maximumFractionDigits: 6})
    returns "123456789,012.123456"
    ```

**Expected behavior**
- `123,456,789,012.123456`

**Version**

- ids-enterprise: v4.17.0-rc.0+

**Screenshots**
image

EdwardCoyle:
This PR fixes a bug in the IDS `en-US` culture file that was causing incorrect number formatting due to an incorrect `groupSize` definition.  It was previously set to `[3,0]`, where it needed to be `[3,3]`.

Additionally, some functional tests were altered to reflect the correctly formatted values with a `[3,3] groupSize.

Closes #1907

All functional tests related to the `Locale` API, specifically the ones that test `Locale.formatNumber()` against `en-US`, should all pass.

----
paging @pwpatton @fitzorama

### DIFF
--- a/src/components/locale/cultures/en-US.js
+++ b/src/components/locale/cultures/en-US.js
@@ -53,7 +53,7 @@ Soho.Locale.addCulture('en-US', {
     minusSign: '-',
     decimal: '.',
     group: ',',
-    groupSizes: [3, 0]
+    groupSizes: [3, 3]
   },
   // Resx - Provided By Translation Team
   messages: {
--- a/test/components/locale/locale.func-spec.js
+++ b/test/components/locale/locale.func-spec.js
@@ -739,7 +739,7 @@ describe('Locale API', () => {
   it('Should format negative numbers', () => {
     Locale.set('en-US');
 
-    expect(Locale.formatNumber(-1000000, { style: 'currency' })).toEqual('-$1000,000.00');
+    expect(Locale.formatNumber(-1000000, { style: 'currency' })).toEqual('-$1,000,000.00');
 
     Locale.set('de-DE');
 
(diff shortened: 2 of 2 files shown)


nickwynja:
@EdwardCoyle Minor thing but there's an incorrect comment here as well: https://github.com/infor-design/enterprise/blob/f6e19755a81fed34b5056d799af9c323a9daf8df/test/components/locale/locale.func-spec.js#L1106

EdwardCoyle:
There are a couple of those comments in here so I'll fix the relevant ones.  FYI There are also some legit tests that modify the groupSize to `[3, 0]` and even `[0, 0]` against the US locale.

tmcconechy:
Viewing this I thought I verified this on the CLDR as correct. But would need to recheck next week

tmcconechy:
I rechecked this and the groupsize and on the original issue is both [3, 0] so i think this was correct an unneeded. See:

https://github.com/globalizejs/globalize/blob/v0.1.1/lib/cultures/globalize.cultures.js#L2515
and
https://github.com/infor-design/enterprise/issues/441#issue-342055154

PR #1916, opened 2019-04-05
1840 - Fixed trailing zeros were getting ignored with german locale

### LINKED ISSUE #1840: Trailing Zeros are being ignored when displaying thousands values in German Locale
**Description:**
Let's say we have a value 145000, which gets displayed as 145,000 in US locale ( with comma as separator, as expected ).

But in german locale, the same value should be displayed as 145.000 ( because dot is thousands separator in german locale ) , but it gets displayed as just 145 ( no dot at the end ). We tested this in 4.3.3 and in 5.1.1 as well

However, if the value doesn't end with zeros, lets say the value is 145135, in german locale it gets displayed as expected 145.135 ( where dot is thousands separator as per german locale )

**To Reproduce**
1. Create a datagrid with one of the columns displaying numbers - lets say Column "Quantity" - Assign the values greater than 100,000. Let's say 145000
2. Have a custom formatter that displays values in german locale on column "Quantity"
3. Now, you should be seeing just 145 in the column, but not 145.000 as dot is the thousands separator in german locale

**Expected behavior**
In german locale, we expect that, whenever a value is >= 100,000 , trailing zeros should not be ignored, instead those need to be displayed with thousands separator

**Version**
Tested in 4.3.3 as well as 5.1.1 , so assuming this is issue since 4.3.3

**Screenshots**
german-numbers

**Platform**
 - Desktop
 - Windows 10
 - Chrome, IE, Edge, Firefox

**Additional context**: None

deep7102:
Fixed trailing zeros were getting ignored when displaying thousands values with German Locale.

Closes #1840

http://localhost:4000/components/datagrid/test-custom-number-formats.html?locale=de-DE
- All functional tests related to the Locale API should pass
- Open above link
- Review last two columns (Integer, Decimal)
- See 1st row should be (Integer:145.000, Decimal:145.000,00)
- And 2nd row should be (Integer:283.423, Decimal:283.423,00)

### DIFF
--- a/app/views/components/datagrid/test-custom-number-formats.html
+++ b/app/views/components/datagrid/test-custom-number-formats.html
@@ -17,6 +17,8 @@
       columns.push({ id: 'price1', name: 'Price', field: 'price', formatter: Formatters.Decimal, numberFormat: {minimumFractionDigits: 3, maximumFractionDigits: 3}, editor: Editors.Input, mask: '####,00'});
       columns.push({ id: 'price2', name: 'Price ($)', field: 'price', formatter: Formatters.Decimal, numberFormat: {style: 'currency'}});
       columns.push({ id: 'quantity', name: 'Quantity', field: 'quantity', formatter: Formatters.Decimal, numberFormat: {style: 'percent'}});
+      columns.push({ id: 'quantity', name: 'Integer', field: 'quantity', formatter: Formatters.Integer});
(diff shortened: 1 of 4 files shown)


janahintal:
QA Passed. Moving to Done. 
Tested in: http://localhost:4000/components/datagrid/test-custom-number-formats.html?locale=de-DE in all OS Browsers & Devices.

PR #1938, opened 2019-04-09
1934 - Fixed focus was changed after enable/disable tabs

### LINKED ISSUE #1934: Tabs: Calling disableTab() steals focus to next available enabled tab
**Describe the bug**
We have some hotkey logic that can call disableTab(), on keydown. When disabling, the focus jumps to the tab header of the next active tab. User can no longer type after this and has to manually set focus back to the field.

**To Reproduce**
This file should go in app/views/components/tabs.

1. Place this file in app/views/components/tabs [example-enable-disable-individual-tabs-with-keydown.zip](https://github.com/infor-design/enterprise/files/3060042/example-enable-disable-individual-tabs-with-keydown.zip)
1. Go to http://localhost:4000/components/tabs/example-enable-disable-individual-tabs-with-keydown.html
1. Click on 'Disable "Form Elements" tab on keydown'
1. Press a key, like down arrow
1. Focus will jump to 'Another' tab

**Expected behavior**
Tab should disable, but focus should stay in the field, as happens when enabling a tab.

**Version**

- ids-enterprise: master

**Additional context**
Sample file attached

deep7102:
Fixed focus was changed after enable/disable Tabs.

Closes #1934

http://localhost:4000/components/tabs/example-enable-disable-individual-tabs-with-keydown.html
- Open above link
- Click on input field (Disable "Form Elements" tab on keydown)
- Press down arrow key
- Tab "Form Elements" should disable
- Focus should stay in the input field

### DIFF
--- a/app/views/components/tabs/example-enable-disable-individual-tabs-with-keydown.html
+++ b/app/views/components/tabs/example-enable-disable-individual-tabs-with-keydown.html
@@ -1,0 +1,103 @@
+<div class="row">
+  <div class="six columns">
+
+    <h2>Tabs Test: Disable/Enable Individual Tabs Focus Steal Test</h2>
+
+    <p>The inputs below this paragraph programmatically enable/disable the "form elements" Tab. When a Tab becomes
+      disabled, its corresponding tab panel is checked for any/all IDS Enterprise controls, which are also
+      programmatically enabled/disabled.</p>
+
+    <p>This functionality is different from the <a class="hyperlink" href="./programmatic-disable-enable">Enabling/Disabling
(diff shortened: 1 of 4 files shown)


EdwardCoyle:
Per our Slack chat, I wasn't seeing this as an issue.  I think it's that on the main tabs example, the tabset is the last chunk in the page that has focusable items inside it, so pressing the "tab" key one more time escapes the browser and into the OS.  If you press Enter, then use the arrow keys instead, it seems focus is still present in the tabset:

Apr-11-2019-10_07-AM

PR #1939, opened 2019-04-09
1814 - Fix ellipsis on header for IE11

### LINKED ISSUE #1814: Cropped Header for Builder Panel When Text is Long
**Describe the bug**
In the Builder Panel Header, when the text is very long, it is being cropped. This becomes an issue for us when app is viewed on mobile because space becomes limited and fewer characters can be shown. 

Minor issue on our end, mostly occurs on mobile devices. Can this be done like the headers of other pages wherein an ellipses will be shown when the text is too long?

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/patterns/builder-wizard.html'
2. Change the Builder Panel title to a longer text. Make the browser smaller or when viewed in Mobile. The title is cropped.
image

**Expected behavior**
Can this be shown like in the Header of Pages wherein an ellipses is shown.

image

### LINKED ISSUE #1822: DataGrid: tabbing through a new row moves focus to next line for a lookup column
Describe the bug
A datagrid with actionableMode: true and cellNavigation: true is setting focus to the next cell in the next row when tabbing from a lookup column where the user has typed in a value.

To Reproduce
1. Using the attached test-addrow-tabbing.html
2. Click on the 'Add Row' button
3. Use the tab key to tab across to the Product Id column
4. Enter a Product Id value of 1
5. Hit the tab key

Expected behavior
When tabbing from the Product Id column, the focus should move to the Price column of the existing row (the newly added row)

Actual behavior
When tabbing from the Product Id column, the focus is moving to the Price column of the next row. Jumping out of the newly added row.

marknolaninfor:
Closes #1814 

1. Open  http://localhost:4000/patterns/builder-subtitle-test.html on IE 11
2. ensure the the sub-header is not cropped and has an ellipsis

### DIFF
--- a/app/views/patterns/builder-subtitle-test.html
+++ b/app/views/patterns/builder-subtitle-test.html
@@ -46,7 +46,7 @@
           <div class="title">
             <h2>
               <span class="panel-title">Builder Panel Title - Test a really really really really really really really really long title</span><br>
-              <span class="panel-subhead">Builder Panel Subheader - Test also a really really really really really really really really long subtitle</span>
+              <div class="panel-subhead">Builder Panel Subheader - Test also a really really really really really really really really long subtitle</div>
             </h2>
           </div>
           <div class="buttonset">
--- a/src/components/datagrid/datagrid.editors.js
+++ b/src/components/datagrid/datagrid.editors.js
@@ -794,7 +794,6 @@ const editors = {
       self.input.on('change', () => {
         setTimeout(() => {
           container.parent().focus();
-          grid.setNextActiveCell(event);
           grid.quickEditMode = false;
         }, 1);
       });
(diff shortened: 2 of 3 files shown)


marknolaninfor:
Note: 
Change for #1814 has also been checked in as I believe that this pull request was accepted.

"upon testing in http://localhost:4000/patterns/builder-subtitle-test.html using Windows 10 IE11, the sub-header is still cropped and doesn't have ellipsis."

tmcconechy:
Updated the PR to match the change, i think thats what we meant, here.

janahintal:
QA Passed. Moving to Done. Thanks!
Tested in: http://master-enterprise.demo.design.infor.com/patterns/builder-subtitle-test.html

Issue #1614, opened 2019-02-11
Datagrid: Add option to disable rows from selection depending on item property


guoliang:
**Is your feature request related to a problem? Please describe.**
Would like to be able to disable rows from being selected.

```
{
	id: "selectionCheckbox",
	sortable: false,
	resizable: false,
	formatter: Soho.Formatters.SelectionCheckbox,
	align: "center"
}
```

**Describe the solution you'd like**
* Use a function to determine if a row is disabled (true or false on metadata)
* dim the entire row with some opacity
* block selection / activation with both the row.

**Describe alternatives you've considered**
- using the isReadonly column function is disabled for selection checkboxes

**Additional context**
- similar to https://github.com/infor-design/enterprise/issues/1590

tmcconechy:
You can conditionally make a cell readonly with the isReadonly function. But im not 100% certain if this will work on selection checkbox because depending on your mode you may still be able to click the row to select it. https://github.com/infor-design/enterprise/blob/master/app/views/components/datagrid/test-readonly-links.html#L42-L50

We would have to format this issue as an enhancement i think to disable rows from selection.

guoliang:
was isReadonly setting property introduced in a later version? I don't seems to get it to work on 4.9.1

guoliang:
Does not work with the checkbox. I found this [issue](https://github.com/infor-design/enterprise/issues/1590). It sounds what I am looking for

tmcconechy:
I can see it is disabled for selection checkbox:
https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L3174

Soo, I think it wont work on that column as you can also select the row. So we probably need a disabled rows feature. Can you reformat the ticket with the Feature request template? https://github.com/infor-design/enterprise/issues/new?template=feature_request.md

tmcconechy:
Yeah sort of like #1590 but i think this was descoped at one point. The idea we had for the feature was "disabled rows".

- Use a function to determine if a row is disabled (true or false on metadata)
- dim the entire row
- block selection / activation.

That ticket is just about the button format, but similar.

tmcconechy:
@ericangeles This is what i suggest as a starting point

1. add a disabled() function on the column that you can assign a function to like the isReadonly one
1. in the example add some phoney logic to disable a few rows
1. add a class is-disabled from the row
2. dim the row with disabled opacity and the checkbox too
3. if you try to select a row and its is-disabled return out.

[...]

PR #1945, opened 2019-04-11
1614 - Add disable row option to datagrid

ericangeles:

Adding a feature that will disable row/s in Datagrid.

Fixed some possible bugs/issues:
- It shouldn't be checked/selected when toggle the Select All checkbox.
- It shouldn't count the disabled row/s when toggle the Select All checkbox.
- Disabled hover state.

Closes https://github.com/infor-design/enterprise/issues/1614

- Pull this branch
- Run http://localhost:4000/components/datagrid/example-disabled-rows.html


tmcconechy:
@ericangeles When do you plan on finishing this up? We freeze EOD friday. Thanks!

ericangeles:
@tmcconechy @deep7102 , disable row function is now working. Will just fix some of the issues that @deep7102 mentioned.

tmcconechy:
@deep7102 try another test i think i fixed all those

EdwardCoyle:
Just one thing I saw - the description on this PR says that the hover state shouldn't be displayed on disabled rows, but I'm seeing it anyway:

Apr-30-2019-9_18-AM

Also, should it be possible to focus on cells that are inside disabled rows?  My hunch is that it shouldn't, but I'm not sure what the rules for accessibility on a datagrid would be here.

tmcconechy:
I guess disabled is disabled so no hover or ability to click the row to focus. Not sure if we want to get into changing the keyboard to jump it though, seems like a lot of work.

tmcconechy:
@EdwardCoyle i fixed the hover state but elected to allow the rows to be focused for the time being for keyboard

Issue #1948, opened 2019-04-11
Soho.Formatters.Checkbox does not validate '1', 1 correctly


guoliang:
**Describe the bug**
Soho.Formatters.Checkbox does not validate '1', 1 correctly

**Expected behavior**
Checkbox does not get checked even though value is 1, or '1'

**Version**
4.9.0, 4.18.0-dev

guoliang:
Resolved #1949

PR #1949, opened 2019-04-11
1948 - Fixes datagrid checkboxes when string  '1' and '0 is used as the source

guoliang:
Fixes the check on 1, '1'

https://github.com/infor-design/enterprise/issues/1948

Add a data grid component, with some data.
Add a column which use Soho.Formatters.Checkbox

Column should be checked if data contains either true, 1, or '1'

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "marked": "^0.3.17",
     "mmm": "^0.2.2",
     "moment": "^2.19.2",
-    "node-sass": "^4.9.2",
+    "node-sass": "4.11.0",
     "protractor": "^5.4.1",
     "protractor-image-comparison": "^2.0.1",
     "pygmentize-bundled": "^2.3.0",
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -30,6 +30,9 @@ module.exports = {
     'protractor': true
   },
   'rules': {
+
+    'linebreak-style': 0,
```
tmcconechy: Id rather not include this. I think your trying to get around the Windows crlf you can fix that like this:

To fix up the crlf line feeds in your soho components.
 
git config core.autocrlf false
git rm --cached -r .
git reset --hard

guoliang: fixed

package.json
```diff
@@ -126,7 +126,7 @@
     "marked": "^0.3.17",
     "mmm": "^0.2.2",
     "moment": "^2.19.2",
-    "node-sass": "^4.9.2",
+    "node-sass": "4.11.0",
```
clepore: Do we need to worry about testing anything special with this `node-sass` bump? I know it's updates have given us headaches in the past?

guoliang: Perhaps not an important bump. I had to do this bump though since node-sass did not play well with the NodeJS version I have installed.

tmcconechy: I also tested this out and it seemed fine to me to update this.

clepore: Okay cool. Just checking 👍

PR #1967, opened 2019-04-16
964 - Message component accessibility

### LINKED ISSUE #964: Accessibility: Pass checks and contrast for Message component.
We need to check a handful of items related to accessibility on the Message Component.  Use the [HTML Code Sniffer](https://squizlabs.github.io/HTML_CodeSniffer/) tool or a similar one for detecting WCAG 2.0AA violations.  Make sure you have access to Windows environment with JAWS to test the announcer.

**Text color contrast issues**
NOTE: Available after #703 is merged.  Many of the examples on http://localhost:4000/components/message/example-index have contrast issues due to the lighter text color.  We should probably use a darker matching color, or black on the titles.  We may need to use a darker shade of blue on the primary buttons:

screen shot 2018-10-09 at 5 13 26 pm
screen shot 2018-10-09 at 5 13 30 pm

**Announcer Issues**
Double check how the announcer issues error messages.  It should properly announce that it's an error, as opposed to a regular message (@tmcconechy feel free to add notes about this one here).

To fix this look at the attributes... 
aria-labelledby
role

I think this should correctly announce that its an "alert", all of these are alerts of sorts. But it is only going to read whats in aria-labelled by. So "Application Error". So we might just want to check if the message type is in the string already, if its not add it in an audible label. For example if the message is "This record already exists", then add an audible label so it reads the type "Info: This record already exists" or "Error: This record already exists". Other than that i think its not too bad at the moment.

**Keyboard Control**
Ensure that while the message window is open, it's not possible to use the tab key to navigate to any items outside the scope of the Modal area (items beneath the overlay shouldn't be focusable without closing the message window).  Also check this in IE.  This isn't necessarily an issue now, but it's happened in the past and is worth re-checking / writing a e2e test for.

**Platform**
all

davidcarlsonberg:
Fixed a couple colors to pass contrast requirements for accessibility on the message component.

Closes #964.

- Pull branch
- Run app
- Browse to http://localhost:4000/components/message/example-index
- Click buttons to open
  - after clicking each button, open HTML_CodeSniffer to ensure there are no errors
- Verify in a Windows environment with JAWS that the screenreader reads "Alert" on the message title when a button is clicked and a message modal is opened.


davidcarlsonberg:
ToDo:
[ ] Add e2e test for tabbing through buttons in modal, to ensure there is no potential to tab outside of the modal.
[ ] Collaborate on JAWS instructions to create a situation where "Alert" is read by screen reader. Right now, I am only able to get the screen reader to read the message-title and message-text, not `aria-labelledby`

clepore:
Did you do you two to-do's in a previous comment?

src/components/modal/_modal.scss
```diff
@@ -198,6 +198,10 @@
   .busy.overlay {
     background-color: $modal-bg-color;
   }
+
+  .hyperlink {
+    color: $theme-color-palette-azure-70;
+  }
```
clepore: It'd probably be better to use one of the link tokens than a direct color here, see ([link tokens](https://github.com/infor-design/design-system/blob/d5f72ac3bb4fcabc9ab8b690a8f7427df80b574c/design-tokens/props/link.json).

src/components/modal/_modal.scss
```diff
@@ -441,7 +439,7 @@
   }
 
   .btn-modal-primary {
-    color: $modal-btn-primary-color;
+    color: $theme-color-palette-azure-70;
```
clepore: While we don't have modal tokens yet, we do have brand colors [here](https://github.com/infor-design/design-system/blob/d5f72ac3bb4fcabc9ab8b690a8f7427df80b574c/design-tokens/theme-soho/theme.json#L11-L23). 

The goal is to try to use less general color palette tokens as we can, and more context-based tokens. It may not work in this case, but worth checking into.

PR #1977, opened 2019-04-18
1922 - Fixed validation is-alert type was not working

### LINKED ISSUE #1922: Validation - isAlert is no longer supported - regression
**Describe the bug**
re: SOHO-7224: 4.0 Validation Alert - ability to not set border, font, control icon color

This functionality was added in February 2018 and has subsequently been lost at some point. Please refer to the jira ticket above for the requirements. When the isAlert property is set to true, the border color, control text color, control icon color is displaying the color for the alert rather than displaying the default font color. 

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/validation/test-message-types.html
2. Click on the Add Inline Message button
3. The text and date fields display an error icon with the message 'Message type is error'; the control border, control text, and control icon is red
4. Click on the Remove Message button
5. The error is removed
6. Click on the Is Alert checkbox
7. Click on the Add Inline Message button
8. The text and date fields display an error icon with the message 'Message is alert; type is error'
9. This is the error > the control border, control text, and control icon is red

**Expected behavior**
When Is Alert is checked, the control border, control text, and control icon should not have the color of the alert; it should maintain the default font color. This should be true for all of the validation types: error, alert, success, info, and icon.

**Version**

- ids-enterprise: v4.17.1

**Screenshots**
image

**Platform**

**Additional context**

deep7102:
Fixed for when the isAlert property set to true, the border color, control text color, control icon color was displaying the color for the alert rather than displaying the default color.

Closes #1922

http://localhost:4000/components/validation/test-message-types.html
- Open above link
- Click on the "Add Inline Message" button
- The text and date fields should display error icon with the message "Message type is error"
- The text and date fields border, text and calendar icon should be red color
- Click on the "Clear All Messages" button
- The error should be removed
- Click on the "Is Alert" checkbox
- Click on the "Add Inline Message" button
- The text and date fields should display error icon with the message "Message is alert, type is error"
- The text and date fields border, text and calendar icon should be default color


src/components/validation/validator.js
```diff
@@ -18,12 +18,14 @@ const COMPONENT_NAME = 'Validator';
  * @property {string} message
  * @property {string} type
  * @property {boolean} showTooltip
+ * @property {boolean} isAlert
```
EdwardCoyle: @clepore @tmcconechy @davidcarlsonberg @jmacaluso711 
Per our Slack chat, do we want to come up with a more accurate name for this?  Some of the ones that were floated were:
```
isHelpMessage
isVanillaAlert
isStandardAlert
```
Since it looks like we're adding it fresh, we should name it well now before we have to deprecate it for something later

davidcarlsonberg: `isInfo`?

deep7102: there was another from slack `isPlainAlert`

davidcarlsonberg: `isBasicAlert`? Personally I say we stay away from the word 'Alert', so not sure why I'm suggesting this.

deep7102: isHint
isNotice
isNotification

tmcconechy: I think we should avoid alert, error, notification as those are other things.

fitzorama: isFieldMessage?

EdwardCoyle: @fitzorama that sounds good to me.  Does Landmark identify this type of alert with that terminology?

clepore: My votes would be `isNotice` or `isHelpText`

EdwardCoyle: `isHelpText` also sounds reasonable to me

clepore:
Tests finally passed, took 7 "restarts". Most of the issues were with a `toData()` missing exception, some listview tests, and some flextoolbar tests.

src/components/validation/validator.js
fitzorama: @EdwardCoyle Internally in Landmark, these are known as field alerts.

fitzorama: btw, renaming this property breaks the soho-alert component in ids-enterprise-ng.

clepore: We should probably do a PR to address the change. Thank you!

clepore: @fitzorama We only changed the variable name internally, so it shouldn't...?

fitzorama: @clepore no, you're right; my bad

Issue #1970, opened 2019-04-17
Uplift: Architecture for component CSS changes for themes


nickwynja:
**Is your feature request related to a problem? Please describe.**
With more changes coming with uplifted component styles, we need to think through the architecture of having separate CSS files for uplifted theme styles. It's currently possible to have styles written directly in `uplift-theme.scss` but it could get out of hand with extensive overrides and have the file get unorganized.

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
TBD

clepore:
Here is how we are planning to handle theme uplift-specific component styles.

1. Using the Icons component as an example...
    ```sh
    src/components/icons/
      - accordion.scss         # <- Mostly all styles here
      - accordion-uplift.scss  # <- Fine tuning for uplift only (only created if needed)
    ```
1. The `src/theme/uplift-theme.scss` will include those uplift files as overrides as follows:

    ```css
    @import '../core/controls'; /* contains the normal component sass files */

    @import '../components/icons/icons-uplift';  /* (again only exists if its needed) */
    ```

EdwardCoyle:
Documenting the part needed to make this work with the IDS Builder:

- [ ] A stripped down version of `src/themes/controls_uplift.scss` will need to be "generated" to the temp folder if a custom list of components is requested.  This would be handled the same way that `src/core/_controls.scss` is currently built.  In `scripts/build.js`, [there's a code path](https://github.com/infor-design/enterprise/blob/master/scripts/build.js#L698-L710) for rendering this file with a subset of the components, sorting them in a specific order based on complexity (simpler components first, then the heavier stuff).  We would need a similar codepath for dealing with the contents of the Uplift theme.

- [ ] Just below that, there's a code path that deals with [generating the CSS index file](https://github.com/infor-design/enterprise/blob/master/scripts/build.js#L713-L727).  When doing custom builds, this is necessary because all the generated index files run out of the `/temp` folder as not to conflict with the standard index files. We'd need to account for the new Uplift component index here, and do a find/replace on its path to make sure it's correct.  My guess is we'd optionally detect the uplift theme here and only do this if necessary?

- [ ] We'd also need to add an entry for the new uplift component file [under this line](https://github.com/infor-design/enterprise/blob/master/scripts/build.js#L103)

Doing those three things should make the IDS builder compatible with Uplift.

PR #2007, opened 2019-04-24
1970 - Add uplift-specific sass files for components

clepore:
Adds a way to implement files specific to uplift only.

- [x] Verify with @EdwardCoyle that the custom builder doesn't need tweaked to handle these new files.
    - Update: See https://github.com/infor-design/enterprise/issues/1970#issuecomment-487588088

#1970 

1. `npm run start`
    - verify the styles that were originally in `src/themes/uplift-theme.scss` are still applied in the app (see the diff)

---

🚫 An e2e or functional test for the bug or feature.
🚫 A note to the change log.


clepore:
Yep, it found an example file ref in the directory that I forgot to remove.

EdwardCoyle:
I left a [comment about this](https://github.com/infor-design/enterprise/issues/1970#issuecomment-487588088) detailing the changes needed for the build.  I'd be happy to help get that up and running.

clepore:
After looking over it, you could probably do it faster than I could, so be my guest, @EdwardCoyle. There may or may not be an -`uplift.scss` for each component. I think that's the only stipulation.

EdwardCoyle:
**Noting:** As of my two commits, running `npm run build` is properly including uplift-specific CSS files in `dist/css/uplift-theme.css`.  Since we know there is a `_header-uplift.scss` source code file, I tested by doing the following:

- run `npm run build` to kick off a standard build
- open `dist/css/uplift-theme.css` after the build completes
- search for `.header h1 {` in the file.  There are 5 entires, but the very last one sits on its own just above the CSS for the hierarchy component, and is coming from the `_header-uplift.scss` source code.
- running the same search on the other three themes should result in only 4 entries, and the one above should not be present in any of those.

Screen Shot 2019-05-03 at 12 21 12 PM

**Another note:** this should work when running a custom build as well, but there are some known issues with some component combinations that will throw errors when trying to build CSS with the custom builder.  Until we fix all open custom builder tickets, we can't expect that it's working 100%.

clepore:
@EdwardCoyle If that is as good we can get it until now, and it doesn't break anything we possibly had before, my vote is to :shipit:

scripts/build.js
```diff
@@ -505,6 +507,8 @@ function sortFilesIntoBuckets(files, srcFilePath) {
   const startOfMid = getFurthestIndexOf(componentsJSFile, searchTerms.mid);
   const startOfComplex = getFurthestIndexOf(componentsJSFile, searchTerms.complex);
 
+  debugger;
```
tmcconechy: Can be removed

scripts/build.js
```diff
@@ -949,6 +966,8 @@ if (!commandLineArgs.components) {
   requestedComponents = commandLineArgs.components.split(',');
 }
 
+debugger;
```
tmcconechy: Remove?

EdwardCoyle:
@deep7102 @jmacaluso711 @ericangeles @davidcarlsonberg I'm purposely not reviewing this one since I did a bunch of work for it... aside from Tim, up to you guys to ✅ it over.

PR #2055, opened 2019-04-30
1941 - Fixed sanitize console.log was not working with datagrid

### LINKED ISSUE #1941: Datagrid: Using console.log in field data breaks datagrid layout
**Describe the bug**
If you use _console.log_ as a field in a row in a datagrid, the grid will break. This leads to unusual behavior, such as some entries not showing, cells disappearing and the content of cells not showing. I looked through the code for the HTML sanitation, and found that _console.log_ is sanitized when it's used, but found that the regex might be incorrect for detecting it, as it seems that is sanitizing too much. Note that _<script></script>_ is sanitized correctly, which also happens in the same sanitation method. 

**To Reproduce**
1. Edit a datagrid sample to contain _console.log_ in one of the fields.

**Expected behavior**
_console.log_ is either escaped or sanitized.

**Actual result**
See screenshot

**Version**

- ids-enterprise: 4.17

**Screenshots**
image

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 73.0.3683.86 (Official Build) (64-bit)

**Additional context**
Where sanitation is performed: https://github.com/infor-design/enterprise/blob/26d314b58d3444b9fe985792f43867faf7c65737/src/utils/xss.js#L62

deep7102:
If use console.log with in field data datagrid, the grid was not render properly. This PR will fix this xss issue.

Closes #1941

http://localhost:4000/components/datagrid/example-editable.html
- Open above link
- Grid should render properly
- See 2nd row in column "Product Name"
- This cell should be blank


test/components/utils/xss.func-spec.js
```diff
@@ -118,21 +118,29 @@ describe('Xss Utils', () => {
...
 
-    result = xssUtils.sanitizeHTML('<script>alert(\'hello world\')</script>');
-
-    expect(result).toEqual('');
+  it('Should santize console methods', () => {
+    expect(xssUtils.sanitizeHTML('console.log')).toEqual('');
+    expect(xssUtils.sanitizeHTML('console.log(\'hello world\')')).toEqual('');
+    expect(xssUtils.sanitizeHTML('console.log(\'hello world\');')).toEqual('');
+    expect(xssUtils.sanitizeHTML('console.log("hello world");')).toEqual('');
+    expect(xssUtils.sanitizeHTML('console.log   (object);')).toEqual('');
```
davidcarlsonberg: Are there extra spaces here on purpose?

PR #2078, opened 2019-05-02
2077 - Adding row to bottom with paging does not display the correct page for the new record

### LINKED ISSUE #2077: Datagrid: Adding row to bottom with paging does not display the correct page for the new record
**Describe the bug**
The pagerRefresh function uses the this.pagerAPI._pageCount over the correct value found in this.pagerAPI.state.pages. This can cause the incorrect page to be displayed when the new record is added to the next page.

**To Reproduce**
This will require a test page to be created 

**Expected behavior**
The function works so that the correct page is displayed when a record is added.

**Version**
Master

**Screenshots**
state if the pager when the function was running.
image

marknolaninfor:
Add a row correctly to the bottom so that the pager is updated and new record is displayed.

Closes #2077 

1. http://localhost:4000/components/datagrid/test-addrow-tabbing.html
2. click on the Add Row (bottom) button, until more that 10 rows are within the grid


tmcconechy:
Just have a lint error https://travis-ci.com/infor-design/enterprise/jobs/197076329 otherwise looks good. I like this fix, its a great idea to do it this way.

PR #2104, opened 2019-05-06
1909 - Added support for custom tooltip callback with column chart

### LINKED ISSUE #1909: Column Chart:  Include existing tooltip content in the tooltip callback setting
Need to add data to custom tooltip callback with column chart.

Currently it can not really modify existing contents, but we can overwrite.
Need to modify the custom tooltip callback with current data
```
$('#column-bar-example').chart({
      type: 'column',
      dataset: dataset,
      tooltip: function (req, res) {}
    });
```

deep7102:
Added support to existing custom tooltip content in the callback setting.

Closes #1909

http://localhost:4000/components/column/test-custom-tooltips.html
- Open above link
- Hover 1st column and see the tooltip (three lines info, name, value)
- Hover 2nd column and see the tooltip (three lines name, short name, value)
- Hover 3rd column should not show tooltip
- Any other columns should show tooltip	with two lines (name, value)

### DIFF
--- a/app/views/components/column/test-custom-tooltips.html
+++ b/app/views/components/column/test-custom-tooltips.html
@@ -1,0 +1,91 @@
+
+<div class="row">
+  <div class="two-thirds column">
+    <div class="widget">
+      <div class="widget-header">
+        <h2 class="widget-title">Column Chart Example</h2>
+      </div>
+      <div class="widget-content">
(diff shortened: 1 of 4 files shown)


janahintal:
QA Passed. Will Move to Done. Thanks
Verified in All OS, Browser and Devices : http://master-enterprise.demo.design.infor.com/components/column/test-custom-tooltips.html

PR #2110, opened 2019-05-07
391 - Fixed button states in toolbar for editor

### LINKED ISSUE #391: SOHO-8072: [IE11, Chrome] - Editor controls can be all triggered at same time
**More Important Bug**
- Type a line line “this is a header”
-  select and make it H4
-  select and make it H3 (both H4 and H3 show it cant be both)

Markup is showing it as just an H3, so H4 shouldnt show.

**Additional info**
When using the SoHo Editor the user is able to trigger all controls at the same time on one selection.

1. The problem only persists when a word (or characters) is selected and the controls are pressed one after another.
    all controls triggered
1. If the user clicks away on a different word, the currently applied styling is properly shown in the controls toolbar.
    word selected - no problem

Essentially, it would be ideal if the user has selected `H2` and decided to change it to `H3`, `H2` would get unselected. This applies to other controls that have no logical way of being applied together on one selection, i.e., can't have indent left and indent right at same time.

deep7102:
Fixed button state for toolbar buttons when clicked one after another.

Closes #391

http://localhost:4000/components/editor/example-index.html
- Open above link
- Type a line "this is a header"
- Select and make it H4
- See toolbar button H4 should be blue color
- Select and make it H3
- See toolbar button H4 should be default color
- Only H3 button should be blue color at this point

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,6 +15,7 @@
 - `[Listview]` Improved accessibility when configured as selectable (all types), as well as re-enabled accessibility e2e tests. ([#403](https://github.com/infor-design/enterprise/issues/403))
 - `[Datagrid]` Fixed charts in columns not resizing correctly to short row height. ([#1930](https://github.com/infor-design/enterprise/issues/1930))
 - `[Datagrid]` Fixed an issue for xss where console.log was not sanitizing and make grid to not render. ([#1941](https://github.com/infor-design/enterprise/issues/1941))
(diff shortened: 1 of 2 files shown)


tmcconechy:
Looking at this again i think we should just leave this as is @davidcarlsonberg because thats the why it works. For example on an new line if you type: "This is a header" and select the header text then the whole line becomes a header. 

So i think its fine , it has to make the whole "line" into the header.

PR #2112, opened 2019-05-07
2062 - Changes the scroll area on the app menu when using a footer

### LINKED ISSUE #2062: Personalization: When scrolling app menu with new color scheme bottom is left the wrong color
**Describe the bug**
application-menu element is not set to the theme personalization color

**To Reproduce**
1. Create full scroll example
2. Scroll down to see non colored part

**Version**

- ids-enterprise: 4.19 dev

**Screenshots**
image

**Platform**
 - Windows
 - 10
 - chrome
 - Version 74.0.3729.108 (Official Build) (64-bit)

tmcconechy:
Changes the css so that the outside part of the app menu scrolls and the footer is fixed when using a footer. Added an extra color in the personalization styles for cases where that may appear (like no footer and not many items)

Fixes #2062 

http://localhost:4000/components/applicationmenu/example-index.html?colors=80000
- open the menu with the burger button
- whole menu should be color (no black area)
- can also change color in the ... menu

http://localhost:4000/components/applicationmenu/example-personalized-roles.html?colors=206B62
- resize the height of the page so nothing scrolls
- foot should be on the bottom
- resize the height of the page so things scroll, scrollbar will be on the outside of all the three parts of the menu, and the footer will be at the bottom no matter what (you may now have to scroll to it)


davidcarlsonberg:
@tmcconechy I don't see a `...` menu on http://localhost:4000/components/applicationmenu/example-index.html

tmcconechy:
@davidcarlsonberg  OK yeah good call i took it off and forgot to add it back will fix, useful for testing

davidcarlsonberg:
I still see the scrollbar inside the menu...maybe I'm not understanding correctly. Apologies for the aspect ratio.
scrollbar (2)

nickwynja:
> I still see the scrollbar inside the menu...maybe I'm not understanding correctly.

This appears to be the expected behavior. "Outside" refers to the scope of the accordion. In prior iterations, only the accordion scrolled within it's own div. In this case, the whole menu scrolls within it's vertical container as expected.

tmcconechy:
OK @davidcarlsonberg i think everything is ok.

- @nickwynja described it. I mean the menu is on the oustide of the 3 sections of the app menu. Compare to http://master-enterprise.demo.design.infor.com/components/applicationmenu/example-personalized-role-switcher.html where its on the middle container.
- http://master-enterprise.demo.design.infor.com/components/applicationmenu/example-index never had the .. menu so i updated the steps. Didnt bother adding it there as its not important

PR #2119, opened 2019-05-08
Adds backward-compat for accessible required input field labels [v4.18.x]

### LINKED ISSUE #2118: Required asterisks no longer appearing
**Description**
A recent change to the required asterisk strategy has caused all our asterisks to disappear. They used to be an ::after pseudo element to the label.required element (CSS solution), but are now inserted by DOM selector when sohoxi.js script is loaded (one-time DOM manipulation solution).
The new solution will not work for _any_ dynamic HTML that is rendered after the sohoxi initialization (which for us is the complete Angular application)

This is the change introduced in v4.18.0:
https://github.com/infor-design/enterprise/pull/1877/files#diff-ce2b1986d94d1af0464dc730fd2bbe29L100

**Expected behavior**
CSS solution for required asterisk kept in place,

**Version**
- ids-enterprise: v4.18.0

EdwardCoyle:
This PR reverses some changes that removed the display of the red asterisk from required form field labels.  The previous method of displaying these asterisks was not accessibility friendly, but has been added back to provide backwards compatibility, and time for deprecation.  The new, more-accessible method of using these asterisks has been made "opt-in" by way of adding an additional CSS class to the label.

Additionally, the Form component's automatic appending of the accessibility-friendly markup has been changed to check for the class.  Documentation and a changelog note for these changes have also been added.

- Related to #421
- Reverses some changes in #1877
- Closes #2118

- Navigate to http://localhost:4000/components/signin/example-index.html. Ensure that only one asterisk is present on each required form field.
- Navigate to http://localhost:4000/components/input/example-index.html. Ensure that only one asterisk is present on each required form field.
- Navigate to http://localhost:4000/components/input/example-required-fields.html. See that all labels are using the asterisk span (use dev tools to confirm).
- Navigate to http://localhost:4000/components/input/test-legacy-required-fields.html. See that all labels are using the original `::after` pseudo-element (use dev tools to confirm).

**Included in this PR:**
- ⛔️ An e2e or functional test for the bug or feature.
- ✅ A note to the change log.

----
paging @volante007

### DIFF
--- a/app/views/components/input/example-required-fields.html
+++ b/app/views/components/input/example-required-fields.html
@@ -1,0 +1,25 @@
+<div class="row">
+  <div class="six columns">
+    <h2 class="fieldset-title">Accessibility-Friendly Required Fields</h2>
+    <p>The fields on this page automatically have accessible markup added to their labels to assist screen-reading tools. For more information, please see IDS Github Issues <a class="hyperlink" href="https://github.com/infor-design/enterprise/issues/421" target="_blank">#421</a> and <a class="hyperlink" href="https://github.com/infor-design/enterprise/issues/2118" target="_blank">#2118</a></p>
+  </div>
+</div>
+
+<div class="row">
+  <div class="one-half column">
+    <div class="field">
+      <label class="required accessible" for="first-name">First Name</label>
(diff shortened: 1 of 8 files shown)


EdwardCoyle:
Updated with new examples and instructions for testing

PR #2123, opened 2019-05-08
1995 - Large Color picker is cut off at mobile sizes

### LINKED ISSUE #1995: Color picker Sizes Component: Responsive alignment issue for Large color picker
**Describe the bug**
Color picker Sizes Component: Responsive alignment issue for Large color picker 

**To Reproduce**
1. Go to url: http://4180-beta0-enterprise.demo.design.infor.com/components/colorpicker/example-sizes.html
2. Make the page responsive
3. Scroll down to "Large" 
4. The frame is cut off. Please see screen shot attached

**Expected behavior**
The frame should not be cut off

**Version**

**Screenshots**
ColorPicker ALignment Issue

**Platform**
 - Device (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g.iOS 8]
 - Browser Name [FF]
 - Browser Version [66.02 (Official Build) (64-bit)]

**Additional context**
None

davidcarlsonberg:
Color picker inputs are not responsive to smaller breakpoints.
This PR puts an example colorpicker into the responsive grid at `form/example-inputs.html` to show the new rules enabling a responsive format. This PR also creates new responsive rules for a generic colorpicker at different input sizes outside of a responsive grid, solving the original issue.

Closes #1995.

- Pull branch and run app
- Visit http://localhost:4000/components/colorpicker/example-sizes.html
  - ensure all colorpickers work as expected and are responsive to smaller breakpoints.
- Visit http://localhost:4000/components/form/example-inputs.html
  - ensure the colorpicker acts responsively within the grid

**This PR includes**:


davidcarlsonberg:
Curious to get your thoughts on this @tmcconechy as it will involve updating the baseline image to pass the visual regression test.

tmcconechy:
Taking a look at this I realized there is something i similarly made already that works similar. Check out http://localhost:4000/components/form/example-inputs.html

The only difference is for this example we want something that is 100% on mobile only so maybe its as simple as something like:

```
@include respond-to(phone) {
   ....width: 100%
}
```

If you add form-responsive class to this example on the parent then it might just work as we expect already? And then we don't need to introduce a bunch of classes no one might use.

Or maybe a variation call `form-responsive-mobile` that does it for all fields including this one on mobile?

src/components/colorpicker/_colorpicker.scss
```diff
@@ -198,7 +198,6 @@
     border: 0;
     max-height: 34px;
     padding: 9px 0 8px 10px;
-    text-overflow: ellipsis;
```
tmcconechy: I think we need to keep this as that made the smaller colorpicker show ellipsis on the text when cut off (previous PR)

davidcarlsonberg:
> 1. Was the classes like `colorpicker-min-width` already there?
Yes, those classes were already there and they set explicit widths on the inputs.
> 2. Can we add an example to components/forms like the other one so we remember this is possible to do and see the inputs working?
I don't think this is needed as this is now using the existing `form-responsive` class.
> 3. When i look at this example http://localhost:4000/components/colorpicker/example-sizes.html do we need the extra classes anymore like `colorpicker-min-*` and `colorpicker-input-*`?
In a refactor, I believe it would be a good idea to remove these. But these were not added here and addressing them is a little out of scope for this ticket. Please see answer to 1.
> 4. Related to 3 when i open this on chrome - pixel 2xl simulator, i dont see it flowing to 100% on mobile. Are we just trying to not break the layout with it or should we just make this go to 100% there because of the new `responsive-form-mobile` class? I think going to 100% simplifies it more?
They don't adhere strictly to 100% because more needs to change including the container width but can add one more rule to change `auto` to `100%` at the smallest size for all.

PR #2128, opened 2019-05-08
2127 - Separate personalize styles into own file

### LINKED ISSUE #2127: Personalization: Styles in JS file are hard to maintain
The `cssRules` var defined in [personalize.js:184-269](https://github.com/infor-design/enterprise/blob/master/src/components/personalize/personalize.js#L184-L269) is difficult to maintain.

* Separate out style definitions to separate file
* Format style definitions for better readability and maintainability

n/a

**Additional context**
n/a

nickwynja:
Improves maintainability of personalized CSS styles by separating into standalone file and formatting. This PR **does not** change any specific styles, only improves the readability of the styles.

Closes #2127

Compare there are no visual differences between:

* https://master-enterprise.demo.design.infor.com/components/personalize/example-classes.html
* http://localhost:4000/components/personalize/example-classes.html

or 

* http://master-enterprise.demo.design.infor.com/components/applicationmenu/example-personalized.html?theme=uplift&colors=7025B6
* http://localhost:4000/components/applicationmenu/example-personalized.html?theme=uplift&colors=7025B6

You can also `diff` the contents of `<style id="soho-personalization">` from each of `master` and this branch, but it's a little difficult since there were previously some missing semicolons.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@
 
 - `[General]` Updated to jquery 3.4.1 to fix a jquery bug seen occasionally. ([#2109](https://github.com/infor-design/enterprise/issues/2109))
 - `[Demo App]` Fixed CSP and handling of image paths for better support of images in examples on IDS demo sites (demo.design.infor.com). ([#1888](https://github.com/infor-design/enterprise/issues/1888))
+- `[Personalize]` Separated personalization styles into standalone file for improved maintainability. ([#2127](https://github.com/infor-design/enterprise/issues/2127))
 
 (nn Issues Solved this release, Backlog Enterprise nn, Backlog Ng nn, nn Functional Tests, nn e2e Test)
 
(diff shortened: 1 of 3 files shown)


tmcconechy:
Looks good except that file will not lint. https://travis-ci.com/infor-design/enterprise/jobs/198709424#L522 Maybe have to ignore some rules at the top / move some things around like the export at the bottom instead.

Wondering if this file can some how be a sass file? (Just thinking as it would look better in the editor)
But probably wont work.

nickwynja:
@tmcconechy:
> Looks good except that file will not lint.

Fixed.

> Wondering if this file can some how be a sass file? (Just thinking as it would look better in the editor)
> But probably wont work.

There might be something we could do there but I imagine it would require some additional loader or cross-compiler or something. For now, you should be able to just set your editor's local filetype to CSS for it to load the CSS syntax highlighter though it might get confused by the `${color}` syntax.

PR #2131, opened 2019-05-09
1963 - Personalize form examples

### LINKED ISSUE #1963: Personalization: Need ability for landmark context form to have personalization background color
Need landmark context form to have personalization color for composite forms. 

Would like to add .is-personlizable class to format some elements in it that arent yet supported.

**Additional context**
Note: not using the .compositeform jquery component or ep markup for this.

image

tmcconechy:

Adds some landmark examples of forms with a personalized header and adds layout fixes and personalization css to support this.

Fixes #1963 

http://localhost:4000/components/personalize/example-form2.html
http://localhost:4000/components/personalize/example-form.html
- in both examples change colors on the .. menu and just check it looks nice.


app/views/components/personalize/example-form.html
```diff
@@ -0,0 +1,177 @@
...
+
+<script>
+  $('body').on('initialized', function () {
+    if ($('.personalize-header').css('background-color') === 'rgba(0, 0, 0, 0)') {
+      // None set
+      $('html').personalize({ colors: { header: '2578A9' } });
+
+      // Intercept the Default Item as it cant work here 100%
+      $('li.is-default').removeClass('is-default');
+    }
```
pwpatton: @tmcconechy do we need to do anything in our app like this? or does just using the new classes take care of it?

tmcconechy: i was thinking this was just in the demo app we have case, as we are using that menu button to do this. So this makes that example not delete the style sheet and instead set a color. As long as you are setting colors you should be ok. But a reset to default might just remove the header color. So something to test.

PR #2133, opened 2019-05-09
392 - Fixed context menu with datagrid was not properly destroyed

### LINKED ISSUE #2140: Datagrid: Pop-up menu in datagrid fires multiple events
**Describe the bug**
When a popup menu is used in a datagrid, the event listener is not being remove after the previous click. Clicking on the same menu will attach another event listener and will cause to fire multiple events. 

**To Reproduce in NG**
1. Run the sample code 'Datagrid - Service' from enterprise-ng-master project
2. In the '_Actions_' column, select the first icon to show the popup
3. Select '_Action One_'. Console log will show the selected menu.
4. Select '_Action One_' again on the the same row.
5. Console will log twice.

**To Reproduce in EP**
1. go to http://localhost:4000/components/datagrid/test-all-formatters.html
2. scroll over to the actions (...) columns
3. open console
4.click .. and select one item and see one event
5. do it again and see two events (doubles every time)

**Expected behavior**
Clicking a menu on the popup should only trigger the event once every single time.

**Version**
- ids-enterprise-ng: v5.3.0

**Screenshots**
soho_datagrid_popup_multiple_events

[link](https://gfycat.com/leadingwindinggazelle) if gif does not work

**Platform**
 - OS Version: [Windows 10]
 - Browser Name [chrome]
 - Browser Version [74.0.3729.131 (Official Build) (64-bit)]

deep7102:
Fixed when using context menu with datagrid was not properly destroyed which being created multiple times.

Closes infor-design/enterprise-ng#392
Closes #2140

http://localhost:4000/components/datagrid/test-destroy.html
- Open above link
- Open developer tools
- Run on console `$('.popupmenu-wrapper').length;`
- Should result 1
- Right click on both rows to open context menus
- While menu is open, run on console `$('.popupmenu-wrapper').length;`
- Should result 2
- Click on top buttons "Destroy" then "Invoke"
- Run on console `$('.popupmenu-wrapper').length;`
- Should result 1
- Right click on both rows to open context menus
- Again while menu is open, run on console `$('.popupmenu-wrapper').length;`
- Should result 2
- So if you run while open menu result should be 2 otherwise 1 as many time menu opens/close or destroy/invoke the grid, `$('.popupmenu-wrapper').length;` 

http://localhost:4000/components/datagrid/test-all-formatters.html
- Open above link
- Open developer tools
- Scroll over to the `Action` column
- Click any button `...` to open a menu
- Click another row
- See console there should not be any error

### DIFF
--- a/app/views/components/datagrid/test-destroy.html
+++ b/app/views/components/datagrid/test-destroy.html
@@ -1,0 +1,82 @@
+<div class="row">
+  <div class="twelve columns">
+    <button type="button" class="btn-secondary" id="destroy">Destroy</button>
+    <button type="button" class="btn-secondary" id="invoke">Invoke</button>
+  </div>
+</div>
+
+<div class="row top-padding">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<ul id="grid-actions-menu" class="popupmenu">
+  <li><a href="#">Action One</a></li>
+  <li><a href="#">Action Two</a>
+    <ul class="popupmenu">
+      <li><a href="#">Action Three</a></li>
+      <li><a href="#">Action Four</a></li>
+    </ul>
+  </li>
+</ul>
+
+<ul id="grid-header-menu" class="popupmenu">
+  <li><a href="#">Sort Column</a></li>
+  <li><a href="#">Hide Column</a></li>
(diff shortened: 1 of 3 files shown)


deep7102:
Hi @tmcconechy update now to fix #2140

Issue #643, opened 2018-08-22
UI-level Keyboard Management Module


EdwardCoyle:
**NOTE:** This ticket is barebones because we need to further flesh this idea out.  Feedback/edits are welcomed and encouraged.

**Is your feature request related to a problem? Please describe.**
Some use cases have become apparent that make it beneficial to have keyboard commands and keypress event listeners handled at a more global level, as opposed to on each individual component instance.

**Describe the solution you'd like**
Some features that the keyboard management system would handle:

- Detection of whether or not a software keyboard on a mobile device is currently active.  Whenever `focusin`/`focusout` events bubble to the top-level Soho UI component, a check is done to see if an input field, textarea or otherwise keyboard-controlled Soho component is the active element, and provides an API that returns a boolean result providing the status, as well as a reference to the element.  This could be used to assume that the keyboard is present, and the behavior of the UI can change based on the visible keyboard.  _Example case:_  Delay the rendering of a popupmenu when the keyboard is present, but immediately render with no delay if the keyboard is hidden.

- might also have a hot key mapper. for external use to map keys to app functionality and for mapping our internal functions to hot keys. And show a dialog with all active hot keys on a hot key or menu item for the user

- make sure all events have the right tense.. fx activated, beforeactivated, afteractived (not activate, beforeactivate, afteractivate). (This change should be changed in NG as well and could be made backwards compatible)

- make sure all event arguments are of type object with consistently named arguements fx (data, id, idx ect)

- consider making all events just call back functions on the api. fx onBeforeShow, onSelected ect

**TODO** - add more functionality

EdwardCoyle:
I thought I needed some basic scaffolding for this related to an issue I was working through, but ended up not needing it at that time.  I've pushed the work to [this branch](https://github.com/infor-design/enterprise/tree/643-global-keyboard-manager) for someone to pick up later, since it's got all the hookups into core completed.

PR #2172, opened 2019-05-16
1906 - Add a fixes so that beforeSelect can be vetoed

### LINKED ISSUE #1906: Datagrid: Add events to allow multiselect only when modifier key is pressed
In M3, we need to be able to do a multiple select, but only when the user is pressing the Ctrl/Command (mac) key at the same time. This is a behavior that thousands of M3 users are accustomed to. The M3 H5 client is currently being rewritten using ids-enterprise-ng components and this feature enhancement for the datagrid is very much needed. 

Option 1:
The optimal solution would be to have a property e.g. modifierKey (available for `multiple` selection datagrids) which would take a supported (by you guys) value e.g. ctrl or alt. Lets say the value ctrl is given, the datagrid control would check If ctrlKey or metaKey (mac) is available on the triggered event and its current value. If the value of the mappings for the modifyer key is true, then go ahead and add the row to the list of selected rows, otherwise unselect all rows and add only the current row to the list of selected rows. If this modifyerKey property is not present on the datagrid options object, continue with default behavior (as today). 

Option 2:
A different solution could perhaps be to add `beforeRowActivated` (or similar event before a row is actually activated or selected) for `multiple` selection datagrids too (currently only available for mixed selection datagrids?). You would have to pass the original event (not only the `row` and `item` which is the case for the `mixed` selection datagrids today), so that the developers can check if `event.ctrlKey` or `event.metaKey` exists and if anyone is true. 

An alternative solution I've tried is to create an angular directive to be added to the same div as the `soho-datagrid` one which has a HostListener for click event. I can then inject the datagrid in the directive and unselect all rows if the modifyer key of my choice is not true. This works to some extent, but not when the rows are selected using keys (tabbing and selecting with enter for example). Also some events are not triggerd by users, so it is not a good solution for that reason too.

tmcconechy:

Adds a beforeSelect callback to datagrid so that you can veto row selection. The requirements listed needed to also know if CTRL or shift is pressed, so re-added the keyboard module to start up #643 , this had the functionality needed to see the pressed keys anytime. To let the events run through had to fix up the contextmenu function. Bonus: can now right click datagrid as long as no context menu functionality is attached for inspection.

Fixes #1906 

http://localhost:4000/components/datagrid/example-beforeselect-veto.html
- Selecting rows is not possible.
- But if you hold down CTRL key or Command (on Mac) and click a row it will select.
- This mimics M3 functionality in a non invasive way

http://localhost:4000/utils/example-keyboard.html
- hold a key of any kind and it will show

- Everything is nicely covered by the new tests
- Retest Datagrid Context Menu
http://localhost:4000/components/datagrid/test-tree-contextmenu.html
http://localhost:4000/components/datagrid/test-contextmenu.html
http://localhost:4000/components/datagrid/test-contextmenu-dynamic.html
http://localhost:4000/components/datagrid/test-contextmenu-in-events.html (new test)

### DIFF
--- a/app/views/components/accordion/test-close-children-on-collapse.html
+++ b/app/views/components/accordion/test-close-children-on-collapse.html
@@ -142,7 +142,7 @@
         $('#multiselect').data('dropdown').open();
         $('#datepicker').data('datepicker').open();
 
-        countTimer = 5;
+        countTimer = 50;
         count();
       });
 
--- a/app/views/components/accordion/test-close-children-on-collapse.html
+++ b/app/views/components/accordion/test-close-children-on-collapse.html
@@ -142,7 +142,7 @@
         $('#multiselect').data('dropdown').open();
         $('#datepicker').data('datepicker').open();
 
-        countTimer = 50;
+        countTimer = 5;
         count();
       });
 
--- a/app/views/components/datagrid/example-beforeselect-veto.html
+++ b/app/views/components/datagrid/example-beforeselect-veto.html
@@ -1,0 +1,59 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+      var grid,
+        columns = [],
+        data = [];
+
+      // Create Some Sample Data
+      data.push({ id: 1, productId: 2142201, productName: 'Compressor', activity:  'Assemble Paint', quantity: 0, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action'});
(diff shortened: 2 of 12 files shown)


EdwardCoyle:
Also noting that this PR probably also covers most -- if not all of -- #643

tmcconechy:
Right except for the breaking consistency changes.

tmcconechy:
Good can make banana bread 😛

hernebrink:
@tmcconechy Do i need to create a ticket for enterprise-ng project in order to get this available for the angular component?

tmcconechy:
I usually track that on tickets like this https://github.com/infor-design/enterprise-ng/issues/472
I went ahead and added this so it was for sure in https://github.com/infor-design/enterprise-ng/pull/491

Issue #2138, opened 2019-05-10
Datagrid cloumns misaligned after flipping the screen to Landscape in iOS device


janahintal:
**Describe the bug**
Datagrid cloumns misaligned after flipping the screen to Landscape in iOS device

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html?locale=he-IL# in portrait mode
2. Wait for the grid to load then flip the screen to Landscape

**Expected behavior**
Grid columns should be aligned

**Version**
ids-enterprise: [v4.18.0] beta0 

**Screenshots**
<img width="1178" alt="Screen Shot 2019-05-08 at 6 26 23 PM" src="https://user-images.githubusercontent.com/45757016/57370366-9510a180-71c2-11e9-976c-07bd6730ecbe.png">

**Platform**
OS: iOS 12
Device: iPhone X

tmcconechy:
This should no longer be an issue.

PR #2182, opened 2019-05-20
Datagrid: Filter operator popdown is not updated when the filter via a script

### LINKED ISSUE #2129: Datagrid: Filter operator popdown is not updated when the filter via a script
**Describe the bug**
If the filter is applied via the api, over a user clicking to change the operator then it is no longer correct in the popdown as only the filter icon is updated.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/example-filter.html
2. in the in stock color filter on only selected items
2. run the following in the console
`$('#datagrid').data('datagrid').applyFilter([{columnId: "inStock", operator: "selected", value: ""}], '')`
3. Then click on the filter for the "In Stock" field

**Expected behavior**
The icon and popdown are in sync and the filter is cleared.

**Version**
v4.18.0

**Screenshots**
image
After the script runs
image
opening the operator for the filter, not that it changes the filter without actually selecting a value.

**Platform**
 - OS Version:  Windows 10
 - Browser Name chrome

marknolaninfor:
The filter operator will render with the correct selected item. Rendering the header will maintain the current filter.

Closes #2129 

Goto 'http://localhost:4000/components/datagrid/test-filter-applyfilter.html'
Click on the apply filter, page and then the render header button
Open the In stock operator to check that it is applied.
Click on the clear filter button


src/components/datagrid/datagrid.js
```diff
@@ -791,6 +791,7 @@ Datagrid.prototype = {
     } else {
       this.clearHeaderCache();
       this.renderRows();
+      this.renderHeader();
```
tmcconechy: Is this needed? I suspect this causes a bug with the PR. If you now to go:

http://localhost:4000/components/datagrid/test-save-settings.html
- type 2 in the price field and wait a sec then type 1 
- you cant as the filter is reloaded and the focus is out of the field.

I think if you can add that back it would fix this and we just need to make sure to save and restore the filter conditions on header render calls.

marknolaninfor: I have altered this again, so that the header is not rerendered on a 'filtered' page, but now rerender the grid, as the clearHeaderCache was causing issue with the header and rows being out of sync.  The change should correct #2138 

http://localhost:4000/components/datagrid/test-save-settings-manual.html was very slow because the initial read was 1000 rows, and then when the settings finally got applied renders, and resetPager was being called when nothing had actually changed, eg the same setting was being applied that was currently being used.

All the events are still triggered but any render should not happen when they should not.

marknolaninfor:
I found an additional issue if the datagrid filters the first record and then the orientation is changed, that the header and body widths are not in sync.

Goto 'http://localhost:4000/components/datagrid/test-filter-applyfilter.html'
Set a Filter like out of stock
Change the orientation

marknolaninfor:
http://localhost:4000/components/datagrid/example-frozen-columns
Will render differently as I have assumed that the center container will be the one that fits within the right and left fixed containers.

marknolaninfor:
issue paging on http://localhost:4000/components/datagrid/test-paging-select-indeterminate-single.html

image

marknolaninfor:
Lookup issue corrected, paging render 

http://localhost:4000/components/lookup/example-paging.html
http://localhost:4000/components/lookup/example-minwidth.html

tmcconechy:
@marknolaninfor FYI i also fixed some related things here https://github.com/infor-design/enterprise/pull/2199

marknolaninfor:
http://master-enterprise.demo.design.infor.com/components/datagrid/example-reorder.html is broken because of this change as it only renders if the input parameter is different from the current settings.

The render should be moved outside, if you want a user to page to a different page and then alter the columns and also maintain the current paging information.

    if (columnsChanged) {
      this.rerender();
      this.resetPager('updatecolumns');
    }

PR #2194, opened 2019-05-20
2091 - Improvements based on QA feedback

### LINKED ISSUE #2091: Datagrid: dirtyRow doesn't include rows and cells whose status changed via rowStatus
Describe the bug
In Landmark, the values of a datagrid maybe updated externally.  For example, a cell contains a button that when clicks opens a model dialog.  The user updates field values and those values are to be reflected in the row.  Fields may or may not have a visible cell to show the updated value. 

1. Using the attached example-change-cell-externally.html
2. Click on the 'Change Cell' toolbar button
3. Notice in the displayed alert that the dirty row count doesn't reflect the row that is 'in-progress'

Expected behavior
Expect that rows and cells whose status has been set to 'in-progress' or 'error' via the  rowStatus method should be included.  Similar to how it worked in 3.5 in inforControlsCombined.js via getDirtyRows.

Additionally need a method to setDirtyCell to visually show that a cell has been updated externally via updateRow method.

tmcconechy:

Based on QA comments, made an option to save the indicator across page

Fixes #2091

http://localhost:4000/components/datagrid/test-dirty-change-cell-externally.html
- press change cell
- also edit a cell on row 2 for example quantity
- go to page two
- go back to page one
- should now show both as dirty still on page one

### DIFF
--- a/app/views/components/datagrid/test-dirty-change-cell-externally.html
+++ b/app/views/components/datagrid/test-dirty-change-cell-externally.html
@@ -113,7 +113,8 @@
       // 1. dirtyRows should also include rows that are in error or in-progress ... like it did in 3.5 getDirtyRows
       // 2. method to setDirtyCell to visually show cell has been updated
       var modifiedRows = gridApi.getModifiedRows();
-      gridApi.setDirtyIndicator(0, gridApi.columnIdxById('quantity'), true);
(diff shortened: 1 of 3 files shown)


tmcconechy:
Accidently got another PR on here. going to reraise 😞

PR #2213, opened 2019-05-22
Datagrid Filter: Filtering for Dropdown Editors not working when valu…

### LINKED ISSUE #1879: Datagrid Filter: Filtering for Dropdown Editors not working when value is a number.
**Describe the bug**
When the value set in the options for a dropdown is a number, the filtering for datagrid is not working. This is the sample options for our app:

image

The values are numbers, and filtering the grid doesn't work. Tried replicating it in one of the IDS test pages, kindly see below.

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/test-filter-editable.html'
2. These are the options available for the "Activity" column:

image

If we try to filter, it works properly:

image

3. If we change the options to have numbers as values (still a string, though) the filter won't work:
image

image

**Version**
v. 4.17.0

marknolaninfor:
It the 'multiselect' values are used and the allowed values are numeric strings, "1", then the filter does not display the correct rows. Also found an issue with clearing the filter that a multiselect filter could set the value to an option if the value was empty, "".

Closes #1879 

goto 'http://localhost:4000/components/datagrid/test-filter-applyfilter.html'
set the activity filter to "number 1"
Click on the clear filter button


marknolaninfor:
Is this pull request going into 4..19?

marknolaninfor:
Added additional fix found, that I caused in #2182 

http://master-enterprise.demo.design.infor.com/components/datagrid/example-reorder.html is broken because of this change as it only renders if the input parameter is different from the current settings.

The render should be moved outside, if you want a user to page to a different page and then alter the columns and also maintain the current paging information.

if (columnsChanged) {
  this.rerender();
  this.resetPager('updatecolumns');
}

Steps:
http://localhost:4000/components/datagrid/example-reorder.html
Page to the second page 
Drag the columns.

tmcconechy:
@marknolaninfor Yes i think this should as it broke something. But we have to PR this into branch 4.19.x I'll close this and do another PR for you to 4.19.x

PR #2242, opened 2019-05-23
2228 - Fix sort indicator was not showing after paging with datagrid [v4.19.x]

### LINKED ISSUE #2228: Datagrid Paging Client Side: Sorting arrow disappears when changing page
**Describe the bug**
Datagrid Paging Client Side: Sorting arrow disappears when changing grid page.

**To Reproduce**
1. Go to http://4190-beta0-enterprise.demo.design.infor.com/components/datagrid/example-paging-client-side.html
2. Click Id column header to sort
3. Click the Next Page button, then Click Previous Page
4. See issue: Sorting arrow icon on Id column header disappears

**Expected behavior**
Sorting arrow icon should stay even after changing grid page.

**Version**
- ids-enterprise: v4.19.0-beta0

**Screenshots**
Screen Shot 2019-05-23 at 7 40 01 PM

**Platform**
 - All Devices, OS, and browsers.

**Additional context**
No additional context.

deep7102:
Fixed sort indicator was not showing after paging with datagrid.

Closes #2228

- Navigate to http://localhost:4000/components/datagrid/example-paging-client-side.html
- Click Id column header to sort
- Click the Next Page button, then Click Previous Page
- See sorting arrow icon should stay even after changing grid page

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -49,6 +49,7 @@
 - `[Datagrid]` Fixed an issue for xss where console.log was not sanitizing and make grid to not render. ([#1941](https://github.com/infor-design/enterprise/issues/1941))
 - `[Input]` Improved alignment of icons in the uplift theme input components. ([#2072](https://github.com/infor-design/enterprise/issues/2072))
 - `[Datagrid]` Added an onBeforeSelect call back that you can return false from to disable row selection. ([#1906](https://github.com/infor-design/enterprise/issues/1906))
+- `[Datagrid]` Fixed an issue where sort indicator was not showing after paging. ([#2228](https://github.com/infor-design/enterprise/issues/2228))
 - `[Listview]` Improved accessibility when configured as selectable (all types), as well as re-enabled accessibility e2e tests. ([#403](https://github.com/infor-design/enterprise/issues/403))
 - `[Locale]` Synced up date and time patterns with the CLDR several time patterns in particular were corrected. ([#2022](https://github.com/infor-design/enterprise/issues/2022))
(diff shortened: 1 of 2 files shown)


tmcconechy:
This works ok. But i notice that the sort indicator "blinks" when you change page but every thing else seems stable. It didnt used to do that on http://4180-enterprise.demo.design.infor.com/components/datagrid/example-paging-client-side.html

I wonder if some other issue is going to work up here. We might want to investigate this more. 

cc @marknolaninfor

marknolaninfor:
There is an issue in loadData, as it would render just the rows, and this could cause the colgroup to be different between the header and the body. There was a change to rerender the complete grid and this is likely to have caused a flicker.

A better way would be to within loadData revert this change to 

` // Resize and re-render if have a new dataset
    // (since automatic column sizing depends on the dataset)
    if (pagerInfo.type === 'initial') {
      this.clearHeaderCache();
      this.restoreUserSettings();
      this.renderRows();
      this.renderHeader();
    } else
      this.clearHeaderCache();
      this.renderRows();
    }`

There is then the issue after clearHeaderCache, as the loaded data could have different widths. Example paging on 'http://localhost:4000/components/datagrid/test-paging-select-indeterminate-single.html'

image

The rerender of the whole grid was an over kill, as the real problem is that the colgroup values, and min-width are out of sync.

image

marknolaninfor:
I have created an alternate fix in #2243

tmcconechy:
Lets go with https://github.com/infor-design/enterprise/pull/2250 as it seems to work a little bit nicer without any flicker.

PR #2262, opened 2019-05-28
2258 - Allow manual personalization [v4.19.x]

### LINKED ISSUE #2258: Personalize: Personalization overlay doesn't go away when managing own style sheets
**Describe the bug**
The way landmark is doing their own themes the overlay can remain after recent changes.

**To Reproduce**
1. Go to http://usspnppatton.infor.com:8080/hcm/EmployeeSelfService/home?csk.cli=true&csk.showusingxi=true&csk.HROrganization=7004
emp02/lawson
2. Click refresh a number of times and you'll see the entire browser screen just looking gray.
3. If you look at the elements you'll see personalize-overlay is still there, delete it and the app displays.

**Expected behavior**
ctrl-shift-refresh causes it to work, but just browser refresh a couple times causes it to fail it doesnt seem like we need to insert the style sheet and overlay again if its already there.

**Version**
- ids-enterprise: [v4.19.0]

**Screenshots**
MicrosoftTeams-image

**Additional context**
Also seeing issues with:
- the color picker setting the wrong classes
- the class in the head being wrong.

tmcconechy:

Landmark used manual themes which would cause an extra overlay to occur in some cases. In addition fixed some bugs:
- Color picker was not initializing
- Added a config to skip default Personalization Init (This may not be needed to solve landmarks issue 100% but just in case)
- Fix to prevent the theme and overlay from adding extra times

Fixes #2258 

1. Go to: http://localhost:4000/components/personalize/example-form.html?theme=uplift&variant=dark

    - look at the theme on the html tag
    - use theme theme switcher and make sure the tag is correct
    - debug personalize init and make sure its not been called way too many times
    - try with more url params like `theme=soho&variant=contrast`

2. Go to: http://localhost:4000/components/color-picker/example-index
    
    - open the picker
    - pick an item in about he middle of the color set
    - try with all three theme variants
    - make sure the checkbox and borders look ok and you see no errors

3. Landmark
- [x] Verify in landmark cc @pwpatton

### DIFF
--- a/src/components/personalize/personalize.js
+++ b/src/components/personalize/personalize.js
@@ -13,7 +13,8 @@ const PERSONALIZE_DEFAULTS = {
   colors: '',
   theme: '',
   font: '',
-  blockUI: true
+  blockUI: true,
+  noInit: false
 };
 
 /**
@@ -26,7 +27,8 @@ const PERSONALIZE_DEFAULTS = {
  * @param {string} [settings.theme] The theme name (light, dark or high-contrast)
  * @param {string} [settings.font] Use the newer source sans font
  * @param {boolean} [settings.blockUI=true] Cover the UI and animate when changing theme.
-*/
+ * @param {boolean} [settings.noInit=false] If true, prevents automatic setup of personalized theme/colors/font, allowing for manual triggering at a more convenient time.
+ */
 function Personalize(element, settings) {
   this.element = $(element);
   this.settings = utils.mergeSettings(this.element[0], settings, PERSONALIZE_DEFAULTS);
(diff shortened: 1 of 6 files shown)


pwpatton:
This issue fixes the personalization overlay problem where it didn't go away leaving you with a blank screen.

src/components/personalize/personalize.js
```diff
@@ -258,6 +265,13 @@ Personalize.prototype = {
   */
   setTheme(incomingTheme) {
     const $html = $('html');
+    if (!incomingTheme) {
+      return;
+    }
```
clepore: Not a big deal, but typically its better to define...
```html
const $html = $('html');
```
...after the possible return statement to save memory allocation.

PR #2283, opened 2019-05-31
2264 - Datagrid: Non Displayed Validation Error Tooltip

### LINKED ISSUE #2264: Datagrid: Non Displayed validation message tooltip is not displayed on hover
**Describe the bug**
Non Displayed validation errors, eg validation messages on a non displayed page, would have a tooltip so that the user could navigate to that page and correct the issue. The tooltip no longer displays.

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/test-editable-paging-validation.html'
2. Remove the Product Name value from the first row
3. Page to the next page (the validation will run and display an icon on the toolbar)
4. Hover over the icon to see the tooltip message (Not displayed)

**Expected behavior**
The tooltip should say that there is an error on the first page.

**Version**
4.19

marknolaninfor:
Display the tooltip for the Non displayed validation Errors

Closes #2264

- Go to http://localhost:4000/components/datagrid/test-editable-paging-validation.html
- Remove the Product Name value from the first row
- Page to the next page (the validation will run and display an icon on the toolbar)
- Hover over the icon to see the tooltip message (Not displayed)


tmcconechy:
A few minor issues with this one. We could fix them later.
1. datagrid.js has a couple warnings on trailing spaces
2. the example added http://localhost:4000/components/datagrid/test-editable-paging-validation-notoolbar.html when you add the message to the second page the tooltip is in the wrong position when it appears. When it goes away the space is not reclaimed. I think this is maybe a case we should ignore. Should probably have a toolbar for this to work.
3. The events namespaces `mouseenter.gridtooltip` are already used in placed. This might have a problem so probably need a unique namespace.
4. Should the new events go into the datagrid destroy?

davidcarlsonberg:
I could see an argument to enumerate the errors indicated in the tooltip. E.g., instead of:
tooltip-errors
It would read:
```
Error message(s) on page(s):
1, 2, 4, and 10
```
But that might be nitpicky.

Some screenshots if it's useful for other reviewers:
error-screenshot
warning-screenshot

marknolaninfor:
@davidcarlsonberg 

The message could be converted to a simple sentence, but my concern would be the translation to other languages. The construction of the final part "1, 2, 4, and 10" needs to be translatable, but currently Locale.Translate is not constructed in a way to accept parameter, eg the page numbers and then translate them in to a meaning full sentence, it is not really the case of just substituting words to have the correct meaning in all languages. 

There are cases of one or two  values being passed, but not n, eg an array [1,2,4,10], to Locale.translate. I guess that this could be a seperate ticket to allow for this, as I agree it would be far better to show the one sentence message over an over load to single line statements.

tmcconechy:
If this feature gets a lot of traction i think we can change it up based on user feedback later

PR #2334, opened 2019-06-06
2333 - Pass additional arguments from tooltip function

### LINKED ISSUE #2333: Datagrid: Tooltip function requires additional parameters to be passed
Describe the bug
The callback parameters being passed to the registered tooltip function for a column is only providing the cell and value.

tooltip.content = col.tooltip(cell, value);

To be useful, especially in Landmark, need the full compliment of arguments that other column callback function receive.

tooltip.content = col.tooltip(row, cell, value, col, rowData, api);

claudenbach:

When this feature was made for landmark a few versions ago some additional arguments to the column.tooltip callback function were left out. This attempts to address that. I made changes to the order of the items on conversation with @tmcconechy as this feature is new and these are more consistent. So changed to tooltip api call back to pass the row, cell, value, col, rowData, and api.

Fixes #2333 

- go to http://localhost:4000/components/datagrid/test-custom-tooltip-dynamic.html
- hover the exception cells
- the tooltip will show row, cell and value

** Other Details**


claudenbach:
Related to https://github.com/infor-design/enterprise-ng/issues/354

tmcconechy:
FYI test would probably go in https://github.com/infor-design/enterprise/blob/master/test/components/datagrid/datagrid-api.func-spec.js and include logic that executes the callback and checks the output is the new data you added

PR #2354, opened 2019-06-10
2355- add attachToBody option to standalone-pager for pageSizeChooser menu [v4.19.x]

### LINKED ISSUE #2355: Pager: Need to attach the page size selector to body.
**Describe the bug**
Because we have a position: sticky menu the pager dropdown menu is not sizing correctly when its open in our app. The existing option attachPageToBody fixes this but the pager doesnt have a way to set it.

**To Reproduce**
You cant you have to use our app.

**Expected behavior**
Add a way to feed the option into the pager popupmenu.

pwpatton:

Because we have a position: sticky menu the pager dropdown menu is not sizing correctly when its open in our app. The existing option attachPageToBody fixes this but the pager doesnt have a way to set it.

Fixes #2355 

- go to http://localhost:4000/components/pager/example-standalone.html
- toggle the attach pagesize menu to body checkbox
- look in the dom and notice the page size menu 
   - if checked the page size chooser menu should be at the dom level
   - if unchecked the page size chooser will show in the pager element.

### DIFF
--- a/src/components/pager/pager.js
+++ b/src/components/pager/pager.js
@@ -95,7 +95,8 @@ const PAGER_DEFAULTS = {
   firstPageTooltip: 'FirstPage',
   previousPageTooltip: 'PreviousPage',
   nextPageTooltip: 'NextPage',
-  lastPageTooltip: 'LastPage'
+  lastPageTooltip: 'LastPage',
+  attachToBody: false
 };
 
 function Pager(element, settings) {
@@ -998,7 +999,7 @@ Pager.prototype = {
       }
 
       // Render the button
-      const pageSizeButton = $(`<button type="button" class="btn-menu">
+      const pageSizeButton = $(`<button type="button" class="btn-menu"  data-options="{attachToBody: ${this.settings.attachToBody}}">
         ${recordHtml}
         ${dropdownIcon}
       </button>`).appendTo(pageSizeLi);
--- a/src/components/pager/pager.js
+++ b/src/components/pager/pager.js
@@ -96,7 +96,7 @@ const PAGER_DEFAULTS = {
   previousPageTooltip: 'PreviousPage',
   nextPageTooltip: 'NextPage',
   lastPageTooltip: 'LastPage',
-  attachToBody: false
+  attachPageSizeMenuToBody: false
 };
 
 function Pager(element, settings) {
@@ -999,7 +999,7 @@ Pager.prototype = {
       }
 
       // Render the button
-      const pageSizeButton = $(`<button type="button" class="btn-menu"  data-options="{attachToBody: ${this.settings.attachToBody}}">
+      const pageSizeButton = $(`<button type="button" class="btn-menu">
         ${recordHtml}
         ${dropdownIcon}
       </button>`).appendTo(pageSizeLi);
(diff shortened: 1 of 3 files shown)


src/components/pager/pager.js
```diff
@@ -95,7 +95,8 @@ const PAGER_DEFAULTS = {
   firstPageTooltip: 'FirstPage',
   previousPageTooltip: 'PreviousPage',
   nextPageTooltip: 'NextPage',
-  lastPageTooltip: 'LastPage'
+  lastPageTooltip: 'LastPage',
+  attachToBody: false
```
tmcconechy: I feel like it would be better and more flexible to add a new setting like `pageSizeMenuOptions` which would be default be

```
      const popupOpts = {
        placementOpts: {
          parent: pageSizeButton,
          parentXAlignment: (Locale.isRTL() ? 'left' : 'right'),
          strategies: ['flip']
        }
      };
```

(line 1019- 1027) then replace the options in there so there might be more control over some of the other popup options

tmcconechy:
Also seems like a lint error https://travis-ci.com/infor-design/enterprise/jobs/206849012

src/components/pager/pager.js
pwpatton: not sure how to do that here. Also how do I specify that through ng?
right now I was going to do this:
```
    <div soho-standalone-pager *ngIf="!hidePagerBar" class="scrollable-flex-footer"
        showFirstButton="true"
        showPreviousButton="true"
        showNextButton="true"
        showLastButton="true"
        attachToBody="true"
 ```

Comment (pwpatton) [reply src/components/pager/pager.js]: continue to set it on the pager options then pass through to MeuOptiosn?

Comment (tmcconechy) [reply src/components/pager/pager.js]: would it be like `[pageSizeMenuOptions. attachToBody]="true"` ?

Comment (pwpatton) [reply src/components/pager/pager.js]: that seems like a bigger change for something that I need today tho. Can we refactor after the 4.19 rleease?

File: src/components/pager/pager.js [hunk 2]
```diff
@@ -95,7 +95,8 @@ const PAGER_DEFAULTS = {
   firstPageTooltip: 'FirstPage',
   previousPageTooltip: 'PreviousPage',
   nextPageTooltip: 'NextPage',
-  lastPageTooltip: 'LastPage'
+  lastPageTooltip: 'LastPage',
+  attachPageSizeMenuToBody: false
```
EdwardCoyle: Could potentially be:
```js
pageSizeMenuSettings: {
   attachToBody: false
}
```
... to go along with my other suggestion below.

src/components/pager/pager.js
```diff
@@ -1021,7 +1022,8 @@ Pager.prototype = {
           parent: pageSizeButton,
```
EdwardCoyle: This whole block would become:

```js
const popupOpts  = utils.extend({}, {
  placementOpts: {
    parent: pageSizeButton,
    parentXAlignment: (Locale.isRTL() ? 'left' : 'right'),
    strategies: ['flip']
  }
}, this.settings.pageSizeMenuOpts);
```

... this, combined with the change above, grabs all the popupmenu settings, instead of just the one, like @tmcconechy was suggesting.

pwpatton: see reply above

PR #2372, opened 2019-06-12
2299 - Fix for search not working in datagrid personalization

### LINKED ISSUE #2179: Listview: Calls Initialize too many times
**Describe the bug**
We noticed that listview is calling .initialize() once per each line in the list. Should only need to call itself once

**To Reproduce**
1. Go to the init in initialize and add a console log
2. Go to http://localhost:4000/
3. See how many times in the console

**Expected behavior**
Should only fire once https://github.com/infor-design/enterprise/blob/master/src/components/listview/listview.js#L388

**Version**
- ids-enterprise v4.19.0 (or any)

**Additional context**
Should make a list view example that initializes something like maybe a target/chart. And make sure all rows get auto initialized.

### LINKED ISSUE #2299: Datagrid: Search on Personalize Columns not working
**Describe the bug**

The search box on the in-built personalize columns dialog is not longer filtering the list of columns present in the associated dialog.  It is highlighting the matching letters.

**To Reproduce**

Clone the master repo of enterprise.

```sh
npm i
npm run start
```

Then navigate to `http://localhost:4000/components/datagrid/test-save-settings.html`. 

Open the DataGrid option menu, and select Personalize.

Enter text into the search field.

**Expected behavior**

When entering text I would expect the list of reduce as appropriate.

**Version**

v4.19 (beta.0 and 1)

**Screenshots**

image

**Platform**
 - Windows and Mac
- Chrome and IE11

**Additional context**

tmcconechy:

It was found that the search stopped working when in previous versions we changed the search logic. Updated the datagrid to use the correct listview logic. While in there fixed a semi-related issue where the listview calls initialize once per list item.

Fixes #2179 
Fixes #2299

2179 - http://localhost:4000/components/listview/example-index.html
- go to page
- open the debugger
- in the source tree, find `locahost:4000/src/behaviors/initialize/initialize.jquery.js`, and add a stop on the line under `$.fn.initialize = function (settings) {`
- should hit this twice when reloading the page (second one is from the listview stack trace, first one from the demo app page)
 
2299 - http://localhost:4000/components/datagrid/example-index.html
- click ... and then personalize columns
- note that a hidden column named hidden is unchecked
- and that the first column named id is "hideable: false" so its disabled
- try searching in the search for column names and it should filter
- try reset (clicking the x)

2299 - http://localhost:4000/components/datagrid/example-frozen-columns.html
- click ... and then personalize columns
- note that the two frozen columns are disabled

2299 - http://localhost:4000/components/datagrid/example-colspan.html
- click ... and then personalize columns
- note that the two spanned columns are disabled

2299 - http://localhost:4000/components/datagrid/example-multiselect.html
- click ... and then personalize columns
- note that the checkbox column is absent from the list (because it has no name)

Try anything else you think of....


EdwardCoyle:
> 2299 - http://localhost:4000/components/datagrid/example-index.html
> - click ... and then personalize columns
> - note that the two frozen columns are disabled

Just noting for other reviewers that this example doesn't have any frozen columns.  However, I was able to see this feature working on http://localhost:4000/components/datagrid/example-frozen-columns.html

tmcconechy:
@EdwardCoyle i did mean that page, updated the description above

tmcconechy:
Merging as was previously approved

PR #2376, opened 2019-06-13
2291 - Fix performance for export methods with datagrid

### LINKED ISSUE #2291: Datagrid: Performance of Table Export methods are significantly slower since 4.4.0 version
**Describe the bug**
We recently upgraded from SoHo Xi 4.4.0 to 4.14.0 and shipped this to our customers. They have reported back that the "Export to Excel" of table data "hangs". On investigation, we found that the problem was with large data sets (approx 1000 or more records) and that the processing of these records seems to take much longer in SoHo Xi 4.14.0 than in SoHo Xi 4.4.0, which, to the user, appears like the screen is hung as the CPU use is maxed out. This performance issues is still occurring in the latest 4.18.1 version. This issue has been raised on Infor Xtreme and support are involved.

For example, in the old code, an export of 2000 records takes 9 seconds in Chrome, but in the latest code it takes 137 seconds. Customers are using much larger data sets than this.

I will attach an example HTML file that contains calls to the exportToExcel and exportToCSV methods in the latest SoHo Xi version, along with a call to the old 4.4.0 code for comparison. The table is built using a loop to duplicate 2000 records, then use the buttons on the top of the page to export using the different methods. Time taken is written to alert box on completion and to the console, if open.

**To Reproduce**
1. Use the attached example
2. Click on Export
3. Compare speed to 4.4.0

**Expected behavior**
Should be near as fast as 4.4.0 if possible

**Version**
- ids-enterprise: [e.g. v4.14.0]

**Platform**
 - Browser Name [chrome] but others tested / same issue

**Additional context**
It looks to me that by changing to arrays, instead of strings, it has caused this performance issue.

deep7102:
Fixed performance for export methods with datagrid.

Closes #2291

- Navigate to http://localhost:4000/components/datagrid/test-export-performance.html
- Click on button `Export to Excel` or `Export to CSV`
- It should not take too much longer time to export/download the file

**Other Details**
- [x] Functional test for the bug.

### DIFF
--- a/app/views/components/datagrid/test-export-performance.html
+++ b/app/views/components/datagrid/test-export-performance.html
@@ -1,0 +1,89 @@
+<div class="row">
+  <div class="twelve columns">
+    <button id="exportExcel" class="btn-primary">Export to Excel</button>
+    <button id="exportCSV" class="btn-primary">Export to CSV</button>
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+    var url = '{{basepath}}api/datagrid-sample-data';
+    var grid;
+    var data;
+    var columns;
+
+    $.getJSON(url, function(res) {
+      data = [];
+      for (var i = 0; i < 2000; i++) {
+        data.push(res[2]);
+      }
+
+    	// Define Columns for the Grid.
+      columns = [];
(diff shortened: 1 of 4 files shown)


tmcconechy:
Seems like one new test is failing https://travis-ci.com/infor-design/enterprise/jobs/207860197#L1513 maybe needs more testing on the test 👍

deep7102:
Hi @tmcconechy updated now, example page

PR #2378, opened 2019-06-14
2313-Add minWidth and maxWidth to the columns

### LINKED ISSUE #2313: Datagrid: Request a mechanism to constrain a column width
Describe the feature:
In Landmark, the column widths are not set.  We want the columns width to adjust to the data being loaded into the grid. However if the datagrid contains a rich text column, there is no way to constrain the column width.  So if the rich text column contains a large amount of data the column width becomes huge.  

Here's our issue....  
A list contains a column with a cardview and a column with a text area.  They want the text area column to wrap, showing only the number of lines of data as the cardview.   So if the cardview is two lines, they want the text area to wrap to the two lines.  If more text exists than can be displayed in two lines, then an ellipsis is shown with a tooltip showing the complete value.  Today, the text area column expands to display the entire value making the column width HUGE when it contains a large amount of data.  I tried setting column.width = 200, but if this column is the last column it doesn't stretch to fill the remaining space.  And column.maxWidth is not available.

Provide something like column.maxWidth where we can suggest the largest width size of a column.

Tried using column.width, but if this column is the last visible column in the list it doesn't stretch to fill the remaining space.

Additional requirement:
Need to support showing up to x number of characters in a cell, the full value would be shown via a tooltip.

Landmark LPL for this is:
show up to x characters

Additional requirement:
Need to support showing up to x number of lines of data in a cell, the full value would be shown via a tooltip.

Landmark LPL for this is:
use text area of 3 lines

marknolaninfor:
Request a mechanism to constrain a column width, as this is useful for textarea columns to allow for some wrapping.

Closes #2313 

1. Goto 'http://localhost:4000/components/datagrid/test-columns-min-max-widths.html'
The Description column has a min and max constraint
2. Edit any description so that it is wider than 550 px, the maximum constraint value.
3. Page to next page 
4. Page to next page 
5. Change the edited description to a shorter description

The datagrid has stretchColumnOnChange enabled so the resize can happens on the change. (The column will only shrink on paging)

### DIFF
--- a/app/views/components/datagrid/test-columns-min-max-widths.html
+++ b/app/views/components/datagrid/test-columns-min-max-widths.html
@@ -1,0 +1,50 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+
+      var grid,
+        columns = [],
+        data = [];
+
+      // Some Sample Data
+      data.push({ id: 1, productId: 2142201, productName: 'Compressor', activity:  '', quantity: 1, price: 210.99, status: 'OK', orderDate:  '', portable: false, action: 1, description: ''});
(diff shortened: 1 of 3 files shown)


janahintal:
Hi @EdwardCoyle , noticed that when trying to erase the value of the description, the height changes and when trying to highlight text that I will delete, it is not being deleted. Please see below image
image
image

PR #2381, opened 2019-06-14
2026 - Modal: Tabbing initial focus

### LINKED ISSUE #2026: Accessibility:  Need a tabbable title in modal dialog
**Describe the bug**
Keyboard focus should move to the title when the modal dialog is loaded so the title is read.

**To Reproduce In Landmark**
1. Go to https://hcm-tampm01-prd.inforcloudsuite.com/hcm/EmployeeSelfService/form/Employee%287004,700417%29.LRCFullEmployeeProfile?csk.showusingxi=true&csk.JobBoard=INTERNAL&csk.HROrganization=7004&menu=LRCEmployeeMenu.MyProfile
2. Click on any magnifying glass icon to launch a modal dialog
3. Tab until visual focus is in the modal dialog 
4. Visual focus only appears on toolbar in modal dialog after user tabs through calling page

**To Fix In EP**
1. http://master-enterprise.demo.design.infor.com/components/modal/example-index.html
2. Ensure the title is read by the screen reader. Possibly this means making it the thing that gets focus

**Expected behavior**
When user tabs, visual focus should start on the toolbar in the modal dialog

**See also**
https://jira.lawson.com/browse/LMCLIENT-24204

davidcarlsonberg:
This new test page serves to demonstrate that the first focused item in a modal with a toolbar and a datagrid is the first non-disabled toolbar button.

Resolves #2026.

- Pull branch, run app
- Visit http://localhost:4000/components/modal/test-toolbar.html
  - Click "Open Modal"
  - Observe the first focused and tabbable item is the first non-disabled button in the toolbar.

¿- [ ] An e2e or functional test for the bug or feature.?
¿- [ ] A note to the change log.?

### DIFF
--- a/app/views/components/modal/test-toolbar.html
+++ b/app/views/components/modal/test-toolbar.html
@@ -12,7 +12,7 @@
         </div>
 
         <div class="buttonset">
-          <button type="button" class="btn-tertiary">
+          <button type="button" class="btn-tertiary" disabled>
             <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
               <use xlink:href="#icon-open-folder"></use>
             </svg>
--- a/app/views/components/modal/test-toolbar.html
+++ b/app/views/components/modal/test-toolbar.html
@@ -1,0 +1,145 @@
+<div class="row">
+  <div class="twelve columns">
+    <button class="btn-secondary" type="button" id="open-modal">Open Modal</button><br><br>
+
+    <!-- "Context" Example -->
+    <div id="modal-content" style="display: none;">
+
+      <div id="common-toolbar" class="toolbar" role="toolbar">
+
+        <div class="title">
+          Toolbar Title
+        </div>
+
+        <div class="buttonset">
+          <button type="button" class="btn-tertiary">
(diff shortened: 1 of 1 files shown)


EdwardCoyle:
I'm going to approve since the test demos the case, but I think there's another focus trap surrounding Datagrid inside of Modals we should address:

- Open http://localhost:4000/components/modal/test-toolbar.html
- Click "Open Modal"
- Press `Tab` key once to move focus to the grid
- Press `Shift + Tab` to try and move focus back to the button.  See that it doesn't move focus out of the grid.
- The same occurs when tabbing all the way through all grid cells.  It's not possible to then press `Tab` to get out of the grid and into the Pager bar, or beyond.

Not sure if it's related to this or not... what do you think @tmcconechy ?

tmcconechy:
Yes that is an issue @EdwardCoyle but i see it not only on this branch. It would be good to make a new issue for later. Note that the mentioned issue doesnt occur on http://localhost:4000/components/datagrid/example-index.html

EdwardCoyle:
Yeah, seems to be specific to the tabbing in Modals

PR #2387, opened 2019-06-17
2025 - Images: Make tabbable

### LINKED ISSUE #2025: Accessibility - Need a Tabbable Profile Image
**Describe the bug**
The profile image on the At A Glance page should be focusable, so to do this we need a tabbable image class in EP

**To Reproduce**
Steps to reproduce the behavior in Landmark
1. In landmark go to https://hcm-tampm01-prd.inforcloudsuite.com/hcm/EmployeeSelfService/form/Employee%287004,700417%29.LRCFullEmployeeProfile?csk.showusingxi=true&csk.JobBoard=INTERNAL&csk.HROrganization=7004&menu=LRCEmployeeMenu.MyProfile 
2. Tab once and visual focus is on the At A Glance menu on the left. Tab again and the visual focus will move to the email address under the Seth Burr profile image. 
3. Visual focus never goes to the profile image.

In EP This translates to:
1. On http://master-enterprise.demo.design.infor.com/components/images/list
2. Make all the images focusable and with a focus state
3. Show alt text in the example (and hope the teams implement)

**Expected behavior**
Visual focus should move to the image after user tabs and display a visual indicator

**See also**
https://jira.lawson.com/browse/LMCLIENT-24216

davidcarlsonberg:
Images were not tabbable. Focus state was off in Firefox and IE.

Closes #2025.

- Pull branch, run app
- Visit the below pages in Chrome, Safari, Firefox, and IE. Tab through and ensure proper focus state materializes.
  - http://localhost:4000/components/images


tmcconechy:
Is it just me? But im not seeing the focus styled as expected on chrome mac. Im seeing the default one as if the rules are not applying.

Screen Shot 2019-06-17 at 1 52 29 PM

On all pages http://localhost:4000/components/images I did restart a bunch of times

davidcarlsonberg:
@tmcconoughy What should be the correct visual focus state?

tmcconechy:
Should be the same as http://localhost:4000/components/input/example-index.html 

Screen Shot 2019-06-17 at 2 34 42 PM

But you were using the mixin. So not sure why. Seems like its not applying the rule

```
  border: solid 1px #368ac0;
  box-shadow: 0 0 4px 3px rgba(54, 138, 192, 0.3);
  outline: none;
  outline-color: transparent;
```

src/components/images/_images.scss
```diff
@@ -56,6 +56,31 @@ $images-size: (
   }
 }
 
+// Override border and focus state for Firefox and IE
+html.is-firefox,
```
tmcconechy: Why do we do this just for IE and Firefox. I think this is why i still dont see it on Mac Chrome.
I'd think these rules should apply on all browsers.

davidcarlsonberg: I interpreted this comment and those that follow to mean we shouldn't have a custom focus state unless it is necessary:
https://github.com/infor-design/enterprise/issues/2025#issuecomment-492286486

tmcconechy:
Hmm ok @EdwardCoyle but now i see on this comment https://github.com/infor-design/enterprise/issues/2025#issuecomment-492290250 he said it was fine. I dont see the point of only fixing this in ff and IE.

So i think lets maybe make it work everywhere but they can opt-in by adding the tabindex if they want it to be focusable. So both situations would be possible

janahintal:
Hi @tmcconechy, decimal is still not working in Windows10 Chrome browser for https://master-enterprise.demo.design.infor.com/components/datagrid/test-save-settings-serverside.html

PR #2388, opened 2019-06-17
1606 - add `phraseStartsWith` filterMode to ListFilter, deprecate `startsWith` filterMode

### LINKED ISSUE #384: SOHO-8084: ListFilter/Autocomplete API Test for `null` filterMode
Related to changes made to https://jira.infor.com/browse/SOHO-8070.

Need a test for ListFilter APIs and Autocomplete components that demonstrate unfiltered lists being returned when the filterMode setting is set to {{null}}.  This is the original problem that was causing the visual test in SOHO-8070 not to work.

**What should be tested**
- Pass a list of items to a ListFilter API instance, configured with {{filterMode: null}}
- When the filtering is completed, the full list should be returned with no items removed

The actual bug was in ListFilter, and was fixed in #125 for version 4.7.0.  We can't add the test in master until it's merged back, so this ticket is mostly a reminder to add the test when the time comes.

### LINKED ISSUE #1606: Dropdown: 'startsWith' filter works by word rather than start of a String
**Describe the bug**
The 'startsWith' filter currently works on words of a String, rather than start of a String (see below for example). Currently, there is no way to filter on start of a String.

Example list:
list

If I enter 'C' I expect 'Correction' to be selected. This works
filtered-list-1

However, if I enter 'A' I expect ''Allocated' to be selected but instead 'Not Allocated' is selected. It seems that the 'startsWith' works on word splits.
filtered-list-2

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/dropdown/test-filter-types.html
2. Enter 'S' into 'StartsWith Filter' dropdown.
3. 'American Samoa' selected by default

**Expected behavior**
Expected 'South Carolina' to be selected by default

**Version**
Found in 4.14.0, exists in 4.15.0

**Platform**
 - Device: PC
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version:  Version 72.0.3626.81 (Official Build) (64-bit)

**Additional context**
Discussion had with @tmcconechy and thought process was to change existing 'startsWith' behavior to filter based on start of String, and to add a new filter option such as 'startsWithWord' to filter based on the current behavior of 'startsWith'.

EdwardCoyle:
This PR:
- Adds a new filter mode to the ListFilter component called `phraseStartsWith` for matching search terms only at the beginning of entire strings.
- Deprecates the original `startsWith` filter in favor of the `wordStartsWith` filter, which has a more distinguishable name.  When defining the original one as a setting, there is now a deprecation warning.
- Adds plenty of ListFilter tests for the various filter types (including one that was missing for a `null` filterMode).

- Closes #1606
- Closes #384 (adds a missing test)

- Pull this branch, build, run the app, and run the functional test suite.  All the new tests should pass.
- Any example page such as [this one](http://localhost:4000/components/autocomplete/example-index.html) that implements ListFilter with its default settings should filter using `wordStartsWith`, and not display deprecation warnings in the console.

**Included in this PR:**

----
paging @rvaja-infor


src/components/listfilter/listfilter.js
```diff
@@ -1,12 +1,49 @@
...
+  }
+
+  // Partial dual word match
+  if (term.indexOf(' ') > 0 && text.indexOf(term) > 0) {
+    return true;
+  }
+
+  return false;
+}
+
```
tmcconechy: Extra line here (lint build error)

test/components/listfilter/listfilter-api-string.func-spec.js
```diff
@@ -8,6 +8,24 @@ const caseSensitiveData = [
   'CaLiFoRnIa'
 ];
 
+const phraseData = [
```
tmcconechy: Nice test!

PR #2392, opened 2019-06-18
2390 - Fix source not firing on sort and number sorting in the demo demoapp

### LINKED ISSUE #2390: Datagrid: Incorrect sorting on numeric columns
**Description of problem:**
On the example page http://localhost:4000/components/datagrid/test-calculate-text-width.html, when changing the sort on the "capacity" column, the numbers become incorrectly sorted (e.g., putting 10 and 100 just before 1,000):

capacity-sort

_Originally posted by @davidcarlsonberg in https://github.com/infor-design/enterprise/pull/2342#issuecomment-502748330_

**Expected Behavior:**
This column should sort numerically.

**Version**
Discovered testing `master` at 4.20.x

tmcconechy:

#2390 noted the sort on the number columns was wrong. When fixing found a regression from 4.16 where the source function did not fire. Fixed both. Now the example will show paging/sorting working on serverside data.

Fixes #2390 

- go to http://localhost:4000/components/datagrid/example-paging
- click twice on the id column header to sort reverse
- should show 999 - 990
- can also test the same on http://localhost:4000/components/datagrid/test-calculate-text-width.html on the capacity column

**Other Details**


tmcconechy:
@deep7102 fixed the sorting on time/date it was a demo app issue

EdwardCoyle:
Just something I noticed... the sorting is definitely working, but it looks like it gets incorrect data for a split-second, then corrects.  Watch the ID column in this GIF to really see it:

Jun-19-2019-10_03-AM

It's definitely not making a duplicate AJAX call, but it does run the `resultsText()` callback twice.

tmcconechy:
@EdwardCoyle could catch thats very subtle. I fixed this. The issue is its calling render twice once for whats there in the grid and then again when the callback comes back. So i made it render just once for the callback situation

tmcconechy:
@davidcarlsonberg I fixed the sorting by adding a false/true value every second row

davidcarlsonberg:
On sort, all checkboxes are returning the initial state.

PR #2407, opened 2019-06-19
2115 - Hide the header hamburger on large-form-factor Builder-pattern screens

### LINKED ISSUE #2115: [builder pattern] purpose of the hamburger icon on larger devices.
The following example layout shows a hamburger icon, when page is not on an extra small device, as then it changes to the back button icon and it is then clear on what it is used for.
http://master-enterprise.demo.design.infor.com/patterns/builder-subtitle-test.html

Should the example hide the icon on larger devices?
eg. by adding the following classes 
visible-sm hidden-md hidden-lg hidden-xl

**Additional context**
Example of this were the left hamburger loads the application menu.
image
We have used the styles locally, so the icon with no function is not displayed to larger devices.
image

EdwardCoyle:
This PR Fixes a demoapp issue where a Header Hamburger button is present on the "detail" area of the Builder pattern while on larger form factors.  This button doesn't have any functionality in this pattern while on the "desktop" view.  It should only be present when the viewport lands on the "detail" area while in "tablet"/"phone" view, where it acts as a "Go Back" button.

Additionally, this PR restores the navigation ability that used to be present when clicking an item on the "list" side while looking at the "phone"/"tablet" views.

Closes #2115 

- Pull this branch, run app
- Look at all the `builder` samples in http://localhost:4000/patterns/.
- Ensure that the desktop view on all these components does not display a Header Hamburger button.
- Ensure that on the mobile view, after clicking an item in the list, the detail area shows, and a "Go Back" button with an arrow is present.
- Clicking that button should return the user to the "list" view.

### DIFF
--- a/app/views/patterns/builder-nosidebar.html
+++ b/app/views/patterns/builder-nosidebar.html
@@ -14,15 +14,14 @@
           <div class="buttonset">
             <input class="searchfield" placeholder="Search">
           </div>
-
         </div>
       </header>
 
       <div class="listview scrollable" id="task-listview" data-options="{'template': 'task-tmpl', 'source': '{{basepath}}api/inventory-tasks'}"></div>
     </div>
 
-    <div class="main no-scroll builder-pane" role="main">
-      <header class="header  is-personalizable">
+    <div id="maincontent" class="main no-scroll builder-pane" role="main">
+      <header class="header is-personalizable">
         <div class="toolbar">
           <div class="title">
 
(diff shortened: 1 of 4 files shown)


EdwardCoyle:
@tmcconechy:
- Fixed the error on http://localhost:4000/patterns/builder-nosidebar.html
- Punted the Searchfield issue to #2431 because messing with the Searchfield calculation will need quite a bit of testing.

PR #2409, opened 2019-06-20
2408 - Datagrid: placeholder on additional Formatters

### LINKED ISSUE #2408: Datagrid: placeholder on additional Formatters.
**Describe the bug**
A placeholder property can be placed on a datagrid column, but it is only visible when the Placeholder formatter is in use. We have a requirement to display a placeholder on a field that has a Lookup Editor, so it also requires that the Formatter is set to Lookup to trigger the control. 

Other Formatters that may require a placeholder:
* Decimal
* DropDown

**To Reproduce**
'http://localhost:4000/components/datagrid/example-placeholder.html' needs to be updated to provide an example.

**Expected behavior**
A placeholder can be added to a column without the requirement to use the Placeholder formatter.

**Version**
 v4.20.0 (master)

**Screenshots**

marknolaninfor:
Able to use the placeholder on additional Formatters:
* DropDown
* Lookup
* Decimal

Closes #2408 

1. goto 'http://localhost:4000/components/datagrid/example-placeholder.html'
2. A placeholder is visible on the additional Formatters

### DIFF
--- a/src/components/datagrid/datagrid.formatters.js
+++ b/src/components/datagrid/datagrid.formatters.js
@@ -7,19 +7,18 @@ import { xssUtils } from '../../utils/xss';
 * @private
 */
 function calculatePlaceholder(placeholder, formattedValue, row, cell, value, col, item) {
-    if (placeholder && formattedValue === '') {
-      const getType = {};
-      if (getType.toString.call(placeholder) === '[object Function]') {
-        placeholder = placeholder(row, cell, value, col, item);
-      } else if (item && placeholder in item) {
-        placeholder = item[placeholder];
-      }
-      
-      return placeholder;
-    } 
-    else {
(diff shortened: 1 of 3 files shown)


src/components/datagrid/datagrid.formatters.js
```diff
@@ -515,8 +555,20 @@ const formatters = {
         }
       }
     }
+
```
tmcconechy: These code blocks repeat a bit, i wonder if its possible to extract these three into a shared function?

marknolaninfor: Moved to a function

tmcconechy:
Minor lint issue https://travis-ci.com/infor-design/enterprise/jobs/209866782#L525

PR #2413, opened 2019-06-20
2357 - Fix a treegrid bug so that the is-editable callback has correct rowdata

### LINKED ISSUE #2357: Datagrid / Treegrid: isEditable function may or may not send the correct row data
**Describe the bug**
The treegrid uses a callback function called isEditable to determine if a particular cell is editable. The parameters to the callback function includes row data. The row data may or may not be available since settings.dataset is 'collapsed' data.  When the tree nodes are expanded, the cell editable does not have the row data to make that determination.

**To Reproduce**
1. Use the demo page test-tree-editable.html.
2. Say for description column definition, set isEditable function:
```
columns.push({ id: 'desc', name: 'Description', field: 'desc', editor: Editors.Input, isEditable: cellIsEditable });
...
...
cellIsEditable = function(row, cell, value, columnDef, rowData, dataGrid) {
    return rowData != undefined;
};
```
**Expected behavior**
After expanding one of the tree node,  description cell of the last row has rowData passed in cellIsEditable function.

**Version**
ids-enterprise:  v4.19.0-beta.2

**Platform**
 - OS Version: Windows 10
 - Browser Name: chrome version 75

tmcconechy:

When editing cells the isEditable function can be used to dynamically determine if cells are editable. When using this on a tree grid the row data parameter did not have the correct data. This PR fixes that issue.

Fixes #2357 

- go to http://localhost:4000/components/datagrid/test-tree-editable
- click into the last (description) column
- a toast should be shown with the id that matches the row
- go to http://localhost:4000/components/datagrid/test-post-renderer-tree.html
- check the console. the postcell renderer should print each row out showing the rowdata works there
- go to http://localhost:4000/components/datagrid/test-tree-paging-serverside.html
- click into the last (description) column
- a toast should be shown with the id that matches the row
- go to page two or any page
- click into the last (description) column
- a toast should be shown with the id that matches the row


tmcconechy:
@deep7102 pushed a fix for that

tmcconechy:
@EdwardCoyle Pushed an update for what you noticed. Good catch!

EdwardCoyle:
@deep7102 this one appears to be good now!

tmcconechy:
@deep7102 and all. I have updated this PR as deepak noted. i would retest this now.

CindyMercadoReyes:
Moving this ticket to Failed QA.  On WIn 10 IE11 the data grid is not appearing. 
- Win 10 IE 11
- URL: http://master-enterprise.demo.design.infor.com/components/datagrid/test-post-renderer-tree.html
- The data grid doesn't load.  Please see screen shot attached. 
bs_win10_IE_11 0 (6)

PR #2429, opened 2019-06-24
2428 - Improve a fix for re-measuring datagrid columns

### LINKED ISSUE #2428: Datagrid: column widths no longer responsive to the data
**Describe the bug**
The datagrid used to automatically fit the data within the columns.

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/components/datagrid/test-paging-select-indeterminate-single.html'
2. Note the width and padding on the Product Name field 
3. Page to the end
4. There now is less padding on the right of the field and the value is close to the edge, if there was another value then the complete value would not be viewable.

**Expected behavior**
The data is displayed in the column and viewable

**Version**
4.20

**Screenshots**

**Additional context**
#2404 removed the lines that would recalculate the column widths based on the current data.
image

We have forms that will lazy load data into the datagrid using the loadData function 

image
The widths are not recalculated when the data is loaded.
image

tmcconechy:
When doing https://github.com/infor-design/enterprise/pull/2404 i fixed a problem but it introduced a behavior change thats unindented. This fixes.

Fixes #2428  

- retest #2404 
- But specifically go to http://localhost:4000/components/datagrid/test-editor-dropdown-source
- open the header multiselect and pick some items -> the columns should not change size when selecting
- go to http://localhost:4000/components/datagrid/test-paging-select-indeterminate-single.html
- note the width of the product name column
- go to last page -> the  product name column should adjust in size to the text


marknolaninfor:
@tmcconechy 

I don’t believe this will fix the issue if the data is sorted. Eg decending so the longer valued are loaded first or if there was a filter

PR #2448, opened 2019-06-27
2447 - Bar: Unable to use the {{value}} expression in the tooltip

### LINKED ISSUE #2447: Bar: Unable to use the {{value}} expression in the tooltip
**Describe the bug**
Other chart types allow for the tooltip to be defined as a function, or contain special fields to construct the tooltip. The Bar chart type is missing this and a good comparison is the column chart.

**To Reproduce**
To create

**Expected behavior**
All chart types are able to define the tooltip in the same ways making it easier to change between some chart types.

**Version**
4.20

**Screenshots**
image

**Additional context**
http://localhost:4000/components/bar/test-custom-tooltips.html will be update to have the same tooltip as http://localhost:4000/components/column/example-index.html

marknolaninfor:
Add the same tooltip dynamic functionality to the bar chart type

Closes #2447

Goto 'http://localhost:4000/components/bar/test-custom-tooltips.html'
Hover on the first item, as the custom tip will show the value.

### DIFF
--- a/app/views/components/bar/test-custom-tooltips.html
+++ b/app/views/components/bar/test-custom-tooltips.html
@@ -12,6 +12,8 @@
   </div>
 </div>
 
+{{={{{ }}}=}}
+
 <script>
 $('body').on('initialized', function () {
 
@@ -20,7 +22,7 @@ $('body').on('initialized', function () {
           name: 'Category A',
           value: 373,
           url: 'test',
-          tooltip: 'Custom Tooltip One'
+          tooltip: 'Custom Tooltip One - {{value}}'
       }, {
           name: 'Category B',
           value: 372,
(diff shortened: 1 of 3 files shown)


marknolaninfor:
@tmcconechy 

I will look at reviewing all the charts and move this formatting of the tooltip into charts so it is a shared function. This will Ben done next week

tmcconechy:
Go even better idea 👍

marknolaninfor:
Hi @tmcconechy 

I have looked at the code between three of the chart types, bar, column and pie, with them all being incosistant 
* pie just has a simple text subsitution, includes the additional percentage. (only a few lines as its simple)
* column would allow for a function or and subsitute on any field value in the data set.
* bar had no logic what you is defined is displayed.

I initally looked at the column as the bar chart is basically a different orientation of a column, and copied that logic.

Making a change to centeralise location to generate the tooltop content is the correct way to go, but it is risky because of the inconsistance between charts.

Can this be done in another ticket(s), so that initially all current ways to define and format a tooltip can be applied to a test page with a unit test, and then make the change to ensure there is no regression via the unit test results.

tmcconechy:
Sure @marknolaninfor we can definitely do this in a separate ticket. I was aware that the charts do have some inconsistencies including that one.

PR #2449, opened 2019-06-27
2444 - Fix an issue where filtering will only run every second time [v4.19.x]

### LINKED ISSUE #2444: Datagrid: Filter type text have to press enter twice to filter
**Describe the bug**
A datagrid with a set up of disableClientFilter: true and filterWhenTyping: false, will have the user have to hit enter twice.

**To Reproduce**
1. Unzip [test-editor-dropdown-source.zip](https://github.com/infor-design/enterprise/files/3331443/test-editor-dropdown-source.zip) to app/views/components/datagrid
2. Navigate to /components/datagrid/test-editor-dropdown-source
3. Enter some string into the filter for the Product Name Column
4. Hit Enter once, see that nothing happens
5. Hit Enter again, see that the table empties.

This seems to be related to the restoreFilterClientSide variable being incorrectly used at line [2199](https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L2199) in datagrid.js

**Expected behavior**
Would only need to hit the enter button once.

**Version**

- ids-enterprise: v4.19.3

**Platform**
 - All

tmcconechy:
It was found that the filter would only run every second time do to some bad flag logic. Found this code wasn't needed so removed to fix. While running tests I found there was a time-bomb bug in the timezone code so fixed this (when running at certain times (8:58 fx).

Fixes #2444 

http://localhost:4000/components/datagrid/test-timezone-formats?layout=nofrills&locale=nl-NL
- page should show timezones at 00:00

http://localhost:4000/components/datagrid/test-editor-dropdown-source
- in the text field type com then enter and should see a toast
- in the text field type "" then enter and should see a toast again (this wasnt working)

http://localhost:4000/components/datagrid/example-filter
http://localhost:4000/components/datagrid/example-keyword-search.html
- test for good measure

### DIFF
--- a/app/views/components/datagrid/example-keyword-search.html
+++ b/app/views/components/datagrid/example-keyword-search.html
@@ -42,7 +42,8 @@
         dataset: data,
         frozenColumns: {left: ['productId']},
         toolbar: {title: 'Data Grid Header Title', collapsibleFilter: true, results: true, keywordFilter: true, actions: true, rowHeight: true}
+      }).on('filtered', function () {
+        console.log('Filter Ran');
       });
-
  });
 </script>
--- a/app/views/components/datagrid/test-editor-dropdown-source.html
+++ b/app/views/components/datagrid/test-editor-dropdown-source.html
@@ -84,7 +84,7 @@
     };
 
     columns.push({ id: 'selectionCheckbox', sortable: false, resizable: false, formatter: Formatters.SelectionCheckbox, align: 'center'});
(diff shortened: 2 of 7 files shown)


tmcconechy:
@clepore No but the tests were failing in front of me. This was a problem that had been happening on the nightly builds and CI randomly and since it was finally possible to debug i fixed it. If you want i can push this to master instead but it should be low impact.

tmcconechy:
Yeah your right, I probably should have broke that one out separate - the timezone was actually not working well depending what time it is so nice to have fixed as I'm sure that would be discovered.

CindyMercadoReyes:
This issue is now resolved.

PR #2454, opened 2019-06-28
2351 - Fix header alignment with ellipsis setting

### LINKED ISSUE #2351: Datagrid: Aligned right sort indicators overlap the column header text
Describe the bug
With a column header alignment of right and enableTooltips setting of true, the sort indicators overlaps the column header text.

To Reproduce

1. Using the attached example-overlapping-sort-indicator.html
2. Mouse over the 'Order Date' column header
3. Notice that the sort indicators overlaps the column header text

Expected behavior
Column header text and sort indicators should not overlap

image

tmcconechy:

When using the option to set the header titles to show as ellipsis test with a tooltip some of the centered and right aligned column did not work correctly. This fixes that

Fixes #2351 

- open both  http://localhost:4000/components/datagrid/test-ellipsis-sort-indicator and http://localhost:4000/components/datagrid/test-overlapping-sort-indicator side by side so the headers can be seen.
- alignment of the text should be even
- on the test-ellipsis-sort-indicator example you can size the headers so the ellipsis is shown and hover to see the text in a tooltip
- while on test-overlapping-sort-indicator the text is cutting off on resize
- for good measure look at any alignment tests like http://localhost:4000/components/datagrid/test-align-header-text-toggle
- on all examples that have it available be sure to try all three modes (short, medium, normal)


EdwardCoyle:
@tmcconechy is this right for the content of the tooltip?

Screen Shot 2019-07-01 at 5 16 13 PM

tmcconechy:
@EdwardCoyle that was copied over from another example i just went ahead and removed it (hyperlinkTooltip setting). It is correct but not needed in this example

janahintal:
Failed QA. 
Filter are overlapping. Also for, http://master-enterprise.demo.design.infor.com/components/datagrid/test-overlapping-sort-indicator?theme=soho&variant=contrast&colors=133c59 & https://master-enterprise.demo.design.infor.com/components/datagrid/test-align-header-text-toggle the ellipsis is not showing upon resizing the column. 
image

PR #2455, opened 2019-07-01
2352 - RTL: Colorpicker menu arrows advance oppositely

### LINKED ISSUE #2352: RTL: using the arrows in colorpicker goes the wrong way
**Describe the bug**
I'm not sure that this is a problem, cause i'm not sure how the keyboards works for those using RTL. But it might be so I'll report it for now, maybe someone else knows?
The issue is that when I the colorpicker and try to select a color using the keyboard arrows it moves in the opposite direction (right when I click left arrow and left when I click the right)

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/colorpicker/example-index?locale=ar-SA
2. Click on the colorpicker to open it
3. Move around in the colorpicker using the left and right arrow

**Expected behavior**
I think that the selection should move in the direction of the arrow even in RTL

**Version**

- ids-enterprise: 4.18.1

**Screenshots**
-

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 74.0.3729.169 (Official Build) (64-bit)

**Additional context**
As I mentioned I am not sure if key codes are switched or something for RTL keyboards? but probably not so I think this is a bug. If not then please close this ticket :)

davidcarlsonberg:
When in RTL, the colorpicker left and right arrows advance opposite to their native direction. This PR fixes that issue by adding a RTL check and swapping the key events appropriately.

Closes #2352 .

- Pull branch, run app
- Visit http://localhost:4000/components/colorpicker/example-index.html & http://localhost:4000/components/colorpicker/example-index.html?locale=ar-EG (RTL)
  - Open colorpicker, focus into colorpicker menu, ensure left and right arrows advance correctly (left is left and right is right)

### DIFF
--- a/src/components/popupmenu/popupmenu.js
+++ b/src/components/popupmenu/popupmenu.js
@@ -857,7 +857,7 @@ PopupMenu.prototype = {
 
   handleKeys() {
     const self = this;
-    const isRTL = document.querySelector('html').dir === "rtl";
+    const isRTL = document.querySelector('html').dir === 'rtl';
     // http://access.aol.com/dhtml-style-guide-working-group/#popupmenu
 
     // Handle Events in Anchors
--- a/src/components/popupmenu/popupmenu.js
+++ b/src/components/popupmenu/popupmenu.js
@@ -857,6 +857,7 @@ PopupMenu.prototype = {
 
   handleKeys() {
     const self = this;
+    const isRTL = document.querySelector('html').dir === "rtl";
     // http://access.aol.com/dhtml-style-guide-working-group/#popupmenu
 
     // Handle Events in Anchors
@@ -1047,7 +1048,7 @@ PopupMenu.prototype = {
         }
 
         // Up on Up
-        if ((!isPicker && key === 38) || (isPicker && key === 37)) {
+        if ((!isPicker && key === 38) || (isPicker && key === (isRTL ? 39 : 37))) {
           e.stopPropagation();
           e.preventDefault();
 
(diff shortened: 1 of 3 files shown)


src/components/popupmenu/popupmenu.js
```diff
@@ -857,6 +857,7 @@ PopupMenu.prototype = {
 
   handleKeys() {
     const self = this;
+    const isRTL = document.querySelector('html').dir === 'rtl';
```
tmcconechy: Using `Locale.isRTL` would avoid a DOM selection. but safest way to do that is just using it in the event as it may not be set yet on init. So probably remove this line...

src/components/popupmenu/popupmenu.js
```diff
@@ -1047,7 +1048,7 @@ PopupMenu.prototype = {
         }
 
         // Up on Up
-        if ((!isPicker && key === 38) || (isPicker && key === 37)) {
+        if ((!isPicker && key === 38) || (isPicker && key === (isRTL ? 39 : 37))) {
```
tmcconechy: Use `Locale.isRTL()` here?

src/components/popupmenu/popupmenu.js
```diff
@@ -1085,7 +1086,9 @@ PopupMenu.prototype = {
         }
 
         // Down
-        if ((!isPicker && key === 40) || (isPicker && key === 39 && !isAutocomplete)) {
+        if ((!isPicker && key === 40)
+          || (isPicker && key === (isRTL ? 37 : 39))
```
tmcconechy: Use `Locale.isRTL()` here?

test/components/colorpicker/colorpicker.e2e-spec.js
```diff
@@ -410,3 +410,24 @@ describe('Colorpicker sizes tests', () => {
     });
   }
 });
+
+describe('Colorpicker RTL left and right keys test', () => {
```
tmcconechy: Do we have a non isRTL test? That would be nice. Seems like a good test

davidcarlsonberg: Yes, there's a non RTL one for keyboard selection in the e2e file.

Issue #2193, opened 2019-05-20
Compact Form: Implement Checkbox/Radios/Dropdown


EdwardCoyle:
**Is your feature request related to a problem? Please describe.**
This issue has the same goals as #1955, focusing on the implementation of:
- [x] Checkbox
- [x] Radios
- [ ] Dropdown/Multiselect
- [ ] Different visual states (error, warn, success, info, neutral) for input fields

**Additional Context**
Per @kayiuho's [comment on that issue](https://github.com/infor-design/enterprise/issues/1955#issuecomment-493488958), there is a [Dropbox Paper document](https://paper.dropbox.com/doc/Compact-Form--AdfI1qwaQD8ovJjtd44TMg9EAg-R8idPCwaQcbGBbIZVX3Md) containing designs that is currently in progress.

----

Branch: https://github.com/infor-design/enterprise/compare/2193-form-compact

EdwardCoyle:
Branch has been started here: https://github.com/infor-design/enterprise/compare/2193-form-compact

EdwardCoyle:
#2466 merged the changes for font size, spacing, and the additions of checks/radios.  Outstanding on this issue are the Dropdown/Multiselect, the visual states, and the dirty indicator.

tmcconechy:
Should this stay open?

tmcconechy:
@EdwardCoyle 
Should we close this and move to QA? Or keep it open for the remaining outstanding items? Maybe a new issue for those?

EdwardCoyle:
Might as well keep it open for now.

janahintal:
Hi. Tested this in http://master-enterprise.demo.design.infor.com/components/form-compact/example-index.html?locale=ar-EG & http://master-enterprise.demo.design.infor.com/components/form-compact/example-index.html no issued found so far for radio buttons and checkboxes. Will move to Done.

tmcconechy:
I think we will shelf this until further notice. Most is done and compact form isn't what it was

PR #2466, opened 2019-07-02
2193 - Add Checkbox and Radio field types to the Compact Form

EdwardCoyle:
This PR: 
- brings the style for the Compact Form into alignment with recent design iterations, as outlined in infor-design/design-system#370.  The styles shrink the input field fonts, change the header style, and add a gutter between nested grids (see [V5 Notes](https://paper.dropbox.com/doc/Compact-Form--AgGu0QJk83oONJo1DNMYEB_FAg-R8idPCwaQcbGBbIZVX3Md#:h2=V5:) in Dropbox Paper)
- Allows for the use of checkboxes/radios in the compact form with proper alignment/font sizes.

#2193

- Pull this branch, build, run app.
- Navigate to http://localhost:4000/components/form-compact/example-index.html and http://localhost:4000/components/form-compact/example-index.html?locale=ar-EG (Uplift also works, but isn't 100% tested until we get the fonts in)
- Inspect the checkbox and radio button groups fields to ensure they line up properly with other fields.  - When selecting/highlighting any radio or checkbox, their field containers should change to the blue background color and remain that way until the focus moves away from that group.
- Ensure that on the desktop breakpoint, there is a gutter in-between Section One and Section Two, and that Section Three underneath lines up vertically with Section One.

### DIFF
--- a/app/views/components/form-compact/example-index.html
+++ b/app/views/components/form-compact/example-index.html
@@ -127,71 +127,77 @@
       </div>
 
       <div class="row">
-        <div class="six columns form-section-header">
-          Section Three
-        </div>
-      </div>
+        <div class="six columns">
 
-      <div class="row">
-        <div class="one-one-half columns">
-          <label for="field-07">A Long Label Name</label>
-          <input id="field-07"/>
-        </div>
-        <div class="one-one-half columns">
-          <label for="field-08">Field 08</label>
-          <input id="field-08" value="Equipment Inc."/>
-        </div>
-        <div class="one-one-half columns">
-          <label for="field-09">Field 09</label>
-          <input id="field-09" value="Equipment Inc."/>
-        </div>
-        <div class="one-one-half columns">
-          <label for="field-10">0000-0000-00</label>
-          <input id="field-10" value="Text"/>
-        </div>
-      </div>
+          <div class="row">
+            <div class="twelve columns form-section-header">
+              Section Three
+            </div>
+          </div>
 
-      <div class="row">
-        <div class="three columns">
-          <label for="field-11">Field 11</label>
-          <input id="field-11" value="Text"/>
-        </div>
-        <div class="three columns">
-          <label for="field-12">Field 12</label>
-          <input id="field-12" value="Text"/>
-        </div>
-      </div>
+          <div class="row">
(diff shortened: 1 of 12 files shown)


tmcconechy:
@nickwynja that is actually standard browser behavior. Radios you can arrow checkboxes you tab, so we would have to fight with that...

EdwardCoyle:
@nickwynja it's true that they're different in the sense that a group of radios represents multiple choices for the same "option" (fx: "x" could be 1, 2, or 3), where as a set of checkboxes could each toggle individual options (fx: "x" = true, "y" = false, "z" = false).

There's an argument to be made that we could support a group of related checkboxes that act in a similar manner to the radios, but at that point you could also argue that the right UI choice should be radios.

EdwardCoyle:
@tmcconechy, question for you?

1. Not sure I understand this one.  There should be about an 8px border between any label/input text and the field boundaries.  Is there one/multiple specifically you're seeing without padding?
2. @nickwynja, Kayiu and I tried to figure this out, but it's trickier than just moving the markup around.  Section Two has to logically come before section three in terms of tab order.  The way the markup sits now, even though there's a large gap underneath section one, causes the fields to flow in the correct order, and sit correctly in the responsive view.  We settled on keeping it as-is.  See the screenshot below:
Screen Shot 2019-07-09 at 11 47 32 AM
3. Good catch, there should be borders on these groups!

tmcconechy:
@EdwardCoyle 
1.  i mean the bottom padding on the whole thing is less than the top, left and right (at larger desktop where its 20px). Maybe a non issue because it might continue down.
2. Ok, yeah i figured or would be more like homepages with the absolute positioned containers. (Could be one option)
3. ok cool

EdwardCoyle:
Not much I can do about that one, other than change the text content.  RTL with English copy seems to have that effect with punctuation.

EdwardCoyle:
@tmcconechy I re-balanced the fields in this example a bit to lessen the space.  I had originally laid the fields out according to design document for this pattern, but I think your right that we should just show better layouts.  I also fixed the icon alignment.

@davidcarlsonberg I changed the copy on those inputs to have no punctuation so we don't see any RTL weirdness.

PR #2487, opened 2019-07-15
1876 - Add drag drop and save position for toast

### LINKED ISSUE #1876: Toast: Add configuration to move the toast and have it stay in that postion

Recruiter's at Riverside using Landmark/GHR's ATS software are complaining that the confirmation message is "in the way".   This is due to placement and timing I'm guessing.

Placement: confirmation is typically placed right over the toolbar (top right).  Makes sense, it's were most eyes go as well.  The problem is that it does block what actions I might want to take next.  This happens to power-users who hammer on data and records all day.

Timing: the message isn't snappy one would say.  I know there is an "x" to dismiss it, but that means extra clicks and things to deal with.  The feedback is that doing the "x" is just in the way.

Potential ideas might be to add personalization settings on placement and timing.  E.g. I'm a fast reader, I know the system, i'd cut the message time by 50%.   Or, I'd like confirmations on the bottom left.  Like the browser did/does.

- [ ] Add the ability to drag the toast to any place on the page you want
- [ ] Once it is dragged save the position in local storage
- [ ] Next toast for the user will go in the same place -> or under the last one if stacked

deep7102:
Add api settings for drag drop and save position with Toast.

Closes #1876

- Navigate to http://localhost:4000/components/toast/example-draggable.html
- Click on button `Show Toast Message` to open toast
- Mouse over the toast
- See drag handle (&#8286;) four vertical dots
- Click on drag handle and drag to some other position
- Click `X` in toast or wait for time out to close the toast
- Close the browser window and navigate back to same page
- Click on button `Show Toast Message` to open toast
- See toast should open at previously saved position

- [x] An e2e and functional test for the feature.


tmcconechy:
This is working well. The only small issue i have is with the drag handle placement. It seems a little tight with the text. 
Screen Shot 2019-07-15 at 4 13 43 PM

Maybe we can make the whole thing draggable with just the cursor (except the x) and no handle?

EdwardCoyle:
I'm not sure what the challenges are with the handle but I tend to agree, if it's possible to make the entire thing draggable, that would be better than the handle.

@deep7102 it looks great other than that!  I found a little usability bug that I'm not sure is worth fixing in this iteration, but just to identify it:

- Make a toast and drag it a bit
- Wait about 3 seconds, then create a couple more toasts
- Click and hold the handle of one of the lower toasts on the list.  Optionally move it around, but don't release the mouse button
- When the other toasts expire, they disappear and the one you're dragging moves up in the "list", but your actual mouse cursor stays where the handle was originally located.
- When this happens, it seems the timer for the toast that is being held doesn't re-activate.

See the gif below for a visual:
Jul-16-2019-10_08-AM

deep7102:
@tmcconechy @EdwardCoyle @davidcarlsonberg Updated now,
- Remove the drag indicator and make the whole thing draggable
- Resume the toast timer, when multiple toasts added

PR #2488, opened 2019-07-15
920 - Improve Multiselect filtered selection process

### LINKED ISSUE #920: Dropdown: Select within multiselect dropdown control doesn't let you make multiple selections
**Describe the bug**
In the multiselect control, when a user goes to enter a search query and views filtered options within the dropdown list, selecting an option clears the search query and removes all filtered search results. User has to perform another search query to make additional selections within filtered search results.

I understand that the input at the top of the dropdown used to display active list options selected is also used to perform search queries, but as a user I should be able to make multiple selections within a filtered options list without it clearing my search.

**To Reproduce**

Using the multiselect examples on the SoHo XI web site
https://design.infor.com/code/ids-enterprise/4.10.0/multiselect

1. For the first control in the examples (States), click into control and type "ad" within dropdown search
2. Options will filter to display only 2 choices: "Colorado" and "Nevada"
3. Select either option
4. Results of search are cleared after selection is made, user returns to top of dropdown list

**Expected behavior**
If a user is making selections within an active search, they should be allowed to make multiple selections without their search clearing. 

**Video**
See video: https://screencast.com/t/V4miZiWnzy9l

**Platform**
 All browsers and devices.

**Additional context**
I see how this can be considered a new feature, and it is performing as designed as a single-select control, but this is an issue with multiselect control that came up as part of recent UAT with a client. They are considering this a bad user experience, that you can search within a list and not make multiple selections within the filtering list results as you lose your original search query.

To solve, I've seen other controls include the search input separate from element/input that contains the active list options selected.

EdwardCoyle:
This PR makes a few changes to the way the Multiselect component handles selecting items from a filtered list:

- Selecting an item no longer resets the value of the Search Input
- Selecting an item no longer rebuilds the Multiselect's list

These changes were made to improve the usability of the Multiselect, keeping the state of the Search input in-tact between selections, and to prevent too much context switching.  

Although we'd originally talked about more drastic changes to the Multiselect's list, and potentially a different way to display selected items, I opted to keep the original design more in-tact while changing a bit of the usability.  In the new configuration, the list takes the role of showing what's currently selected instead of the search input.

This PR also fixed a minor unrelated bug where an AJAX request that populates a Dropdown/Multiselect containing items with no `label` attributes could result in every list item becoming `undefined`.

Closes #920 

_To tests the multiselect improvements:_
- Pull this branch, run app
- Open http://localhost:4000/components/multiselect/example-index.html
- Open the first Multiselect at the top
- Type "Co" to filter the list
- Select "Colorado".  The item should become selected, but the list and search input should not reset.
- Backspace until the list is completely empty.  The list should reset and "Colorado" should be selected along with the initially selected "Alaska"
- Press the Escape key.  When the list closes, the in-page pseudo-element should read "Alaska, Colorado".

_To test the `undefined` problem:_
- Open http://localhost:4000/components/multiselect/test-selected-rendering-styles.html
- Open any of the Multiselect components on the page.  None should show `undefined` for all the selectable options.

----

Also requesting we get a look from @picitelli and @fitzorama, who had comments on the original ticket.


src/components/dropdown/dropdown.js
```diff
@@ -2153,10 +2153,10 @@ Dropdown.prototype = {
     return true;
   },
 
-  /*
-  * Function that is used to chekc if the field is loading from an ajax call.
-  * @returns {void} Returns true if the field is attempting to load via AJAX.
-  */
+  /**
+   * Function that is used to chekc if the field is loading from an ajax call.
```
tmcconechy:
```suggestion
   * Function that is used to check if the field is loading from an ajax call.
```

PR #2499, opened 2019-07-16
2365 - Add Locale-specific Uplift fonts to the Demoapp

### LINKED ISSUE #2365: Demoapp: Add Uplift/RTL Font Support
The Demoapp needs to conditionally serve font families for certain locales when displaying the Uplift theme.

- Fonts should be served locally from the app, or pulled in from Google Fonts
- Font files served locally should be copied in from the `ids-identity` package
- Link tags that include the font should conditionally be used only if a corresponding locale requires the font.
- Don't bundle the fonts in the actual IDS build.  We are leaving the inclusion of the extra locale fonts up to the app developers to use if they are actually needed.

**Additional context**
Discussion on Locale-specific Uplift fonts can be seen on infor-design/design-system#385

EdwardCoyle:
This PR implements a middleware handler into the Demoapp that manages the Google Fonts pulled in when loading specific localizations against the Uplift theme.  The fonts were discussed on infor-design/design-system#385.

Closes #2365

Pull this branch, run the app.  Then, check the pages in the table below to ensure that the fonts being loaded are correct. This can be checked by:
- opening a dev tools console
- inspect the page
- open the `<head>` tag, and find the `<link>` tag containing the URL to `fonts.google.com`
- inside the link tag's `href` attribute, check to see that the name of the font, and its subset query string (if provided) match:

Screen Shot 2019-07-16 at 5 02 57 PM

| Test Page Locale | Font | Subset (if applicable) |
| ---- | ---- | ---- |
| [`en-US` (default)](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift) | Source Sans Pro | |
| [`ar-EG`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=ar-EG) | Mada | arabic |
| [`ar-SA`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=ar-SA) | Mada | arabic |
| [`he-IL`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=he-IL) | Assistant | hebrew |
| [`hi-IN`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=hi-IN) | Noto Sans | devanagari |
| [`ja-JP`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=ja-JP) | Noto Sans JP | japanese |
| [`ko-KR`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=ko-KR) | Noto Sans KR | korean |
| [`vi-VN`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=vi-VN) | Source Sans Pro | vietnamese |
| [`th-TH`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=th-TH) | Sarabun | thai |
| [`zh-CN`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=zh-CN) | Noto Sans SC | chinese-simplified |
| [`zh-tw`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=zh-tw) | Noto Sans TC | chinese-traditional |
| [`zh-Hans`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=zh-Hans) | Noto Sans SC | chinese-simplified |
| [`zh-Hant`](http://localhost:4000/components/locale/test-translated-strings.html?theme=uplift&locale=zh-Hant) | Noto Sans TC | chinese-traditional |

- [x] A test page for the feature.


kentonquatman:
Vietnamese can still use Source Sans Pro, but everything else looks good.

tmcconechy:
@kentonquatman BTW did you see a definitive list of what Source Sans supports in terms of languages? Not seeing a really good list out there..

kentonquatman:
@tmcconechy Checking on Google Fonts, it looks like Source Sans Pro supports the following:

- [Cyrillic](https://fonts.google.com/?query=Source&subset=cyrillic)
- [Cyrillic Extended](https://fonts.google.com/?query=Source&subset=cyrillic-ext)
- [Greek](https://fonts.google.com/?query=Source&subset=greek)
- [Greek Extended](https://fonts.google.com/?query=Source&subset=greek-ext)
- [Latin](https://fonts.google.com/?query=Source&subset=latin)
- [Latin Extended](https://fonts.google.com/?query=Source&subset=latin-ext)
- [Vietnamese](https://fonts.google.com/?query=Source&subset=vietnamese)

tmcconechy:
OK russian is cyrillic so we should be covered. Only Thai is missing from what i can see. (and vietnameses can use source sans)

EdwardCoyle:
@kentonquatman just pushed the change for Vietnamese back to Source Sans Pro.

@tmcconechy I didn't see our support for the Thai locale `th-TH` before now... didn't realize we had one.  [Source Sans Pro and Noto Sans don't have a variant for it](https://fonts.google.com/?subset=thai), so I guess we'll have to go with something else?  Or should we just fall back to Sans Serif for now?

tmcconechy:
@EdwardCoyle Yes i just cross checked with https://github.com/infor-design/enterprise/tree/master/src/components/locale/cultures and found it missing. I thought i saw a Noto for it but i see your right now. I guess we can pick any Sans Serif font on https://fonts.google.com/?subset=thai&category=Sans+Serif maybe kanit or prompt? Any thoughts on that one @kentonquatman ?

kentonquatman:
@EdwardCoyle Kanit might be a display typeface. It looks like Sarabun might be a better option for legibility, but we could test both of them.
https://fonts.google.com/specimen/Sarabun

nickwynja:
It may make sense to use [`unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) to specify which font gets used for which specific character set.

This is useful when there are mixed character sets, like Latin characters and Arabic characters on the same page (like in the example of these locale test pages). I'm not sure how often there are mixed character sets on application pages though.

EdwardCoyle:
@nickwynja I *think* the Google Fonts API is actually handling that for us:

Screen Shot 2019-07-17 at 11 16 26 AM

One thing I'm just noticing is that these `@font-face` rules are duplicated... a lot... I'm not sure why the Google Fonts API would serve it this way, but maybe it can be made more efficient.  The total payload for each font doesn't seem to be all that bad:

Screen Shot 2019-07-17 at 11 21 28 AM
Screen Shot 2019-07-17 at 11 21 50 AM

nickwynja:
@EdwardCoyle So then maybe we need to always include Source Sans in `<head>` so that it can be picked up and used for that range?

EdwardCoyle:
@tmcconechy per your comments:
> This example page is giving an error http://localhost:4000/components/locale/test-translated-strings.html?locale=ar-SA&theme=uplift (maybe also update the description to mention this test page)

I took a look and it seems to be some kind of race condition only with the `ar-SA` culture.  Will investigate a bit more.

> The example page will probably fail on IE 11 (and QA will fail this because of Object.keys)

I tested on Win7/IE11 and it seems like it works fine. `Object.keys()` is supported [back to IE9](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Browser_compatibility)

----

@nickwynja somehow I missed that a fresh page load doesn't load the Source Sans fallback :-/  I will add that permanently into the font string.

src/components/typography/_typography-uplift.scss
```diff
@@ -3,16 +3,18 @@
...
-  'zh-tw': ('Source Han Sans', 'Source Sans Pro', 'helvetica', 'arial'),
-  'zh-Hans': ('Source Han Sans', 'Source Sans Pro', 'helvetica', 'arial'),
-  'zh-Hant': ('Source Han Sans', 'Source Sans Pro', 'helvetica', 'arial')
+  'ar-EG': ('Mada', 'Source Sans Pro', 'arial', sans-serif),
+  'ar-SA': ('Mada', 'Source Sans Pro', 'arial', sans-serif),
+  'he-IL': ('Assistant', 'Source Sans Pro', 'arial', sans-serif),
+  'hi-IN': ('Noto Sans', 'Source Sans Pro', 'arial', sans-serif),
+  'ja-JP': ('Noto Sans JP', 'Source Sans Pro', 'arial', sans-serif),
+  'ko-KR': ('Noto Sans KR', 'Source Sans Pro', 'arial', sans-serif),
+  'vi-VN': ('Noto Sans', 'Source Sans Pro', 'arial', sans-serif),
```
tmcconechy: Maybe we dont need vi-VN as per this comment https://github.com/infor-design/enterprise/pull/2499#issuecomment-512082036 it can just be SourceSans Pro?

PR #2521, opened 2019-07-24
2446 - Improve scrolling capability of Homepage container

### LINKED ISSUE #2446: Homepages: homepage.js isn't returning a correct height.
**Describe the bug**
The homepage component element doesn't have a height set on it. 

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/homepage/example-hero-widget.html
2. In devtools highlight the `div.homepage` element.
3. make `overflow: auto` as an element style 
4. See incorrect height

**Expected behavior**
Want to see the height correctly set for the homepage element give it's home page tiles current layout.

**Version**
- ids-enterprise: v4.19

**Screenshots**
image

Notice when I set overflow to auto it shows the home page's true height.

**Platform**
 - OS Version: Windows 10
 - Browser Name chrome
 - Browser Version Version 75.0.3770.100 (Official Build) (64-bit)

**Additional context**

EdwardCoyle:
This PR adds a `height: inherit;` CSS rule to the Homepage container element.  This assists the visibility of the element when it's made scrollable, either by way of adding the `scrollable` CSS class IDS provides, or by adding `overflow` CSS rules manually. 

Closes #2446

- Pull this branch and run the app.
- Open http://localhost:4000/components/homepage/example-hero-widget.html and ensure that the Homepage widget area is still usable. Resize the browser viewport to various sizes and make sure the entire page still scrolls, and all widgets are visible.
- Open http://localhost:4000/components/homepage/example-hero-widget-scrollable-container.html.  Ensure that the Homepage widget area is scrollable on its own (outside of the hero image area), and that all Homepage widgets can be visible via scrolling.

### DIFF
--- a/app/views/components/homepage/example-hero-widget-scrollable-container.html
+++ b/app/views/components/homepage/example-hero-widget-scrollable-container.html
@@ -5,7 +5,7 @@
   <div class="page-container no-scroll">
     {{> includes/header}}
 
-    <div id="maincontent" class="page-container scrollable" role="main">
+    <div id="maincontent" class="page-container no-scroll" role="main">
 
       <!-- BEGIN: hero-widget -->
       <div class="hero-widget is-personalizable">
--- a/app/views/components/homepage/example-hero-widget-scrollable-container.html
+++ b/app/views/components/homepage/example-hero-widget-scrollable-container.html
@@ -1,0 +1,324 @@
+<body class="no-scroll">
+  <a href="#maincontent" class="skip-link" data-translate="text">SkipToMain</a>
+  {{> includes/svg-inline-refs}}
+
+  <div class="page-container no-scroll">
+    {{> includes/header}}
+
+    <div id="maincontent" class="page-container scrollable" role="main">
+
+      <!-- BEGIN: hero-widget -->
+      <div class="hero-widget is-personalizable">
+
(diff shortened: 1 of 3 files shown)


tmcconechy:
@pwpatton Would you please test this on your pages and make sure it works as expected for you?

tmcconechy:
@EdwardCoyle This is probably just an demo example issue but seeing a double scrollbar on http://localhost:4000/components/homepage/example-hero-widget-scrollable-container.html i think we just need to add `no-scroll` to the top `page-container no-scroll`. Seems to me like this will work as expected.

EdwardCoyle:
@tmcconechy just changed it

pwpatton:
This fix is not working for my landmark screens. Asking Ed if he could get on my machine and maybe catch what I'm doing wrong?

pwpatton:
On The double scroller just remove the scrollable from the homepage element

EdwardCoyle:
@pwpatton did some testing and confirmed that `height: inherit;` wasn't really enough for the Landmark cases to fix the height problem.  Since the Homepages layout is JS driven and has absolutely positioned children, we need to look at providing a calculated height for the `div.homepage` element whenever the Masonry layout resizes.

I'm closing this PR and pushing this ticket to 4.21.x

PR #2549, opened 2019-07-29
2489 - Calendar: Vertical scroll at small screen size [v4.20.x]

### LINKED ISSUE #2489: Calendar: Unable to scroll for smaller screens
**Describe the bug**
Calendar: Unable to scroll for smaller screens

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/calendar/example-index.html 
http://master-enterprise.demo.design.infor.com/components/calendar/test-specific-month.html#
2. lick at the date at the bottom of the calendar. There's no way to scroll again to the calendar header.

**Expected behavior**
Be able to scroll for smaller screen size 

**Version**
Beta 4.20

**Screenshots**
image

**Platform**
All OS & Browsers. 

**Additional context**
NA

davidcarlsonberg:
When shrinking a window vertically, the main calendar area was not scrolling correctly to view every day in the monthview. This PR forces the container to fit the content and thus allows for scrolling to view any part of the calendar. Also included a small fix to prevent cutoff of the Day Events section on the right. This has been tested on Chrome, Firefox, Safari, and IE11.

Closes #2489 

- Pull branch, build, run app
- Visit all monthview examples pages: http://localhost:4000/components/calendar
  - Ensure at small window height sizes, the three main panes are scrollable to whatever is populated inside them.
  - Test on major supported browsers as `fit-content` isn't technically fully supported according to https://caniuse.com/#search=fit-content but seems to work just fine.


src/components/monthview/_monthview.scss
```diff
@@ -4,7 +4,7 @@
 .monthview {
   background-color: $popover-bg-color;
   display: inline-block;
-  min-height: 300px;
+  min-height: fit-content;
```
tmcconechy: What sort of fallback will happen on IE edge and 11 here? I guess no min-height?

davidcarlsonberg:
Is it a requirement to have each container (calendar-events, calendar-monthview, calendar-event-details) scroll individually?

src/components/monthview/_monthview.scss
davidcarlsonberg: no min-height is correct

tmcconechy:
Yes sort of like try to go to http://localhost:4000/components/calendar/example-index.html
- scroll back with the previous button to January 2019 where there is one with a lot of event sin a day
- click it and you can see the selected / shown events scroll

So i think scrolling within the calendar (under the days) might work for this case well.
Screen Shot 2019-07-29 at 3 13 05 PM

src/components/calendar/_calendar.scss
```diff
@@ -308,15 +309,12 @@
   flex: 0 0 auto;
   height: 100%;
   margin-top: -1px;
+  min-height: 235px;
   overflow: auto;
   vertical-align: top;
   width: 221px;
 
   &.has-only-one {
-    .accordion-pane {
```
tmcconechy: I think we need this rule for the right calendar-event-details detail pane only and also change it to - 46px as there is a tiny gap

davidcarlsonberg: I see a 1px gap at the bottom don't see a gap at the bottom but it seems unrelated to this style block; it looks the same with or without that rule. The way I'm able to get rid of the tiny gap is to account for the `margin-top: -1px;` on the parent container by adding this to the `calendar-event-details` element: `height: calc(100% + 1px);`.

tmcconechy: Sorry this comment is related to point 1 not point 2 (as to why the event details pane doesnt extend down anymore).

Regarding the extra scrollbar i was looking at the -1px rule im not sure why thats there. If it looks ok with out that you can probably remove

davidcarlsonberg: Unfortunately, when you add in that height calculation and override the `height: auto;` inline style, then you can't scroll the event if it's a single event with a short window size.

davidcarlsonberg: I actually don't think the gray is that bad on the bottom of `calendar-events-details` because when you collapse multiple accordion items there it's also gray. Gray is just the background color to that pane. Makes sense to me...

tmcconechy: OK how but just adding a bottom border (same as side border) then for that?

PR #2571, opened 2019-07-31
2526 - Fix an infinite loop when calling updateColumns

### LINKED ISSUE #2526: Datagrid Filter: disableClientFilter and load data causing an infinite loop when updating columns
**Describe the bug**
Calling updateColumns on a datagrid will emit a filtered event when disableClientFilter is true for options. The filtered event is from the [restoreFilter ](https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L763)flag being set to true, which causes the [filtered event to be emitted](https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L1271).

Does a rerender of the header need to emit the filter event if it's disableClientSideFilter?

**To Reproduce**
1. Copy [test-filter-load-data-update-columns.zip](https://github.com/infor-design/enterprise/files/3432387/test-filter-load-data-update-columns.zip) into app/views/components/datagrid
2. Navigate to '/components/datagrid/test-filter-load-data-update-columns'
3. Attempt to set a filter on the Action column
4. See error - Page is unresponsive, and if you open the browser debugger, an infinite loop of filtered events are being printed in the console.

**Expected behavior**
An updateColumn call should not cause a filtered event to occur when disableClientSideFilter is set to true.

**Version**

- ids-enterprise: v4.20.0-dev

**Platform**
 - All

tmcconechy:

This fix solves an issue when calling `updateColumns` during a filter callback. Also testing rolling back chrome driver but may remove this.

Fixes #2526

- go to http://localhost:4000/components/datagrid/test-filter-load-data-update-columns.html
- select an item from the filter row multiselect
- watch the console and should just see one event called.
- retest: http://localhost:4000/components/datagrid/test-editor-dropdown-source (example should behave as it did before)


docs/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # What's New with Enterprise
 
+## v4.21.0
+
+### v4.21.0 Fixes
+
+- `[Datagrid]` Fixed a bug where calling update rows in the filter callback will cause a infinite loop. ([#2526](https://github.com/infor-design/enterprise/issues/2526))
```
davidcarlsonberg:
```suggestion
- `[Datagrid]` Fixed a bug where calling update rows in the filter callback will cause an infinite loop. ([#2526](https://github.com/infor-design/enterprise/issues/2526))
```

nickwynja:
Possible issue here, I'm not sure:

1. I selected rows using the multi-select column on the left of the grid.
2. Then used the dropdown on Action column to filter.
3. Multiple events are triggered.

This seems to only happen when first selecting a row through multi-select and happens whether the row item remains checked or not. This could be irrelevant but wanted to post in case.

tmcconechy:
@nickwynja Im not seeing any more then one filtered event when selecting a filter item. The multiselect /row selection is really unrelated here. The problem is the example is "serverside" so when doing that the rows are not persisting selection. We could update the example to do that but its not really the point of the issue.

PR #2608, opened 2019-08-05
2446 - Add metadata param to the Homepage `resize` event with container height

### LINKED ISSUE #2446: Homepages: homepage.js isn't returning a correct height.

EdwardCoyle:
This PR adds an extra parameter to the Homepage component's `resize` event, containing metadata about the current state of the Homepage component.  One of the properties of this parameter is the calculated height of the Homepage container.

The `containerHeight` property is calculating the height of the Homepage container internally based on blocks, the number of available columns and rows, and the gutters.

Closes #2446

- Pull this branch and run the app
- Navigate to http://localhost:4000/components/homepage/test-hero-widget-calculate-height.html
- Resize the browser viewport once
- Look in a dev tools console.  See that the `console.log` is correctly reporting an object after the `metaData:` string that contains a `containerHeight` property.

----
paging @pwpatton


EdwardCoyle:
@pwpatton can you check this change against your implementation and see if the number we are passing through the resize event works for setting container height?

@tmcconechy on http://localhost:4000/components/homepage/test-hero-widget-calculate-height.html, in some cases, there seems to be an extra widget's worth of height added to the total.  I'm wondering if [my calculation](https://github.com/infor-design/enterprise/pull/2608/files#diff-22d84c71524b1b9addb8568d1f0d468eR64) is missing a check somewhere?

EdwardCoyle:
@tmcconechy I pushed a change that accounts for a currently-empty row in the `rowsAndCols` matrix.  Added a test for it as well.

tmcconechy:
Yes looks good now to me. I think just need to fix those tests and this seems good. Adding @pwpatton for testing

pwpatton:
@tmcconechy Yes, this seems really good.

Issue #2605, opened 2019-08-05
Contextual Action Panel: Extra space on mobile when there is longer text.


tmcconechy:
**Describe the bug**
Using the CAP (workspaces example). If the text is longer there will be an overflow causing a white space on the right.

**To Reproduce**
1. Go to http://localhost:4000/components/contextualactionpanel/example-workspaces.html
2. Edit the title in the DOM so it says something longer like "Promote Pearl Reynolds Promote Pearl Reynolds"
3. Resize to mobile or use a mobile device

**Expected behavior**
No white gap and better text overflow

**Version**
- ids-enterprise: [e.g. v4.20.0]

**Screenshots**
![Screen Shot 2019-08-05 at 1 48 47 PM](https://user-images.githubusercontent.com/814283/62484234-c5e38100-b787-11e9-8494-2eea0dc29e80.png)

**Platform**
 - Device  [Desktop Responsize or IOS or Android newer devices]

**Additional context**
Infor Go Video/Slide show.

CindyMercadoReyes:
This issue is now resolved. Closing out this ticket.

PR #2609, opened 2019-08-06
2605 - Fix white space on the right side of workspace on mobile view

ericangeles:

On mobile, I truncated the title text from `481px (min)` to  `968px (max)` with a text width of `150px` and `250px` from `961px (min)`. So even the title has a very long text, it will be on ellipsis state regardless of the size of the browser.

Closes https://github.com/infor-design/enterprise/issues/2605

- Pull this branch and run the app
- Navigate to http://localhost:4000/components/contextualactionpanel/example-workspaces.html
- Resize the browser to mobile or use mobile device
- The title text should be truncated and white space in workspace should not be seen.


docs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # What's New with Enterprise
 
+## v4.22.0 Fixes
```
tmcconechy: This will make it in 4.21 so you can put this in that section

src/components/toolbar-flex/_toolbar-flex.scss
```diff
@@ -12,6 +12,7 @@
   align-items: center;
   display: flex;
   flex-flow: row nowrap;
+  height: 100%;
```
tmcconechy: We removed this due to a bug. So we cannot add a height here. Is there some reason you need this? If for the CAP perhaps can put an explicit height on the toolbar isolated under the contextual-action-panel fx..Add it to this rule instead:

```
.modal.contextual-action-panel .flex-toolbar {
  margin: 0 auto;
  padding: 0 5px;
  height: inherit;
}
```

EdwardCoyle:
@ericangeles I just opened #2624 with a fix for the vertical centering of the Flex Toolbar in CAP.  I'd say to just remove the `height: 100%` in this PR and leave the rest.  When we merge `4.20.x` back into `master` on release, the fix will be present there.

ericangeles:
All good! Thanks @tmcconechy and @EdwardCoyle 🍌

PR #2630, opened 2019-08-08
2588 - Pager: Numbers illegible in Uplift Dark theme 

### LINKED ISSUE #2588: Pager font color is not readable when in uplift theme dark mode
**Describe the bug**
Pager font color is not readable when in uplift theme dark mode

**To Reproduce**
1. Go to https://4200-beta1-enterprise.demo.design.infor.com/components/pager/example-listview.html?theme=uplift&variant=dark
Notice that the pager font color is not that readable. 

**Expected behavior**
Pager text should be readable

**Version**
Beta 4.20

**Screenshots**
image

**Platform**
All Browser, OS and devices

**Additional context**
No any other context about the problem here.

davidcarlsonberg:
The pager numbers are illegible in uplift dark theme.

Closes #2588 

- Pull branch, build, run app
- Visit localhost:4000/components/pager/example-listview.html?theme=uplift&variant=dark
  - Ensure the pager numbers are legible.

### DIFF
--- a/src/themes/theme-uplift-dark.scss
+++ b/src/themes/theme-uplift-dark.scss
@@ -348,7 +348,7 @@ $radio-checked-bg-color: $radio-color-checked-initial-background;
 $radio-disabled-checked-bg-color: $radio-color-checked-disabled-background;
 
 // Pager
-$pager-text-color: $theme-color-brand-primary-base;
+$pager-text-color: $theme-color-palette-azure-100;
 $pager-selected-color: $font-color-extrahighcontrast;
 $pager-focus-border-color: $input-color-focus-border;
 $pager-hover-color: $font-color-extrahighcontrast;
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,7 @@
 - `[Contextual Action Panel]` Fixed an issue where if the title is longer, there will be an overflow causing a white space on the right on mobile view. ([#2605](https://github.com/infor-design/enterprise/issues/2605))
 - `[Datagrid]` Fixed a bug where the value would clear when using a lookup editor with a mask on new rows. ([#2305](https://github.com/infor-design/enterprise/issues/2305))
 - `[Fileupload Advanced]` Added custom errors example page. ([#2620](https://github.com/infor-design/enterprise/issues/2620))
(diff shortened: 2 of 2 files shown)


src/themes/theme-uplift-dark.scss
```diff
@@ -348,7 +348,7 @@ $radio-checked-bg-color: $radio-color-checked-initial-background;
 $radio-disabled-checked-bg-color: $radio-color-checked-disabled-background;
 
 // Pager
-$pager-text-color: $theme-color-brand-primary-base;
+$pager-text-color: $theme-color-palette-azure-100;
```
tmcconechy: How does $theme-color-brand-primary-alt look here? Im wondering because there are other cases exactly like this like on http://master-enterprise.demo.design.infor.com/components/datepicker/example-index.html?theme=uplift&variant=dark&colors=0053B1

So just wondering if we need to change something else to improve this in other areas like darken up both `$theme-color-brand-primary-alt` and `$theme-color-brand-primary-base`

davidcarlsonberg: I tried `brand-primary-alt` first and didn't think it looked much better in terms of legibility. I think the same for the example you post above. I think it would be worthwhile to darken or lighten `$theme-color-brand-primary-alt` and `$theme-color-brand-primary-base` given these examples. Does that mean involving design?

tmcconechy: Yeah i guess it does or can we change tokens @clepore and @nickwynja ?So a fix might be getting that changed or changing it, making a new version and adding it to this branch in package-lock.json we might as well fix it for all the places not just the one.

nickwynja: Yes, they can be updated. There are some `$theme-color-font-*` tokens that might be better suited for this.

tmcconechy: We could use ~$theme-color-font**` but it seems like the primary and alt color are just not dark enough to show on dark dialogs/panes. So what about changing those up ?

nickwynja: The tokens haven't been thoroughly tested for Dark and Contrast in Uplift. I'd recommend having @kentonquatman review these and update accordingly.

davidcarlsonberg: Closing this pull request in favor of design looking at the affected components and determining a better color and token.

Issue #2637, opened 2019-08-09
getCulturePath does not handle if sohoxi.js filename has a hash part


guoliang:
**Describe the bug**
getCulturePath does not handle if sohoxi.js filename has a hash part

**To Reproduce**
1. Add a hash part to filename e.g. sohoxi.hashedpart123.js
2. Change locale to e.g. en-NZ

**Expected behavior**
Language en-NZ loaded in Soho.Locale.cultures

**Version**
- ids-enterprise: 4.18.1

guoliang:
I created a PR https://github.com/infor-design/enterprise/pull/2638

tmcconechy:
OK, please assign yourself to issues you plan to address. (We cant assign some users only those attached to the project). Thanks for fixing this!

guoliang:
I don't think I can assign myself to this issue either. At least I cannot see that option

tmcconechy:
Ok strange? I wasnt sure about that either. Oh well.

PR #2638, opened 2019-08-09
2637: Adds regexp to handle sohoxi file with hashed part

guoliang:
Refer to https://github.com/infor-design/enterprise/issues/2637


tmcconechy:
MErged as was previously approved.

PR #2641, opened 2019-08-09
2312- Change datagrid to not show the dirty indicator on row indicator cells

### LINKED ISSUE #2312: DataGrid: Row and cell indicators overlap cell value
Describe the bug
The row status indicator and the cell dirty indicator overlaps the value in the cell.  Making the cell value difficult to read.

To Reproduce

1. Using the attached test-overlaps-row-cell-indicators.html
2. Click on the 'Add Row (top)'
3. Click on the Product Id cell for the new row
4. Click on the lookup icon to open the lookup selection window
5. Select one of the rows

Notice that the row indicator and cell indicator overlaps the value in the cell.  
Screen Shot 2019-06-04 at 2 58 03 PM

Expected behavior
Cell value should be fully visible.

tmcconechy:

In the case of a row status AND dirty indicator on the same cell a UI problem occurs. In order to fix this i made a rule that if the row is new then the dirty indicator will not be shown on every cell for that row (since the whole row is new its all dirty anyways). Also if the cell is the first cell with the indicator and that cell is dirty the indicator will take precedence.

Fixes #2312 

- http://localhost:4000/components/datagrid/test-overlaps-row-cell-indicators.html
- edit the first or second cell on the second row and commit the edit -> dirty indicator will be shown
- hit new row (top)
- on the new row (with the new indicator) edit the first and other cells -> dirty indicator will not be shown since the whole row is new


tmcconechy:
@deep7102 I added a fix to move the dirty indicator as you noted on new rows.

PR #2642, opened 2019-08-09
536 - Make the onKeydown fire on any type of key

tmcconechy:

While addressing 536 is was noted that the datagrid onKeydown event only fired on tab. Expanded this to fire on any key

Fixes #536 (will add a PR for the NG part)

- go to http://localhost:4000/components/datagrid/test-editable-onkeydown.html
- click in any cell and type
- any key you type will omit an event shown in the toast and veto (nothing will happen otherwise as the event is vetoed)


tmcconechy:
Have to change this to a callback response to work well in NG.

PR #2652, opened 2019-08-13
2157 - Fix key navigation for inline editors with datagrid

### LINKED ISSUE #2157: Datagrid: Tab on editable cells with inlineEditor=true not triggering the cellChange event
We have datagrids with coulmns set to inlineEditor = true. When the cell values are changed and we tab thru it does not trigger the 'cellChange' event, like it does for a normal Input editor.

This is the SOHO example that has it:
https://latest-enterprise.demo.design.infor.com/examples/saleshub/product-search.html

To Reproduce

Open the developer console(F12)
1) Select the qty cell of say, row1, Change value to 3, Click on the row2 qty cell - chk the console - Works Fine .. shows details of the row and cell changed.
2) Row2, qty cell - change value and Tab - Check console - No log 

Expected behavior:
Wld be great if the cellChange event is triggered on tab key when the cell is marked as inlineEditor

Platform

Browser Name [e.g. chrome, IE, safari]
currently using SOHO version 4.17.1

deep7102:
Fixed key navigation for inlineEditor with datagrid.

Closes #2157

- Navigate to http://localhost:4000/components/datagrid/test-editable-inline-editor.html
- Open developer tools
- Use tab key to focus any cell in column `Quantity`
- Change value and press tab to go next column
- See console, it should fire and log the info


tmcconechy:
i tested and it works but there is a conflict @deep7102 https://github.com/infor-design/enterprise/pull/2652/conflicts as i made a similar change. I think you will have to look at that and compare to this change https://github.com/infor-design/enterprise/pull/2642/files#diff-7a47cbfa760f2056d4f9b91cf583d3fcR7822

PR #2661, opened 2019-08-19
1038 - Configure Node-Sass to only import Sass partials once

### LINKED ISSUE #1038: Custom Builds:  Create `node-sass` importer to prevent duplicate Sass partial imports
To correctly link together SCSS dependencies of our more complex components for custom builds, we need the ability to create `@import` statements linking to dependencies on top of each source file that needs them.

With our current sass builder, this isn't possible.  Linking to the same SCSS partial in multiple files causes that batch of CSS rules to appear in the compiled files as many times as it's linked.

- [x] Create a small wrapper around the actual `node-sass` process (currently we use a grunt plugin)
- [x] Add custom importers to `node-sass` that are capable of detecting previously-loaded imports.
  
Light reading:  https://paul.wellnerbou.de/2015/05/18/avoid-multiple-imports-of-the-same-scss-file-with-sass/

We should try to get one of the established `node-sass` importer plugins to correctly work with our build stack (I tried to get [node-sass-import-once](https://github.com/at-import/node-sass-import-once/ and [node-sass-once-importer](https://github.com/maoberlehner/node-sass-magic-importer/tree/master/packages/node-sass-once-importer)

The article above also describes a method of writing a mixin to conditionally run an import if it hasn't previously been loaded.  This method's less desirable because we'd have to create a strange syntax around all of our imports, but it could work.

EdwardCoyle:
This PR makes a change to the build system that configures `node-sass` to only import required partials once, instead of allowing them to be imported (and therefore generate rules) multiple times.  This is done by incorporating the [`node-sass-once-importer`](https://github.com/maoberlehner/node-sass-magic-importer/tree/master/packages/node-sass-once-importer) into the build system.

What the importer allows is for imports of SASS dependencies in any files that need them (for example, `_button.scss` to be included in Pager and Listview, and to not be imported more than once), which is similar to how the ES6 importing system currently works.

This change will allow us to liberally place dependencies on other SASS partials inside of mid-to-complex components, and expect the source code to be cleaner in the process.  This will also improve our ability to split the SASS source code into many smaller files and correctly pull in only what's necessary when generating custom builds.

See the changes at the top of `_listview.scss` in this PR for an example of how it works.

Closes #1038 

_Test that the build actually runs:_
- Pull this branch
- Run `npm prune && npm i` to include the new dependency
- Run a build/minify/verify with `npm run clean && npm run build && npm run minify && npm run build:verify`
- Ensure this process completes with no errors and the files are verified properly.

_To test that there are no duplicate imports:_
- Start the demoapp and navigate to http://localhost:4000/components/listview/test-empty-message.html
- Open a dev tools instance.
- Find the `div.empty-message` element in the Inspector tab.  Ensure that the CSS rules coming in from `_emptymessage.scss` do not show up more than once in the Styles tab (which would indicate multiple imports):
Screen Shot 2019-08-19 at 12 39 07 PM
- Open http://localhost:4000/components/listview/example-paging.html
- In dev tools, inspect the `li.pager-prev > a` element in the Pager Toolbar.
- Ensure that none of the CSS rules coming from `_pager.scss` are duplicated:
Screen Shot 2019-08-19 at 12 44 53 PM

**NOTE:** Listview is the only component modified for the new import system in this PR.  All other components should be the same, so it's worth a quick smoke test of any other components you like to ensure nothing else has changed unexpectedly.


EdwardCoyle:
To those testing: I updated that baseline image to reflect the fact that the styles from `_emptymessage.scss` are actually being imported now.  It appears we've had [this source code](https://github.com/infor-design/enterprise/blob/master/src/components/emptymessage/_emptymessage.scss) for a couple years, but it somehow never made it's way into the [SASS component listing](https://github.com/infor-design/enterprise/blob/master/src/core/_controls.scss) (search for "emptymessage" in that file and it won't appear there).

PR #2665, opened 2019-08-19
2665 - Datagrid: Select checkbox announce description

### LINKED ISSUE #2031: Accessibility: Datagrid does not announce selected row
**Describe the bug**
The selected column in the datagrid is announced by AT as "row selected" instead of what user is selecting. It may be advisable to add the name of the item user is selecting to the select span name.

**To Reproduce in Landmark**
1. Go to https://hcm-tampm01-prd.inforcloudsuite.com/hcm/EmployeeSelfService/form/Employee%287004,700417%29.LRCFullEmployeeProfile?csk.showusingxi=true&csk.JobBoard=INTERNAL&csk.HROrganization=7004&menu=LRCEmployeeMenu.MyProfile
2. Click on any magnifying glass icon to launch a modal dialog
3. While using AT, select a row within the datagrid 
4. AT does not announce the selected row. 

**How to Fix for EP**
- This example has the same problem: http://master-enterprise.demo.design.infor.com/components/datagrid/example-mixed-selection.html
- Fix this so that selected row is announced with some setting like a primary id column, that is specified and that is read when the user tried to select a row.

**Expected behavior**
AT should announce the selected option and not just that a row is selected. 

**Additional context**
- component library example has the same problem: http://master-enterprise.demo.design.infor.com/components/datagrid/example-mixed-selection.html
- see also: https://jira.lawson.com/browse/LMCLIENT-24215

davidcarlsonberg:
There is a lack of detail in the announced description in the aria-label when one tabs into a checkbox cell.
This PR adds a new setting to allow optional strings in an array that determine what extra information from a row is concatenated into the `aria-label` for that row's selection checkbox.
PR also creates and utilizes a new xss utility to ensure a string allows spaces and is alphanumeric.

Closes #2031 

- Pull branch, build, run app
- Visit http://localhost:4000/components/datagrid/example-mixed-selection in a Windows environment with JAWS enabled.
  - ensure the announcement is more descriptive than just "Select" if options are provided.


src/components/datagrid/datagrid.formatters.js
```diff
@@ -311,7 +311,8 @@ const formatters = {
     if (!value) {
       isChecked = api.isRowSelected(item);
     }
-    return `<div class="datagrid-checkbox-wrapper"><span role="checkbox" aria-label="${(col.name ? col.name : Locale.translate('Select'))}" class="datagrid-checkbox datagrid-selection-checkbox${(isChecked ? ' is-checked no-animate' : '')}" aria-checked="${isChecked}"></span></div>`;
+    const ariaString = ((item.id.length > 0) && (item.type.length > 0)) ? '' : ` ${item.id} ${item.type}`;
```
tmcconechy: The only problem with this is that `type` is a specific column in this example. So it would only work if they happen to have a type column. What i would suggest is a new setting on the datagrid like 'idColumns' where you can list one or more columns that they might want to specify to show info for this.

Or maybe we can take the first column after the checkbox or first two? Anyways i think this needs some more configuration as it wont always be correct with just `id` and `type`.

davidcarlsonberg: I think `item` is always available, yes?

tmcconechy: Yes item is (which would be the current row of data thats passed in the dataset). We just need a way to configure some column names to show here out of that. I think maybe idColumns would be a useful thing that could have the chance to use later. So an array of strings you can pass that correspond to the columns that make the row unique.

PR #2675, opened 2019-08-21
2526 - Fix an issue running examples in IE and a follow up bug

### LINKED ISSUE #2526: Datagrid Filter: disableClientFilter and load data causing an infinite loop when updating columns

tmcconechy:

This PR fixes a follow up comment on 2526. In order to solve part of the issue i configured the dropdown on the filter row to close in the example.

Fixes #2526 
 
http://localhost:4000/components/datagrid/test-filter-load-data-update-columns.html
- open the filter drop down
- pick a value
- click out and close
- open filter then select Place On-Hold, then close dropdown
- open filter again, deselect Place On-Hold then select Request Review, close dropdown
- values should be in sync...
- open the page in IE 11 - will now load


davidcarlsonberg:
I see one, what I would call, minor issue. The placeholder in the closed dropdown remains until you open and close the dropdown a sequential time. In Chrome.
dropdown

tmcconechy:
@davidcarlsonberg I fixed the last issue you mentioned.

PR #2677, opened 2019-08-21
2664 - fix lifecycle bug between Flex Toolbar and Popupmenu in SPA environments

### LINKED ISSUE #2664: Toolbar Flex Component: Menu button's popup menu lost when attachToBody used
**Describe the bug**
The popupmenu associated with a menu button is getting lost when using attachToBody: true. When tracking thru the code, on init the menu is being moved up to the <body> tag (which is correct), but after the page is rendered, I don't see the menu anywhere in the dom.

I think the menu is getting moved, and then destroyed. Maybe something from the memory leak fix a couple weeks ago?

Also, we need to use the attachToBody:true option, since without it iOS broswers put the popupmenu content underneath the contents below a toolbar. Only when viewing it on a physical device.

**To Reproduce**
I currently can only reproduce this on a landmark system. I can share my system with the person who works on it.

**Expected behavior**
Menu button should open correctly.

**Version**

- ids-enterprise: 4.21.0-dev

**Screenshots**
This is seen in the browser console after the page is rendered:
image

**Platform**
 - All

EdwardCoyle:
This PR makes the following changes:
- The Popupmenu's `teardown` method has been cleaned up to prevent errors when the `removeOnDestroy` setting is enabled.
- Fixed the Popupmenu's `refreshMenuItem` setting to better handle situations where an AJAX call may have changed a top-level menu item's contents if it contains a submenu.
- Fixed a small bug in `ToolbarFlexItem` which would propagate IDs from a menu button into its corresponding list item in the More Actions menu, causing duplicate IDs.

Closes #2664 

- Pull this branch, run app
- Open http://localhost:4000/components/popupmenu/test-flex-toolbar.html
- Open both the menu button and the More Actions button and make sure they render.  Closing either menu should not throw console errors.
- Click the "Trigger `updated()` on Toolbar" button.  It should reset the entire toolbar without throwing console errors.
- Click the "Destroy Menu Button" button.  Then try to open the menu button.  It should no longer have a menu associated with it, and do nothing.
- Click the "Invoke Menu Button" button. Then try to open the menu button.  The menu should render as expected, and no console errors should be thrown:

Screen Shot 2019-08-21 at 5 13 43 PM

----
Paging @Sovia


EdwardCoyle:
This PR still needs some work... the failing e2e tests are legitimate.  This is the diff from the worst one:

flextool-index-open-more-menu-overflowed-menu-button-submenu-chrome-450x1000-dpr-1

The spilled-over menu button seems to be missing its children in the test.

EdwardCoyle:
After some more troubleshooting we're seeing issues in scenarios where:
- Flex Toolbar is invoked with no contents
- Contents are loaded and an `updated()` call to the Flex Toolbar API is made
- Items in the toolbar's buttonset area that should be setup internally on the More Actions button as a `predefinedItems` property are not propagating, and therefore being lost when the AJAX call is made.

From @Sovia's environment, an example of the NG controls building a toolbar with no contents to start:

MicrosoftTeams-image copy

This case needs to be replicated in an EP test before we can accurately fix.

PR #2688, opened 2019-08-23
2102 - Add support for tree grid with frozen columns

### LINKED ISSUE #2102: Datagrid TreeView: expandRow is not triggered when the first column in a treeview is frozen
**Describe the bug**
When a datagrid treeview has a frozen first column, clicking on '+' to expand a row does not fire expandRow event. As a result, the children node data is not fetched.

**To Reproduce**
Test a treeview example with first column set as frozen.

**Expected behavior**
Clicking on '+' expands the tree node.

**Additional context**
Related to
#464
#1638

tmcconechy:

Tree and Frozen Columns did not work together. This PR addresses that.

Fixes #2102 

http://localhost:4000/components/datagrid/test-tree-frozen-columns
- ensure you can scroll right with the tree nodes frozen on the left
- expand and collapse some nodes
- note that this pattern will not work well on mobile as there is not enough room (this is fine)

http://localhost:4000/components/datagrid/test-tree-filter.html
- retest this as per https://github.com/infor-design/enterprise/pull/836


davidcarlsonberg:
After filtering on "HMM" the expand button is disabled. Is this intentional?
treegrid

tmcconechy:
No, but also not the point here. I will remove the filter from this, i think the problem is the "back end" is not wired up to handle this properly.

tmcconechy:
@davidcarlsonberg FYI i checked this out and talked to @deep7102 this is actually correct with the button being disabled. We did this on purpose because in this case no children match the filter so you cannot expand and see them (they are effectively filtered out).

tmcconechy:
Buttttt.. I added disabled styles so it looks visual disabled now.

PR #2727, opened 2019-09-04
Fix dependency and build issue

tmcconechy:

Set chokidar to fix at 1.0.1 as 1.0.2 has issues in that fails as per https://github.com/paulmillr/chokidar/issues/855

Also removed the node version check on npm i for now as this fails on jenkins as that version is not there.

NA

- pull this branch
- `rm -rf node_modules && npm i`
- start grunt watch
- save a js file and should reload

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "release:final": "release-it",
     "test": "npm run mdlint && npm run eslint && npm run stylelint && npm run functional:ci && npm run e2e:ci",
     "node-version-check": "npx check-node-version --node 10.16",
-    "postinstall": "npm run node-version-check"
+    "//postinstall": "npm run node-version-check"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "release:final": "release-it",
     "test": "npm run mdlint && npm run eslint && npm run stylelint && npm run functional:ci && npm run e2e:ci",
     "node-version-check": "npx check-node-version --node 10.16",
-    "postinstall": "npm run node-version-check"
+    "//postinstall": "npm run node-version-check"
```
EdwardCoyle: Is this right?

tmcconechy: Yes its a way to comment it out so it wont run. Wanted to revisit it later. But we can remove?

PR #2772, opened 2019-09-10
NA - Fix a regression on empty widgets [v4.21.x]

tmcconechy:

While testing a chart PR i noticed 4.21 has a regression no on empty widgets. So scoped the new CSS better.

- NA

- go to http://localhost:4000/components/bar/test-empty.html - should look as before as per https://4200-enterprise.demo.design.infor.com/components/bar/test-empty.html
- go to http://localhost:4000/components/datagrid/test-empty-card.html - should be able to scroll

### DIFF
--- a/src/components/datagrid/_datagrid.scss
+++ b/src/components/datagrid/_datagrid.scss
@@ -734,7 +734,7 @@ $datagrid-short-row-height: 25px;
 
 .card,
 .widget {
-  .empty-message {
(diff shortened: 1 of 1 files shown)


tmcconechy:
@davidcarlsonberg Yes i just meant in general about the scrolling. Actually reduced the padding a bit to recenter in some cases so i think thats fine.

PR #2792, opened 2019-09-13
2699 - Add support for context menu event with charts

### LINKED ISSUE #2699: Charts: Inconsistent context menu events in charts
**Describe the bug**

 Bar and Pie/Donut throw a contextmenu event with 3 params including ui and data. We need all the charts to throw the same event so we can grab the index of the element that caused the event to be throw. Using the index we can have the context menu to show for the chart element clicked on.

```
column:  self.element, d3.select(this).nodes()[0], d
pie: self.element, d3.select(this).select('path').nodes()[0], d
bar: (dont see any event at all)
line: (dont see any event at all)
line: (dont see any event at all)
bubble: (dont see any event at all)
radar: (dont see any event at all)
scatter: (dont see any event at all)
```

**To Reproduce**
1. We need to construct an example of a right click menu on all chart elements that uses contextmenu and surfaces a menu

**Expected behavior**
All Chart need to throw a consistent event when a chart element is clicked on for a context menu. 

**Version**
- ids-enterprise: [e.g. v4.21.x]

**Platform**
 - N/A

deep7102:
Added support for context menu event with charts.

Closes #2699

http://localhost:4000/components/area/example-contextmenu.html
http://localhost:4000/components/bar/example-contextmenu.html
http://localhost:4000/components/bar-grouped/example-contextmenu.html
http://localhost:4000/components/bar-stacked/example-contextmenu.html
http://localhost:4000/components/bubble/example-contextmenu.html
http://localhost:4000/components/bullet/example-contextmenu.html
http://localhost:4000/components/column/example-contextmenu.html
http://localhost:4000/components/column-grouped/example-contextmenu.html
http://localhost:4000/components/column-stacked/example-contextmenu.html
http://localhost:4000/components/donut/example-contextmenu.html
http://localhost:4000/components/line/example-contextmenu.html
http://localhost:4000/components/pie/example-contextmenu.html
http://localhost:4000/components/positive-negative/example-contextmenu.html
http://localhost:4000/components/radar/example-contextmenu.html
http://localhost:4000/components/scatterplot/example-contextmenu.html
http://localhost:4000/components/sparkline/example-contextmenu.html
- Open above links
- Open developer tools
- Do right click to fire `contextmenu`
- Should show popup menu
- See console log for `Contextmenu`


src/components/radar/radar.js
```diff
@@ -105,6 +105,7 @@ Radar.prototype = {
    * @returns {object} The component prototype for chaining.
    */
   init() {
+    this.namespace = utils.uniqueId({ classList: [this.settings.type, 'chart'] });
```
tmcconechy: Wonder why its necessary to use a unique id as a namespace (not bar or column or whatever as the component name?)

tmcconechy:
Only small thing (seems like not on this PR but recent). 
- http://localhost:4000/components/bubble/example-contextmenu.html
- hover a bubble
- one item is "undefined"

src/components/radar/radar.js
deep7102: Good catch, I agree no need unique id here, I had some issue with event types so I was playing around left it there while debugging. But yes going to remove this unique id since no need and it make extra call -thanks

tmcconechy: Ok Cool Note that i saw this in several of the charts using the unique ID

deep7102: Oh! we need this unique id, after retesting saw we have some events bind with `$('html')` and `$('body')` so if someone use more then one on page and later one of them get `destroy` it will unbind those events for all, so need unique events to bind for each chart.

deep7102:
Hi @tmcconechy updated now, to fix undefined item in tooltip with bubble chart

PR #2804, opened 2019-09-17
2660 - Minify Culture Files

### LINKED ISSUE #2660: Minify culture js files before shipping to npm
No problems related.

- [x] minify the culture files in the `dist/` folder (@tmcconechy do you forsee any issues with this?)
- [ ] find a way to test the compressed culture files
- [x] replace UglifyES with Terser
- [x] re-test the build

At the moment, `ids-enterprise` is shipped to npm with `sohoxi.min.js`. However all js files in `dist/js/cultures` are not minified. These files should be minified as well, like `sohoxi.min.js`.

and also perhaps, minifying process should switch to use `terser` instead of `UglifyJS`

**N.A.**

**Additional context**
**N.A.**

EdwardCoyle:
This PR adds the following functionality to the build system:
- Replaces the now unmaintained UglifyES with [Terser](https://github.com/terser/terser)
- Runs all Locale culture files through the minify script
- Adds a setting to `SohoConfig` called `minifiedCultures` that can be used to tell the locale system to load the compressed `.min.js` version of the cultures instead of the uncompressed versions

Closes #2660

- Pull this branch
- In a terminal, run `npm i && npm run clean && npm run build` to get a fresh build.
- Run `npm run minify` for minification.  There's a new message that should display when it completes successfully:
Screen Shot 2019-09-17 at 1 18 29 PM
- Run `npm run build:verify`.  There should be no errors and all files should be accounted for.
- Run the demoapp.
- Navigate to a URL like http://localhost:4000/components/locale/test-big-number.html?minify=true&locale=ar-SA, with the query param for `minify` set to true.
- Open a dev tools console.  Inspect the bottom of the `<head>` tag in the page to make sure the culture files fetched by Locale contain the `.min.js` at the end, instead of just `.js`:
Screen Shot 2019-09-17 at 1 26 25 PM


app/src/js/custom-route-options.js
```diff
@@ -30,6 +30,12 @@ module.exports = function customRouteOptions(req, res) {
   const customOpts = new CustomOptions(req, res);
   const url = req.originalUrl;
 
+  // All routes will require a minified `sohoxi.js` and locale cultures,
+  // if the query param was previously set.
+  if (res.opts.minify) {
+    SohoConfig.minifiedCultures = true;
```
tmcconechy: `minifyCultures` feels better for a name to me

scripts/minify-js.js
```diff
@@ -71,22 +124,41 @@ function minify() {
...
       });
     }).catch((e) => {
+      logger('error', e);
       reject(e);
       process.exit(1);
     });
   });
 }
 
+debugger;
```
tmcconechy: Left debugger in..

davidcarlsonberg:
I get an error on `npm run build:verify`:
error

tmcconechy:
Im also seeing @davidcarlsonberg 's error. And on the step where i run http://localhost:4000/components/locale/test-big-number.html?minify=true&locale=ar-SA 

Im getting errors and the page isnt loading. I did `npm run start` first after following the steps in order and it looks like the sohoxi.js was not generated ?

EdwardCoyle:
@tmcconechy @davidcarlsonberg try this again?  I changed the check for whether or not to produce a sourcemap.  I'm guessing when I originally tested this, I hadn't run the `clean` properly.

tmcconechy:
The verify command is now fixed. However i cannot run the demo app. 

1. Navigate to a URL like http://localhost:4000/components/locale/test-big-number.html?minify=true&locale=ar-SA, with the query param for minify set to true is giving all sorts of errors as if shohoxi.js is not there.

Tired this with clean folder `rm -rf node_modules && npm i` and then `npm run start`. Then load the page (with any parameters)

EdwardCoyle:
I think the problem you're seeing is that `npm run build` doesn't run the `minify` or `build:verify` scripts.  IIRC, this was by design to make dev a bit faster, because the minify script can be a time hog.

When we make a release, these files do exist in the `dist/` folder and zip files because the release process runs `npm run build:release`.  I ran that followed by an `npm run quickstart` for the demoapp, and was no longer seeing the missing files problem.

PR #2815, opened 2019-09-19
434 - Add Percent Locale examples

### LINKED ISSUE #434: SOHO-7138: 4.0 Mask does not format percentage values according to Locale
When an input field is using a Mask pattern for suffix of percent sign, the value does not get properly formatted according to the Locale.

Most locales are either '#,##0 %' or '#,##0%; tr-TR has '%#,##0'; ar-EG has a different symbol.

We do have the ability to set the appropriate percent sign on the Mask options, but the Mask does not properly format the value according to Locale.

Wondering if percentFormat should be part of the Mask options?

Landmark does not use currency format, but believe a similar issue exists for it.

See the attached screenshots and use the attached test-new-mask-prefix-suffix.html that has changes for suffix.

To duplicate, using the attached test
- Enter a value in the Percentage Suffix input field
- Enter the same value, without a suffix, in the Input Value field
- Compare the Percentage Suffix value with the Transformed Locale Suffix value which was created with the Input Value field and Locale.formatNumber using percent options.

Expectations are the values match.

tmcconechy:

The logic was already in place for 434 so added an example and some tests.

Closes #434 

- go to http://localhost:4000/components/mask/test-percent-locale
- type: 100
- should show to the en-US format: 100 %
- http://localhost:4000/components/mask/test-percent-locale?locale=ar-SA
- type: 100
- should show to the ar-SA format: 100 ٪
- http://localhost:4000/components/mask/test-percent-locale?locale=tr-TR
- type: 100
- should show to the tr-TR format: %100


tmcconechy:
Did this from wrong branch redoing

PR #2823, opened 2019-09-23
2797 - Fix keyword search to reset on paging and sort

### LINKED ISSUE #2797: Datagrid: Keyword highlighting no longer works in paged tables
Since upgrading to version 4.20.1, when you enter a keyword search in the DataGrid, it no longer highlights the matched string in the table rows. This used to work in older releases.

Working Example: https://design.infor.com/code/ids-enterprise/4.14.0/demo/components/datagrid/example-paging-client-side

Non-working Example: https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/example-paging-client-side

Hint: sohoxi.js keywordSearch method on the DataGrid looks like the problem...

`      if (!this.settings.paging) {
        this.highlightSearchRows(term);
      }`

We've actually documented this as a feature in our Product Guides across multiple tables, so I'd like the functionality re-instated.

tmcconechy:

The feature done on 2797 did not work for sorting and filtering. Found that afterpaging did not fire for client side paging (maybe never have actually). Refactored some code to an afterPaging method but kept the event in case it is used.

Fixes #2797 

- go to http://localhost:4000/components/datagrid/example-paging-client-side
- 0 search for "ssor 1" -> rows will be highlight and filtered
- sort on any sortable column -> rows will remain highlight and filtered
- go to page 2 -> rows will remain highlight and filtered
- retest http://localhost:4000/components/lookup/example-multiselect-paging-serverside.html
- and http://localhost:4000/components/datagrid/example-paging

### DIFF
--- a/app/views/components/datagrid/example-paging.html
+++ b/app/views/components/datagrid/example-paging.html
@@ -66,7 +66,7 @@
             response(res.data, req);
           });
         },
-        toolbar: {title: 'Data Grid Header Title',filterRow: true, personalize: true, results: true, keywordFilter: true, actions: true, rowHeight: true, collapsibleFilter: true}
+        toolbar: {title: 'Data Grid Header Title', filterRow: true, personalize: true, results: true, keywordFilter: false, actions: true, rowHeight: true}
       })
       .on('selected', function (e, args) {
         console.log('Selected: ', args);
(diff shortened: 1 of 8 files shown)


davidcarlsonberg:
Not sure what I'm testing here: http://localhost:4000/components/lookup/example-multiselect-paging-serverside.html ... just that nothing is broken?
Should highlighting happen here?: http://localhost:4000/components/datagrid/example-paging

tmcconechy:
http://localhost:4000/components/lookup/example-multiselect-paging-serverside.html
- Yes just test nothing seems broken

http://localhost:4000/components/datagrid/example-paging
- No because its server side paging and thats not implemented in the serveside demo app service. This really only works on client side / non paging

PR #2837, opened 2019-09-25
2832 - Add a fix datepicker destroy and mask invocation

### LINKED ISSUE #2832: Datepicker: mask is lost after updated is called
**Describe the bug**
The mask on a date picker is lost after updated is called on the component.

The teardown method is causing this.element.data('mask') to return "" instead of undefined, so the mask doesn't get re-inited [here](https://github.com/infor-design/enterprise/blob/master/src/components/datepicker/datepicker.js#L418).

**To Reproduce**
1. Copy test-mask-after-update
[test-mask-after-update.zip](https://github.com/infor-design/enterprise/files/3648059/test-mask-after-update.zip) into app/views/components/datepicker.
2. Navigate to '/components/datepicker/test-mask-after-update.html'
3. Enter some numbers manually into the field, see that slashes are automatically put in and the date is formated
4. Clear out the field, and click the updated button
5. Enter some numbers manually into the field, see that no slashes are added and no formatting occurs.

**Expected behavior**
The mask should still work after updated is called on the datepicker component.

**Version**

- ids-enterprise: v4.22.0-dev and earlier

**Screenshots**
DatepickerMaskIssue

**Platform**
 - All

tmcconechy:

Fixes an issue that the mask is not reinvoked when datepicker is destroyed.

Fixes #2832 

- go to http://localhost:4000/components/datepicker/test-mask-after-update.html
- type in a value and not the mask works
- clear the value
- hit destroy
- re-type in a value and not the mask works
- restest http://localhost:4000/components/datepicker/example-validation.html


src/components/datepicker/datepicker.js
```diff
@@ -1653,6 +1652,8 @@ DatePicker.prototype = {
     if (maskApi) {
       maskApi.destroy();
     }
+    this.element.removeAttr('data-mask');
+    this.element.removeData('mask');
```
pwpatton: @tmcconechy Does compeltely removing these cause the init to rebuilt them?

tmcconechy: Yes because of this line https://github.com/infor-design/enterprise/blob/master/src/components/datepicker/datepicker.js#L418 (before was '')

pwpatton: Ah see it now :)

docs/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - `[Datagrid]` Made the text all white on the targeted achievement formatter. ([#2730](https://github.com/infor-design/enterprise/issues/2730))
 - `[Datagrid]` Fixed keyword search so that it will again work with client side paging. ([#2797](https://github.com/infor-design/enterprise/issues/2797))
 - `[Datepicker]` Moved the today button to the datepicker header and adding a setting to hide it if wanted. ([#2704](https://github.com/infor-design/enterprise/issues/2704))
+- `[Datepicker]` Fixed a bug in datepicker that the destroy would not re-add masking functionality. [2832](https://github.com/infor-design/enterprise/issues/2832))
```
davidcarlsonberg:
```suggestion
- `[Datepicker]` Fixed a bug in datepicker where the destroy method does not readd the masking functionality. [2832](https://github.com/infor-design/enterprise/issues/2832))
```

PR #2840, opened 2019-09-25
434- Refix : Improve the percent example

### LINKED ISSUE #434: SOHO-7138: 4.0 Mask does not format percentage values according to Locale

tmcconechy:

Refix the issue for 434 as it was not possible to have the percent show after typing 1 or 10 Only 100% worked.

Fixes #434 

- go to http://localhost:4000/components/mask/test-percent-locale
- type: 100
- should show to the en-US format: `100 %`
- clear field and type: 10
- should show to the en-US format: `10 %`
- clear field and type: 1
- should show to the en-US format: `1 %`
- http://localhost:4000/components/mask/test-percent-locale?locale=ar-SA
- type: 100
- should show to the ar-SA format: `٪ 100`
- clear field and type: 10
- should show to the ar-SA format: `٪ 10`
- clear field and type: 1
- should show to the ar-SA format: `٪ 1`
- http://localhost:4000/components/mask/test-percent-locale?locale=tr-TR
- type: 100
- should show to the tr-TR format: `%100`
- clear field and type: 10
- should show to the ar-SA format: `%10`
- clear field and type: 1
- should show to the ar-SA format: `%1`

### DIFF
--- a/app/www/test/js/sv-SE.f0r73571n6purp0537051mul473h45h.js
+++ b/app/www/test/js/sv-SE.f0r73571n6purp0537051mul473h45h.js
@@ -51,6 +51,8 @@ Soho.Locale.addCulture('sv-SE', {
   numbers: {
     percentSign: '%',
     percentFormat: '### %',
+    percentSuffix: ' %',
+    percentPrefix: undefined,
     minusSign: '-',
     decimal: ',',
     group: ' ',
--- a/app/www/test/js/sv-SE.f0r73571n6purp0537051mul473h45h.js
+++ b/app/www/test/js/sv-SE.f0r73571n6purp0537051mul473h45h.js
@@ -52,7 +52,7 @@ Soho.Locale.addCulture('sv-SE', {
     percentSign: '%',
     percentFormat: '### %',
     percentSuffix: ' %',
-    percentPrefix: undefined,
+    percentPrefix: '',
     minusSign: '-',
     decimal: ',',
     group: ' ',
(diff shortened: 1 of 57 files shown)


app/www/test/js/sv-SE.f0r73571n6purp0537051mul473h45h.js
```diff
@@ -51,6 +51,8 @@ Soho.Locale.addCulture('sv-SE', {
   numbers: {
     percentSign: '%',
     percentFormat: '### %',
+    percentSuffix: ' %',
```
davidcarlsonberg: I don't think there should be a space between the value and the unit, i.e. the integer and the percentage sign.

tmcconechy: No this is actually correct.
1. didnt want to effect percentFormat in case its used so added percentSuffix 
2. There are cases where there is a space after and not depending on locale. some might be `100 %` some might be `100%` (and some are `%100` or `% 100`

PR #2850, opened 2019-09-26
2702 - Popupmenu: won't open after submenu structure exit

### LINKED ISSUE #2702: Popupmenu with submenus won't open
**Describe the bug**

**To Reproduce**
1. Go to ids-enterprise sample app
2. Click on popupdemo
3. Use this code for popup
```
    <ul soho-popupmenu id="action-popupmenu">
      <li soho-popupmenu-item><a soho-popupmenu-label>Cut</a></li>
      <li soho-popupmenu-item><a soho-popupmenu-label>Copy</a></li>
      <li soho-popupmenu-item><a soho-popupmenu-label>Paste</a></li>
      <li soho-popupmenu-item [subMenu]="true">
        <a soho-popupmenu-label>Paste Special</a>
        <ul>
          <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 1</a></li>
          <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 2</a></li>

          <li soho-popupmenu-item [subMenu]="true">
            <a soho-popupmenu-label>Paste Special</a>
            <ul>
              <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 1</a></li>
              <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 2</a></li>
              <li soho-popupmenu-item [subMenu]="true">
                <a soho-popupmenu-label>Paste Special</a>
                <ul>
                  <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 1</a></li>
                  <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 2</a></li>

                  <li soho-popupmenu-item [subMenu]="true">
                    <a soho-popupmenu-label>Paste Special</a>
                    <ul>
                      <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 1</a></li>
                      <li soho-popupmenu-item><a soho-popupmenu-label>Sub Menu 2</a></li>
                    </ul>
                  </li>
                </ul>
              </li>
            </ul>
          </li>
        </ul>
      </li>
      <li soho-popupmenu-separator></li>
      <li soho-popupmenu-item><a soho-popupmenu-label>Name and project range</a></li>
      <li soho-popupmenu-item [isDisabled]="isInsertCommentDisabled"><a soho-popupmenu-label id="x">Insert comment</a></li>
      <li soho-popupmenu-item [isDisabled]="isInsertNoteDisabled"><a soho-popupmenu-label>Insert note</a></li>
      <li soho-popupmenu-item><a soho-popupmenu-label >Clear notes</a></li>
      <li soho-popupmenu-separator singleSelectableSection="true"></li>
      <li soho-popupmenu-heading>Additional Options</li>
      <li soho-popupmenu-item isChecked="true" isSelectable="true"><a soho-popupmenu-label>Conditio nal formatting</a></li>
      <li soho-popupmenu-item isSelectable="true"><a soho-popupmenu-label>Data validation</a></li>
    </ul>
```
4. Move cursor to sub-menu (it will open) now move mouse outside menu up, then around the menu on the left, down and then back to nested menu -> it won't open

**Expected behavior**
Popupmenu submenus should open no matter the direction the cursor came from

**Version**
- ids-enterprise-ng: 5.5.1

**Platform**
 - Browser Name: Chrome

**Additional context**
see video
popup-bug

davidcarlsonberg:
Reentering the popupmenu from outside the popupmenu submenu structure would cause the entire structure to close. This PR change intercepts the exit and closes the submenus as is conventional on a number of visited sites.

Closes #2702 

- Pull branch, build, run app
- Visit http://localhost:4000/components/popupmenu/example-submenus.html
- Hover through the submenus until the last one is open
  - Exit the submenu to the body
  - Circle around the submenus and reenter
    - The submenus should not close upon reentry


src/components/popupmenu/popupmenu.js
```diff
@@ -1729,17 +1729,19 @@ PopupMenu.prototype = {
...
+          self.openSubmenu(menuitem);
+        }, 300);
 
-      $(document).on(`mousemove.popupmenu.${this.id}`, (documentE) => {
-        tracker = documentE.pageX;
-      });
+        $(document).on(`mousemove.popupmenu.${this.id}`, (documentE) => {
+          tracker = documentE.pageX;
+        });
+      };
```
davidcarlsonberg:
```suggestion
      }
```

src/components/popupmenu/popupmenu.js
```diff
@@ -1729,17 +1729,19 @@ PopupMenu.prototype = {
...
-      }, 300);
+        clearTimeout(timeout);
+        timeout = setTimeout(() => {
+          self.openSubmenu(menuitem);
+        }, 300);
 
-      $(document).on(`mousemove.popupmenu.${this.id}`, (documentE) => {
-        tracker = documentE.pageX;
-      });
+        $(document).on(`mousemove.popupmenu.${this.id}`, (documentE) => {
```
tmcconechy: I think we will need to make sure this new event goes in `detach` so the event is cleared to avoid a memory leak.

davidcarlsonberg: The issue you see is preexisting and I am only able to recreate it when zooming through the submenus. It's a symptom of the use of timeout. Maybe address as a follow up?

Not sure I understand. That's not a new event and it is cleared inside `mouseleave` just below. Do you still think it needs to go into `detach`?

tmcconechy: Ok i see it removes it self right away. I was checking in mousemove.

tmcconechy: But concerning the bug. Maybe have a look i was able to reproduce it very easily. Was seeing other wierd things like it still closing but couldnt figure out the steps. Up to you, but would make it perfect again

davidcarlsonberg:
A little more context:
This fix is based on the user removing their mouse from the popupmenu structure, circling around the structure, and then reentering; effectively passing the timeout threshold that allows a more forgiving diagonal navigation of the submenus (exiting and entering quickly between menus. One could argue that that user behavior is less than advisable. 
The original fix checked for reentry of the mouse into any submenu and if true, wouldn’t close the structure. This second fix prevents any possibility of that by closing the structure on any exit (perhaps mercilessly).
So, we could advise against the reported, and somewhat aberrant, user behavior, revert to the original fix, or indulge in a second timeout.

deep7102:
Yes, agree with @tmcconechy this feels like closing too aggressively.

EdwardCoyle:
Here's a video of me kinda playing around with it (not sure how helpful it is):

Oct-2-2019-9_39-AM

Compare this to the [behavior seen on `master` currently](http://master-enterprise.demo.design.infor.com/components/popupmenu/example-submenus.html).  I think we probably don't want the menu closing immediately upon the mouse leaving.  Most context menus I've seen in other applications need a click or tap to close, and don't even adhere to what we had before (submenus closing on a timer).

My preference would be to opt for something that goes aggressively in the other direction, only doing the closing of a submenu when:

- another higher-level option is highlighted for a period of time.
- the user purposefully clicks/taps out of the whole menu structure.

I'm not sure how feasible that would be with the time constraints, but if we can at least revert back to the original fix, and maybe revisit this in another sprint, that would be better.

PR #2868, opened 2019-10-01
2434  - Appmenu mobile scroll issues

### LINKED ISSUE #2434: App Menu: Scroll issues on mobile
**Describe the bug**
Open app menu on mobile breakpoints accept dragging/scrolling within body which causes jumping.

**To Reproduce**
1. On a small touch device or simulator, go to http://master-enterprise.demo.design.infor.com/components/applicationmenu/example-index.html
2. Open the application menu
3. Scroll down on the application container section 

**Expected behavior**
The App menu in this state is essentially a modal so it should block input to the container behind it.

**Version**
- ids-enterprise: all versions

**Screenshots**

Jun-24-2019 17-16-00

Jun-24-2019 17-25-27

jmacaluso711:
This PR disables scroll bounce on touch devices when the app menu is open.

Closes #2434 

* Pull Branch
* In a mobile device simulator go to http://localhost:4000/components/applicationmenu/example-personalized-roles.html


deep7102:
On simulator iPhone 6 Plus with IOS 8.1, do not open
http://localhost:4000/components/applicationmenu/test-personalized-role-switcher-long-title.html

image

tmcconechy:
I was wondering if using `vh` would be a problem with browser support but seems supported on IE11 ect. We havent used it much.

jmacaluso711:
@tmcconechy should be ok here as it's just for touch devices.

tmcconechy:
Yeah gotcha @jmacaluso711 and I saw it was supported but was wondering if maybe @deep7102 issue was related? We can limit support if we have to but its not ideal if we can avoid if thats the case?

jmacaluso711:
Yeah, I'm not able to install IOS 8.1 in my xcode simulator, so haven't been able to test

tmcconechy:
In addition to @deep7102 's comments i found that if i run this (iphoneX 11) that when the menu closes the title disappears. We should probably address that as well.

Screen Shot 2019-10-02 at 8 53 28 AM

PR #2877, opened 2019-10-02
2873 - Add code to keep the current locale

### LINKED ISSUE #2873: Contextual Action Panel: Locale reverts back to the default after opening a contextual action panel
**Describe the bug**
When opening a contextual action panel the locale reverts back to the default (en-US). We would like this to be fixed as soon as possible to be able to support separate language and locale in the Homepages application.

Depending on the settings for the current user our application may use different language and locale which are both different from the default. This works as expected, numbers and dates are formatted according to the current locale etc. After opening any contextual action panel the locale reverts back to en-US and after that all numbers and dates are formatted in en-US instead of the expected locale.

When the contextual action panel is opened the Initialize function is invoked (see behaviors\initialize\initialize.js). The settings parameter is empty so it defaults to the en-US locale set in INITIALIZE_DEFAULTS.

**To Reproduce**
1. Test with an application configured to use non-default language and locale
2. Open a contextual action panel
3. The current locale reverts to en-US

**Expected behavior**
The current locale should remain after opening a contexual action panel.

**Version**
4.21.0

**Platform**
 - Affects all platforms

tmcconechy:

When initializing the CAP it will call initialize() internally and this changes the current locale.
Made a fix to keep the active locale unless one is passed in and a test.

Fixes #2873 

- open http://localhost:4000/components/contextualactionpanel/test-locale.html
- it will show labels and locale info in the text field in English
- open http://localhost:4000/components/contextualactionpanel/test-locale.html?locale=de-DE
- it will show labels and locale info in the text field in German
- try any other locale you want


EdwardCoyle:
Note that I also tested @SofiK's case on http://localhost:4000/components/contextualactionpanel/example-index.html?locale=ar-sa, and it seemed to work for me.

PR #2881, opened 2019-10-02
2824 - Vertical Tabs: .alternate overrides .is-personalizable

### LINKED ISSUE #2824: Vertical Tabs: When in .is-in-responsive-mode and .is-personalizable then don't apply the .alternate class
At the moment `.alternate` is being applied in tabs.js when tabs are vertical whether responsive or not. This causes the `.is-personalisable` background theme color to be overridden by the `.alternate` class. 

Vertical tabs should participate in `.is-personalisable` (and use the theme background color) when in responsive mode (`.is-in-responsive-mode`). 

For an example of the mark that could be used see

http://master-enterprise.demo.design.infor.com/components/tabs-vertical/example-responsive.html

Add `is-personalizable` to the tab-container. 
      <div id="tabs-vertical" class="vertical tab-container no-scroll `is-personalizable`" data-options='{ "verticalResponsive": true }'>

The result should look like it does now when tabs are vertical
image

 but like the screenshot here when they are responsive 

image

davidcarlsonberg:
The `alternate` class overrides the `is-personalizable` class on `tabs-vertical`. A feature request has been made to allow vertical tabs to be participate in the theming allowed by `.is-personalizable`. This PR checks for `is-personalizable` before adding the `alternate` class.

Closes #2824 

- Pull branch, build, run app
- Visit http://localhost:4000/components/tabs-vertical/example-responsive.html
  - `.is-personalizable` is now in the markup and should prevent `.alternate` from applying to `.tabs-vertical`. Alternatively, if `.is-personalizable` is not present, `.alternate` will apply for `.is-in-responsive-mode`.


docs/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # What's New with Enterprise
 
+## v4.23.0
```
tmcconechy: Do we want to just merge this in today and have it in 4.22. That should be fine as the issue is small

tmcconechy:
Just need to update this one Visual regression test https://travis-ci.com/infor-design/enterprise/jobs/241250638#L2254 as one would expect

Issue #2800, opened 2019-09-16
Menu Component: Tooltip cut off on Mobile


CindyMercadoReyes:
**Describe the bug**
Menu Button Component: On the sink page all the drop down menu's are cut off or not appearing

**To Reproduce**
1. Go to url: http://4210-rc0-enterprise.demo.design.infor.com/components/toolbar/example-inside-form-tag.html
2. Press down on the three button "More" icon
3. The tooltip is cut off 
4. Please see screen shot attached. 
5. try both the menu on the header at the top and on the toolbar. they are both doing it

**Expected behavior**
The tooltip/drop down menu should appear when the user clicks on the more icon on their device

**Version**

- ids-enterprise: [e.g. v4.20.0 ]

**Screenshots**
![IMG_6899](https://user-images.githubusercontent.com/39101346/64983239-132a3680-d88e-11e9-8494-90ae1d6aee23.PNG)
![IMG_6900](https://user-images.githubusercontent.com/39101346/64985309-951c5e80-d892-11e9-9dc7-6f51f4fa0f0f.PNG)

**Platform**
 - Device (if applicable) [e.g. iPhone XS 12]

**Additional context**
None

tmcconechy:
One idea for this is to make that popup use `attachToBody` setting. This may work?

tmcconechy:
Looks like here https://github.com/infor-design/enterprise/blob/master/src/components/toolbar/toolbar.js#L245

1. we could add `attachToBody: true` to solve but only on condition of `(env.os.name === 'ios'` to be safer.

EdwardCoyle:
Per a chat with @pwpatton, I'd like to make a bit more of a comprehensive fix for this problem.   It seems that there's really no better way to handle this than to just default `attachToBody: true` in an iOS setting.  This should probably be done on all Popupmenus, instead of at the Toolbar level.  FX @ericangeles's change is great, but it doesn't work for Flex Toolbars.  I'd like to take this fix and push it down to the Popupmenu level instead.

johnbarcena:
tested on http://master-enterprise.demo.design.infor.com/components/toolbar/example-inside-form-tag.html
Test PASSED

johnbarcena:
However, menus' for more actions are not working and the color sub menu is not available.
see ticket #2986

tmcconechy:
We'll take a look at #2986

PR #2901, opened 2019-10-07
2800 - Toolbar: Fix toolbar/dropdown menu cut off on iOS device

ericangeles:
This PR fixes a problem with dropdown/toolbar menu on iOS device (Safari & Chrome browser) is being cut off when clicking the more button. Adding `attachToBody: true` if the environment is **iOS**.

Closes https://github.com/infor-design/enterprise/issues/2800

- Pull this branch, then run the app
- Use iOS device or browserstack (Safari and Chrome )
- Navigate to http://localhost:4000/components/toolbar/example-inside-form-tag.html
- Click `...` button
- toolbar menu shouldn't cut off

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## v4.22.0
 
 ### v4.23.0 Fixes
-- `[Toolbar]` Fixed a bug where the dropdown/toolbar menu is being cutoff in iOS device. ([#2800](https://github.com/infor-design/enterprise/issues/2800))
+
+- `[Toolbar]` Fixed a bug where the dropdown/toolbar menu is being cut off on iOS device. ([#2800](https://github.com/infor-design/enterprise/issues/2800))
 
 ### v4.22.0 Deprecation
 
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## v4.22.0
```
tmcconechy: I think we need to move the 4.22.0 down and add a new 4.23 section. Cant do that with a pull request suggestion so you will have to manually fix that.

src/components/toolbar/toolbar.js
```diff
@@ -242,9 +243,11 @@ Toolbar.prototype = {
 
     // Setup an Event Listener that will refresh the contents of the More Actions
     // Menu's items each time the menu is opened.
+    const isMobile = env.os.name === 'ios';
     const menuButtonSettings = utils.extend({}, this.settings.moreMenuSettings, {
       trigger: 'click',
-      menu: this.moreMenu
+      menu: this.moreMenu,
+      attachToBody: isMobile
```
tmcconechy: Does android have this issue? We may just need this for all safari.

I am seeing on mac safari the toolbar "shifts up" and this may fix that. So perhaps do this for isSafari? `env.browser.name === 'safari'`

PR #2921, opened 2019-10-08
NA - Hierarchy card border colors

jmacaluso711:
Corrects the panel border color on the hierarchy cards in the uplift theme

related to design-system ([#423](https://github.com/infor-design/design-system/issues/423))

- pull branch
- go to http://localhost:4000/components/hierarchy/example-stacked.html
- the border color should now be `$theme-color-palette-graphite-30` (#D1D1D1)


jamie-norman:
I think a subtle box shadow on the hierarchy cards would be good. I'm in.

kevinwhitedesign:
I see this as a little problematic~ I think part the uplift strategy was to get away from the borders and use more shadows, correct me if I'm wrong @kentonquatman @elizabethhartley @danirosefishinfor

kevinwhitedesign:
side note: we can keep borders for high contrast

tmcconechy:
But the cards do have a box shadow http://master-enterprise.demo.design.infor.com/components/cards/example-index.html?theme=uplift&variant=light&colors=0563C2 (sort of with border)? Im just suggesting maybe the hierarchy panels should be the same?

kevinwhitedesign:
Yeah they do have shadows, that's fine. I think removal of borders is ideal for uplift cards, widgets, panels etc. As that's how it was defined. If more contrast is needed, we can make the shadow more dense. However, because they are not needed to pass any accessibility I think it was ok as is.

kentonquatman:
I'm going to advocate for using borders with a drop shadow in the Uplift theme. I I mocked-up a few treatments, and it looks washed out using just shadows and it looks busy when the shadows are darker. (Note: I'm on a Dell monitor so shadows might be rendering lighter for me in general).

### Just Shadows:
Just Shadows
### Darker Shadows:
Dark Shadows
### Border + Shadows:
Border + Shadow

kevinwhitedesign:
Misheard, I thought we were looking at this: http://master-enterprise.demo.design.infor.com/components/cards/example-index.html?theme=uplift&variant=light&colors=0563C2 but Jamie's feedback was referring to this: https://design.infor.com/code/ids-enterprise/latest/demo/components/hierarchy/example-index?theme=uplift&variant=light&colors=0563C2 I think same borderless thinking could apply here. Check out this example from the wild (microsoft planner): https://www.dropbox.com/s/vnwr1imtwu813mg/Screenshot%202019-10-09%2013.33.48.png?dl=0

jmacaluso711:
@kevinwhitedesign Making them consistent with cards for now.

PR #2923, opened 2019-10-08
2809 - Fix bubbling destroy event with popupmenu

### LINKED ISSUE #2809: Popupmenu: Destroy event is bubbling up to other controls since not using triggerHandler
**Describe the bug**
In the destroy function for the Popupmenu it is triggering a 'destroy' event on the menu element which is causing the event to bubble up to parent elements and cause bad problems in our app. I believe Soho's normal way is to use "triggerHandler" instead of "trigger" so that the event does not bubble up.

image

**To Reproduce**
I found this bug in the CSD application where we had a colorpicker in a CAP modal. After using the colorpicker its popupmenu would close (be destroyed) and cause a destroy event to be triggered on the CAP modal. You may be able to reproduce the bug at this example if you add a listener on the modal for the 'destroy' event -> http://master-enterprise.demo.design.infor.com/components/colorpicker/test-modal.html

**Expected behavior**
Soho code should use "triggerHandler" instead of "trigger" to prevent event bubbling.

**Version**
v4.22.0

**Platform**
- Device: Dell Laptop
- OS Version: Windows 10
- Browser Name: Chrome
- Browser Version: 76.0.3809.132 (Official Build) (64-bit)

deep7102:
Fixed bubbling destroy event with popupmenu.

Closes #2809

- Navigate to http://localhost:4000/components/colorpicker/test-modal.html
- Open developer tools
- Open modal
- Open color picker popup menu and pick any color
- Color picker popup menu should close
- See console should not fire destroy event

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### v4.23.0 Fixes
 
 - `[Contextual Action Panel]` Fixed an issue where the CAP close but beforeclose event not fired. ([#2826](https://github.com/infor-design/enterprise/issues/2826))
-- `[Popupmenu]` Fixed an issue where the destroy event was bubbling. ([#2809](https://github.com/infor-design/enterprise/issues/2809))
+- `[Popupmenu]` Fixed an issue where the destroy event was bubbling up to other parent components. ([#2809](https://github.com/infor-design/enterprise/issues/2809))
 
 ### v4.23.0 Chores & Maintenance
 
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### v4.23.0 Fixes
 
 - `[Contextual Action Panel]` Fixed an issue where the CAP close but beforeclose event not fired. ([#2826](https://github.com/infor-design/enterprise/issues/2826))
+- `[Popupmenu]` Fixed an issue where the destroy event was bubbling. ([#2809](https://github.com/infor-design/enterprise/issues/2809))
```
```suggestion
- `[Popupmenu]` Fixed an issue where the destroy event was bubbling up to other parent components. ([#2809](https://github.com/infor-design/enterprise/issues/2809))
```

Issue #2964, opened 2019-10-11
Message : unexpected scrollbar appears even if its not needed. 


janahintal:
**Describe the bug**
Message : unexpected scroll appears even if its not needed. 

**To Reproduce**
1. Go to http://4220-rc0-enterprise.demo.design.infor.com/?theme=uplift&variant=dark&colors=0053B1# and minimize the view to responsive 
2. Go to Message components 
3. Click Success Example button 

**Expected behavior**
Modal should not have a scrollbar

**Version**
RC 4.22 Regression

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/66627954-e28cb100-ec2f-11e9-8c0c-c2af58c1202a.png)

**Platform**
All Browser, OS and Devices

**Additional context**
N/a

tmcconechy:
On review of this issue Im closing #2964 as wont fix as you will need to scroll if the text is much bigger. So it is actually working correctly its just that the text is just too much to fit with shorter break points.

So really it's working ok.

tmcconechy:
I was still able to reproduce this..

- load Windows Chrome (78)
- go to http://localhost:4000/components/message/example-index.html
- click the success example button (notice the scroll bar)

CindyMercadoReyes:
This is still appearing for Win 10 Edge 18. Please see screen shot attached. 
![bs_win10_Edge_18 0](https://user-images.githubusercontent.com/39101346/69166734-b8959c80-0ac1-11ea-9d6e-e610794f05a0.jpg)

janahintal:
This issue is now resolved and now working in http://master-enterprise.demo.design.infor.com/
![image](https://user-images.githubusercontent.com/45757016/69526124-73cba300-0fa4-11ea-8e26-cc00d2f2ceb7.png)

PR #2971, opened 2019-10-14
2964 - Message: Fix unexpected scrollbar

ericangeles:
This PR fixes unexpected scrollbar appears in message component on modal.

Closes https://github.com/infor-design/enterprise/issues/2964

- Pull this branch, build, run app
- Navigate to http://localhost:4000/components/message/example-index.html
- Click Success Example button
- Scrollbar shouldn't visible in the body content

You can also test the demo app
- Go to Message Component of demo app (http://localhost:4000/)
- Click Success Example button
- Scrollbar shouldn't visible in the body content


src/components/modal/_modal.scss
```diff
@@ -323,7 +323,7 @@
   }
 
   .modal-body-wrapper {
-    overflow: auto;
+    overflow: hidden;
```
deep7102: Hi @ericangeles this fix will stop user to scroll any content.

EdwardCoyle: Yeah unfortunately, the modal body area needs to be scrollable.  It's primarily for larger forms (think Contextual Action Panel), or for responsive settings where screen size might be smaller than the modal itself.

tmcconechy:
Im thinking maybe we should just close #2964 as wont fix as you will need to scroll if the text is much bigger. So it is actually working correctly.

In this example if i click error after the padding change i still see a scroll bar in chrome - on mac with visible scrollbars set. So no matter what padding we use we could still see the issue depending on if the text is bigger or the browser is shorter. So really its working ok.

I vote we close this PR and close 2964 as a non fix.

Screen Shot 2019-10-15 at 4 17 51 PM

Thumbs up if you agree 👍

PR #2984, opened 2019-10-15
2834 - Adjust placement of generated (x) button on Contextual Action Panels

### LINKED ISSUE #2834: Contextual Action Panel: Close button
**Describe the bug**
A close button is added to the contextual action panel when showCloseButton is set. But the close button is added to the modal header and not the button set of the toolbar as in the demo used to test for issue #1943.
The close button actually wraps around and appears below the header. Also, the text 'Close' is partially visible.

**To Reproduce**
In any contextual action panel sample,
(1) set showCloseButton to true
(2) add four buttons so an overflow button is displayed in the toolbar
(3) Bring up the contextual action panel

**Expected behavior**
The close button is on the top right corner of the contextual action panel. It is just an icon button with no text.

**Version**

- ids-enterprise:  v4.22.0-dev

**Screenshots**

**Platform**
 - OS Version: Windows 10
 - Browser Name Chrome
 - Browser Version Version 77.0.3865.90

**Additional context**

EdwardCoyle:
This PR normalizes the placement of `(x)` Close buttons on Contextual Action Panels when configuring with the inner Modal setting `showCloseBtn`.  Previously, there was a bug that would incorrectly display this generated button inside the content area.

This PR also:
- Ensures that the generated button's text is translated via Locale.
- Adds several CAP tests demonstrating the addition of a close button in various configurations, including while using a Flex Toolbar, defining via settings, and defining via prebuilt HTML markup.

Closes #2834 

Pull this branch, build, and run the demoapp.  Open all four of the following test pages.  Each Contextual Action Panel should have a close button defined in their toolbars.  In the standard toolbar, the `(x)` button spills into the overflow menu.  In the Flex Toolbars, the `(x)` button is separate from the buttonset and should be displayed directly in the toolbar:

- http://localhost:4000/components/contextualactionpanel/test-close-button?theme=uplift
- http://localhost:4000/components/contextualactionpanel/test-close-button-from-settings?theme=uplift
- http://localhost:4000/components/contextualactionpanel/test-close-button-flex?theme=uplift
- http://localhost:4000/components/contextualactionpanel/test-close-button-flex-from-settings?theme=uplift

Modals should also be tested for accuracy:

- http://localhost:4000/components/modal/example-close-btn?theme=uplift

----

paging @aaralikar @Sovia


tmcconechy:
cc @aaralikar

ericangeles:
Just noticed when resizing the modal, it tends to break the buttons in the header section. Not sure if we have a settings for that, or a css rules to display only the icons when hitting the tablet and/or mobile viewport.

See here:
http://localhost:4000/components/contextualactionpanel/test-close-button-flex?theme=uplift

I also echoed @tmcconechy's comment on modal's icon position.

deep7102:
I see same as @tmcconechy and @ericangeles

EdwardCoyle:
@tmcconechy and @deep7102: 
> Does the x on modal feel like its not placed properly with the text on the title?

It's never been aligned to the text on the header when it's just the Modal.  This is the Soho version:

Screen Shot 2019-10-16 at 4 33 09 PM

I think that its placement a byproduct of this problem though:

> does the close x feel bigger?

Yeah it's definitely bigger in Uplift.  I didn't change any sizes for the button icon itself in this PR.  The size difference was a direction we had from the design team in the past... although I'm not sure what @kevinwhitedesign and @kentonquatman would say now?  The Soho icon was much smaller, so I think it would make sense if this one is also.

@ericangeles:
> Just noticed when resizing the modal, it tends to break the buttons in the header section. Not sure if we have a settings for that, or a css rules to display only the icons when hitting the tablet and/or mobile viewport.

We don't have any settings for that.  It's an issue that's always been present with the Flex Toolbar, where because it's more manually defined, if there's no more menu button present, the buttons won't spill over and will just end up cut off.  It's also an issue of maybe building this sample without correctly setting a size on the CAP (which is something we ask the app developers to do with CAPs). I could probably fix this in the sample by just adding a more menu, slimming down the number of buttons, or making the CAP wider.

tmcconechy:
Yeah we should probably check with design on the X placement but i imagine would be similar to a card with toolbars. I realize this was like that before but this brought it more to my attention. If we want to fix it later thats cool.

Same with the difference in the X icon size. i agree maybe make them the same with css or maybe its a sketch issue?

PR #2985, opened 2019-10-15
2982 - Fix the language from resetting [v4.22.x] [v4.21.x]

### LINKED ISSUE #2982: Timepicker/popover: Language is reset to value of Locale
**Describe the bug**
We are setting language and locale separately, as described here:
https://design.infor.com/code/ids-enterprise/latest/locale#code-example---using-a-different-language-and-locale

When opening a popover / timepicker, the language is changed to the value of locale (possibly due to initialize() being called for sub-content).

Pseudo Example: 
Locale.set("fi-FI"); // Finnish
Locale.setLanguage("sv"); // Swedish

In attached GIF below, it's seen that
1. Initially, language is Swedish in the components (in the GIF, "Obligatoriskt" is shown for the required input field)
2. Open timepicker. "Aseta aika" is the button text, which is Finnish 
3. After this, all soho controls use Finnish translations instead of Swedish
lang-issue

We also noticed that some texts are always translated based on Locale, and not Language, in some components. Example below is the datepicker (the issue is not due to popover/timepicker being opened, it's like this from the start)
image

**Expected behavior**
If using Locale.setLanguage(), that language should be used for translations in all IDS components, and not be changed as a result of automatic plugin initializations. 

**Version**
- ids-enterprise: 4.21.2

tmcconechy:

When testing #2982 i noticed some aspects of the way #462 where done did not entirely work correctly with the language. And the language could not actually be seperate from the locale. Added some tests and fixed this problem. Also fixed an issue in validation on time in single digit 24 hr locales. Will test this pr in 4.22 and then merge back to 4.21.x by request

Fixes #2982 

- go to http://localhost:4000/components/timepicker/example-index.html?locale=fi-FI
- open the picker and notice the set time button is in finish
- hit the button to select the current time and notice its 1.00 as is the format in fi-FI
- in the console type `Locale.setLanguage('sv')` 
- open the picker and notice the set time button is in swedish
- hit the button to select the current time and notice its 1.00 as the main format is still fi-FI
- compare the format to http://localhost:4000/components/timepicker/example-index.html?locale=sv-SE 
- also test some pages for each component added for the cases of a different locale and lang.
http://localhost:4000/components/timepicker/test-specific-locale.html
http://localhost:4000/components/datepicker/test-two-locales-same-page.html
http://localhost:4000/components/monthview/test-specific-locale.html
http://localhost:4000/components/monthview/test-specific-locale-lang.html
http://localhost:4000/components/calendar/test-specific-locale.html
http://localhost:4000/components/calendar/test-specific-locale-lang.html


tmcconechy:
cc @volante007

PR #3000, opened 2019-10-18
2899 - Fix shrinking of context menus on small viewports

### LINKED ISSUE #2899: Context menu sizing and positioning in some situations is incorrect.
**Describe the bug**
The size of the browser window can effect the sizing and positioning of the context window. Sometime even a small change in the browser window size can make the context window come up an huge differences in size. Left positioning seems to be right but the top positioning is not always correct either.  This issue for us is seen in the right click full page.

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/contextmenu/example-page-rightclick.html
2. Size browser height smaller 
3. Right click to show context menu
4. Keep repeating step 2-3 and each time making the window small to see issue. In the case if the example the menu height is not to tall and so it requires a smaller window to replicate. If you have a taller context menu the behavior cn be seen on full size browser windows.

**Expected behavior**
For the context menu to size and position properly taking screen size and menu dimensions and use those to properly size the menu

**Version**
 IDS Enterprise Components - v4.22.0-dev

**Screenshots**

Screenshot 2019-10-04 13 03 29
**This is clicking near the bottom of the screen with the menu fitting**

Screenshot 2019-10-04 13 03 41
**Made the window smaller. This is clicking near the bottom but the menu comes up on the top and to short**

Screenshot 2019-10-04 13 04 02
**Made the windows smaller. this is clicking near the bottom and now the menu is so small it seems unusable and nowhere near the click area.**

Screenshot 2019-10-04 13 04 10

**Made it even smaller but now the menu sizes to large with no way to get to the lower menu items.**

**### Again this issue does not just happen in smaller window sizes. It can happen in  full window browsers when the menu has a lot of items in it.**

**Platform**
Chrome - Version 77.0.3865.90 (Official Build) (64-bit)

**Additional context**

EdwardCoyle:
This PR makes a few changes to the routine that handles the "shrinking" of elements on the page positioned by our `Place` behavior.  In some cases (specifically context menus), placing the element in a viewport that would cause it to shrink would incorrectly cut the size of the element by more than half.

Closes #2899

- Pull this branch, build, run app
- Open http://localhost:4000/components/contextmenu/example-page-rightclick.html
- Shrink the viewport to a size that would cut off the context menu (I used 640x296 in the e2e test)
- Right click anywhere in the page to open the context menu.
- Examine the menu's position to make sure it is not bleeding off of any edge, and that it fills the available viewport space as much as possible.  Due to the container offsets, there will be approximately 15px between the actual viewport "edge" and the menu's edge.

Screen Shot 2019-10-18 at 11 11 28 AM


tmcconechy:
Seeing one thing that might be worth looking at:
1. http://localhost:4000/components/contextmenu/example-field-options.html
2. make the page a height so that input full is just barely in view
3. open the menu

Seems like it is leaving space on the top but the arrow side is too far down. It didn't do this before

**Now**
Screen Shot 2019-10-18 at 1 43 26 PM

**Master**
Screen Shot 2019-10-18 at 1 45 59 PM

Also in both of them the menu could be pushed a few pixels over as the scroll bar is a bit cut off. Maybe 5px padding?

EdwardCoyle:
The field options sample seems to have been doing that for some time - I tested it back to http://4150-enterprise.demo.design.infor.com/components/contextmenu/example-field-options.html.

The part where it cuts down the sizing on top is due to the fact that the containment on field options menus is set to the scrollable area, so it won't bleed into the header even if there appears to be space.  If we want to fix that we have to change the field options menus as a whole (and basically any popupmenu button).

As for the menu being close to the right edge, that's something that should be fixed.

tmcconechy:
OK, yeah i see. It doesn't have to bleed into the header then. But hopefully the down arrow is in the same position as before if possible?

Issue #1384, opened 2018-12-12
Spinbox: Fix field sizes on mobile


EdwardCoyle:
Originally reported by @tmcconechy [in this comment](https://github.com/infor-design/enterprise/pull/1362#issuecomment-446269944).  When placing two Spinbox fields side-by-side in a mobile layout, they should not be wider than a single input field.  

**Describe the solution you'd like**
The CSS for the Spinbox and standard input field should be modified to match well on a mobile layout. 

As of `v4.13.x`, This is what http://master-enterprise.demo.design.infor.com/components/form/test-spinbox-compound-field.html currently looks like on iOS:

<img width="256" alt="screen shot 2018-12-12 at 11 40 17 am" src="https://user-images.githubusercontent.com/3249601/49884376-b8e1de00-fe02-11e8-9320-c3217c8f8f63.png">

tmcconechy:
Just a note but i think its the input sizes not the spinbox sizes? They line up so there is 20px on each size but the input is smaller.

EdwardCoyle:
Specced this issue out a bit further [in this comment](https://github.com/infor-design/enterprise/pull/3044#issuecomment-547458890)

jbrcna:
QA **FAILED**

Verified in http://4240-beta0-enterprise.demo.design.infor.com/components/spinbox/example-sizes.html

![image](https://user-images.githubusercontent.com/56722184/70689699-e69c8400-1cef-11ea-8b10-76de40ae7219.png)

janahintal:
QA Passed. Will now move to Done. 
Verified in http://4240-rc0-enterprise.demo.design.infor.com/components/spinbox/example-sizes.html?theme=uplift&variant=light across all platform.

PR #3044, opened 2019-10-29
1384 - Form: Fix field sizes of spinbox on mobile

ericangeles:
This will fix the fied sizes of spinbox on mobile. Added a custom container for spinbox and add width for the spinbox wrapper.

Closes https://github.com/infor-design/enterprise/issues/1384

- Navigate to http://localhost:4000/components/form/test-spinbox-compound-field.html
- Test it on mobile device or use browserstack
- `Age2` and `Age3` spinbox should align in a single column and matches the width of the input fields

### DIFF
--- a/app/views/components/form/test-spinbox-compound-field.html
+++ b/app/views/components/form/test-spinbox-compound-field.html
@@ -8,13 +8,14 @@
       </div>
 
       <div class="field">
-        <label for="test-spinbox-age2">Age2</label>
-        <input id="test-spinbox-age2" name="test-spinbox-age2" type="text" class="spinbox" />
-      </div>
-
-      <div class="field">
-        <label for="test-spinbox-age3">Age3</label>
-        <input id="test-spinbox-age3" name="test-spinbox-age3" type="text" class="spinbox" />
+        <div class="spinbox-field-container">
+          <label for="test-spinbox-age2">Age2</label>
+          <input id="test-spinbox-age2" name="test-spinbox-age2" type="text" class="spinbox" />
+        </div>
+        <div class="spinbox-field-container">
+          <label for="test-spinbox-age3">Age3</label>
+          <input id="test-spinbox-age3" name="test-spinbox-age3" type="text" class="spinbox" />
+        </div>
       </div>
 
       <div class="field">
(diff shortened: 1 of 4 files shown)


EdwardCoyle:
@ericangeles, I agree with @tmcconechy that we probably shouldn't fit in a custom size for the field just to stack them in this manner.  Ideally, if a developer is implementing two spinboxes next to each other like the example, they're doing it with grid columns that will respond to the space available, so ideally the spinbox would fit in cleanly.

First off, I think the notion that all four of the fields on http://localhost:4000/components/form/test-spinbox-compound-field.html are inside a compound field is just not realistic.  If the expectation is that we are going to stack form fields cleanly, we should just change the example into a row/column grid instead.  I'd say change that first.

Secondly, there's a scenario for responsive forms, where the fixed width of these fields should become fluid, taking up the full width of their container:

http://localhost:4000/components/form/example-inputs.html

We should probably extend this functionality to spinbox fields, because I believe that currently the spinbox only has fixed-width sizes.

Additionally, when it comes to the fixed-width fields, we actually have a few "standard" sizes defined for other form field types that you should use for comparison when adjusting the spinbox sizes:

- http://localhost:4000/components/input/example-sizes.html
- http://localhost:4000/components/dropdown/test-sizes.html

The [CSS for the standard inputs exists here](https://github.com/infor-design/enterprise/blob/master/src/components/input/_input.scss#L518-L553).  The difficulty with spinbox is that it needs to adjust that standard input size to account for the buttons on either side of it. You could probably do that with a CSS rule that qualifies inputs that exist inside the `.spinbox-wrapper`.

So I think the bullet points we need to address are:
- [ ] Change the example to the standard IDS row/column grid instead of expecting all four fields to cleanly stack inside a single compound field.
- [ ] Implement fluid sizing for spinbox fields
- [ ] Implement standard sizes for spinbox fields that properly line up against our other input field types (may also require a good test page to demonstrate them side-by-side)

Apologies that this issue wasn't specced out fully before.  I think the things I've listed above would fix the issue in a more "future-proof" way.

tmcconechy:
@ericangeles Do you have ability to fix this up for this release (EOD tomorrow?)
Or maybe we should close this PR off for later?

ericangeles:
> @ericangeles Do you have ability to fix this up for this release (EOD tomorrow?)
> Or maybe we should close this PR off for later?

@tmcconechy maybe we should close this PR as we revisiting the widgets for XM. Will close this ticket. Thanks!

PR #3045, opened 2019-10-29
650 - Fix xss issues in dropdown

tmcconechy:

The xss was not working when doing a backspace in dropdown. (There were CSP errors).

Fixes infor-design/enterprise-ng#650

**Test One**
- pull this and restart the demo app
- go to http://localhost:4000/components/dropdown/test-xss.html
- type a and make sure it filters
- make sure you can select the <> formatted items
- backspace to clear
- make sure no console errors (CSP Error will happen if it injects)

**Test Two**
- go to http://localhost:4000/components/dropdown/test-special-chars.html
- make sure you can select all the items especially `<test>`

**Test Three**
- go to http://localhost:4000
- check the console and make sure there are no CSP errors

**Test Four**
- go to http://localhost:4000/components/dropdown/example-icons.html
- type anything to search
- backspace to clear
- ensure this example is working

cc @anhallbe

### DIFF
--- a/app/src/js/middleware/csp-handler.js
+++ b/app/src/js/middleware/csp-handler.js
@@ -9,6 +9,7 @@ function addNonceToScript(html, nonce) {
   nonced = nonced.replace(/&lt;script/ig, '<scriptno-nonce-lt');
   nonced = nonced.replace(/<script>/ig, `<script nonce="${nonce}">`);
   nonced = nonced.replace(/<script id/ig, `<script nonce="${nonce}" id`);
+  nonced = nonced.replace(/<script src/ig, `<script nonce="${nonce}" src`);
   nonced = nonced.replace(/<scriptno-nonce>/ig, '<script>');
   nonced = nonced.replace(/<scriptno-nonce-lt/ig, '&lt;script');
   return nonced;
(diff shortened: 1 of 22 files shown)


EdwardCoyle:
All the tests appear to work just fine, but I noticed that if you start the demoapp with `npm run quckstart:livereload`, the inclusion of the Livereload script causes a CSP error:

Screen Shot 2019-10-29 at 12 47 35 PM

Maybe `script-src 'self'` doesn't include localhost?

tmcconechy:
ok @EdwardCoyle i've fixed liverload. good catch

deep7102:
Looks good, few things saw as
http://localhost:4000/components/dropdown/test-xss.html
- select item with <>, see the tooltip is show as inner node text only
- if you type`t` to filter, it convert `<` to `&lt;`

http://localhost:4000/components/dropdown/example-icons.html
- type a and see the swatches not show with filter list

tmcconechy:
@EdwardCoyle this change is big enough that it should be retested fully.

EdwardCoyle:
~Found an issue in the Dropdown filtering regarding the icon, but I think we address it in a separate issue.  Note the icon change when text is entered/removed from the filter:~

Untitled_-Nov-5-2019-9_50-AM

I just realized this is probably by design because "no matching result" wouldn't have an icon associated with it.

tmcconechy:
Yes, at the time i told @ericangeles to use that icon as the "empty" icon or the text was mismatched. I think it works ok

PR #3142, opened 2019-11-18
3019 - Fix single digit dates

### LINKED ISSUE #3019: Locale: Cannot parse datetime string using the Finish locale (fi-FI)
The function Soho.Locale.parseDate returns undefined when trying to parse a datetime string on the Finish locale (fi-FI) using the datetime format "d.M.y H.mm" defined in the IDS culture file fi-FI.js. The datetime format is retrieved from Soho.Locale.calendar().dateFormat.datetime.

The parseDate function works when testing with a four digit year in the format ("d.M.yyyy H.mm") but it does not work with the datetime in the culture file that has a single character for year. I don't know if this is an issue in the parseDate implementation or if the configuration of the datetime format is wrong in the Finish culture file.

The function works as expected for other locales that we have tested with but we have not tested with all locales.

**To Reproduce**
1. Call the parseDate function using a valid Finish datetime string.

Non working examples:
Soho.Locale.parseDate("18.10.2019 7.15", Soho.Locale.calendar().dateFormat.datetime, true)
Soho.Locale.parseDate("18.10.2019 7.15", "d.M.y H.mm", true)

Working example (note that the short date format is defined to use four digit year "d.M.yyyy"):
Soho.Locale.parseDate("18.10.2019 7.15", "d.M.yyyy H.mm", true)
Soho.Locale.parseDate("18.10.2019", Soho.Locale.calendar().dateFormat.short, true)

**Expected behavior**
A Date object is returned from the parseDate function when called with a valid datetime string and datetime format.

**Actual behavior**
The value undefined is returned.

**Version**
4.21.3

tmcconechy:

Some strings for dates in the fi-FI locale incorrectly had a one digit year (which doesnt work).
This fixes the issue

Fixes #3019 

- this is covered entirely by tests

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,8 +13,8 @@
 - `[Datepicker]` Fixed missing background color on disable dates and adjusted the colors in all themes. ([#2910](https://github.com/infor-design/enterprise/issues/2910))
 - `[Datepicker]` Fixed a layout issue on the focus state on colored/legend days. ([#2910](https://github.com/infor-design/enterprise/issues/2910))
 - `[Dropdown]` Fix a bug where a dropdown in a datagrid cell would sometimes not display the correct value when selected. ([#2919](https://github.com/infor-design/enterprise/issues/2919))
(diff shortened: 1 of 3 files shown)


tmcconechy:
@deep7102 i fixed the duplicates in changelog

PR #3168, opened 2019-11-21
2676 - Fix hide pager on one page with datagrid

### LINKED ISSUE #2676: Datagrid issue - hidePagerOnOnePage & filter change not working correctly
**Describe the bug**
We're seeing this in our angular app but I'm confident that it's an enterprise issue. Applies to a datagrid that has paging and filtering enabled and is using hidePagerOnOnePage = true. If you enter a value into the filter that results in less than one page of results, the pager control disappears which is what you would expect. The issue is, when you subsequently clear out the filter value, causing more than one page of rows to be the result, the pager does not re-appear. 

**To Reproduce**
1. Setup a paging and filtering enabled datagrid and option hidePagerOnOnePage=true.
2. Display the grid, initially with more than one page of rows.
3. Note the paging control appears at the bottom of the grid, 'Page 1 of XX'.
4. Enter a column filter value that results in < one page or rows.
5. Note that the paging control disappears - this is expected.
6. Clear out the filter value entered in step 4.
7. Note the paging control does not re-appear.

**Expected behavior**
The paging control should appear/disappear based on the current rows displayed count.

**Version**
IDS-NG 5.5.1
IDS-Enterprise 4.16.2

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 76.0.3809.100

deep7102:
Fixed hide pager on one page and filter was not working correctly with Datagrid.

Closes #2676

- Navigate to http://localhost:4000/components/datagrid/test-hide-pager-if-one-page-filter.html
- See pager nav below the grid, should be showing
- Use column `Activity` to filter `Assemble Paint`
- Should filter accordingly (two results only)
- See pager nav below the grid, should not be showing now
- Clear the filter, should show all the rows
- See pager nav below the grid, should be showing again

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### v4.24.0 Fixes
 
+- `[Datagrid]` Fixed an issue where the hide pager on one page and filter was not working correctly. ([#2676](https://github.com/infor-design/enterprise/issues/2676))
 - `[Datagrid]` Fixed an issue where the dirty cell indicator was not updating after remove row. ([#2960](https://github.com/infor-design/enterprise/issues/2960))
 - `[Datagrid]` Fixed an issue where the method getModifiedRows was not working, it had duplicate entries for the same row. ([#2908](https://github.com/infor-design/enterprise/issues/2908))
 - `[Datagrid]` Fixed an issue where the personalized columns were not working when toggle columns and drag drop. ([#3004](https://github.com/infor-design/enterprise/issues/3004))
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### v4.24.0 Fixes
 
+- `[Datagrid]` Fixed an issue where the hide pager on one page and filter was not working correctly. ([#2676](https://github.com/infor-design/enterprise/issues/2676))
```
```suggestion
- `[Datagrid]` Fixed an issue where the hide pager on one page setting was not working correctly when applying a filter. ([#2676](https://github.com/infor-design/enterprise/issues/2676))
```

PR #1344, opened 2018-12-04
1139 - Fixed lifecycle issues with CAPs that are defined via settings

EdwardCoyle:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
This PR fixes a lifecycle issue on Contextual Action Panel components that are defined via settings with a `trigger: immediate` behavior.  In this use case, the CAP as well as its underlying Modal component should be completely destroyed once the CAP closes.  Previously, this was only happening partially, and remnants of the modal were left behind.  This PR fixes that problem.

**Related github/jira issue (required)**:
Fixes #1139

**Steps necessary to review your pull request (required)**:
- Pull this branch, build, and run the demoapp
- Open a browser to http://localhost:4000/components/contextualactionpanel/example-trigger with a dev tools console
- Open the CAP using the in-page button
- Use the dev tools console and type `$('body').data();`. There should be at least two components invoked against the body tag -- one CAP and one Modal at the very least.
- Use the CAP's close button to dismiss it
- Use the dev tools console and type `$('body').data();` again.  There should be no more instances of either a Modal or CAP established against the body tag.
- It should also be possible to re-open the CAP on this example page after dismissing it.

Also run `npm run functional:local` and ensure the new CAP tests pass.

PR #3228, opened 2019-12-10
1344 - Standardize Spinbox field sizes, add responsive size

EdwardCoyle:
This PR fixes a few issues with Spinbox fields:
- Spinbox fields now have standard sizes (xs, sm, mm, md, lg) that can be applied.  Adding the standard `input-**` CSS classes to spinbox fields will cause them to carry to their wrapping elements when the Spinbox component invokes.
- Responsive form fields now properly support Spinboxes.

Close #1344

- Review the new size comparison page for Spinbox and ensure the sizes match up.  The only one on this page that will not match is the default size, which is shorter than the standard 300px field size on purpose:  http://localhost:4000/components/spinbox/example-sizes.html
- Review "Field Nine" on http://localhost:4000/components/form/example-inputs.html, which is now a responsive Spinbox field.  It should be completely fluid and fill the column's width.


tmcconechy:
Looks good, im ready to approve when you finish the updates to the baselines 🍌

PR #3273, opened 2019-12-18
3265 - Fix the List Style Datagrid Styles to work in all themes

### LINKED ISSUE #3265: Datagrid: Header display issue with column groups
Describe the bug
The column group header does not display correctly when the list is made read-only via the isList grid option.  Using various combinations of themes and variants.

To Reproduce

1. Using the attached example-grouped-headers.html
2. Click on More (....) button
3. Select Theme of 'Vibrant'
4. Select Variant of 'Light'
5. Notice that the grouped column headers do not display correctly

Expected behavior
Grouped column headers display correctly.

Version
v4.25.0-dev

Screenshots

tmcconechy:

Issue raised was about the second grouped header not showing up in some themes. While fixing found some other issues and fixed them like
- should not have hover on rows
- some header styles/states didnt look good.

Fixes #3265 

- http://localhost:4000/components/datagrid/example-list-grouped-headers.html
- test in all 2 themes and 3 variants - should like OK
- http://localhost:4000/components/datagrid/example-list
- also test in all 2 themes and 3 variants - should like OK


tmcconechy:
@EdwardCoyle @deep7102 added a new "fix" to show more groups with a seperator (maybe baselines need updating waiting to see)

Issue #3251, opened 2019-12-16
Datagrid: On Fixed Header Example, column's edge moves faster than the pointer when resizing


brianjuan:
**Describe the bug**
Datagrid: On Fixed Header Example, column's edge moves faster than the mouse pointer when resizing

**To Reproduce**
1. Go to http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html
2. Maximize the browser
3. Drag the right edge of Product Name column to the left
4. See issue: Notice that the line moves faster to the left than the mouse pointer
5. Now try to drag to the right, see issue: the line also moves faster to the right

**Additionally**
These pages exhibit some strange resize behavior:

- http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/test-column-resize-issue.html
- http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/test-columns-mixed.html
- http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/test-columns-percent.html
- test tree http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/example-tree

**Expected behavior**
Column's edge/line should be at the same position with the mouse pointer when resizing.

- [x] resize works on percent
- [x] resize works on tree
- [x] resize works on width and min width
- [x] resize cursor stays "close" to the mouse position
- [x] resize affects only adjacent column like now

**Version**
- ids-enterprise: v4.24.0-beta0

**Screenshots**
![datagrid-linemovesfaster](https://user-images.githubusercontent.com/39508864/70882251-fd055100-2009-11ea-9c4f-f6d391c41c4a.gif)

**Platform**
 - All Mac and Windows browsers

**Additional context**
N/A.

brianjuan:
Also happens in:
http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/test-columns-class.html and
http://4240-beta0-enterprise.demo.design.infor.com/components/datagrid/test-columns-colors.html

tmcconechy:
@deep7102 wondering if you have any ideas. I looked at this as well as it was mentioned on the PR but could not figure out why. This would be nice to fix but is not required as its a problem only on some layouts.

tmcconechy:
An observattion was noted on #3308 
```
[...]

PR #3310, opened 2020-01-09
3308 - Fix stretchcolumns min-width

### LINKED ISSUE #3301: Datagrid: Missing stretchcolumn when re-rendering
**Describe the bug**
We found some columns are not showing/displaying in the View/UI in the latest IDS version.  In our app.

**Investigation**

Columns should be between these 2 column, but it is not appearing. 

image

- Verify Hidden set to ‘False’

image

- Verify column name lapv1490_schedules has values, the plannum has value on it. 
image

**Steps to reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/test-horizontal-scroll.html
2. Type the following in the console
```
var datagrid = $('#datagrid').data('datagrid')
datagrid.settings.stretchColumn = "productName"
datagrid.render()
```
3. Notice productName is invisible.

**Expected behavior**
Available columns should be displayed in the UI.

**Platform**
- Chrome Version 79.0.3945.88 (Official Build) (64-bit)
-data-sohoxi-version="4.25.0-dev"

### LINKED ISSUE #3308: Datagrid: Stretch column can have a width smaller than expected due to width set at 99%
**Describe the bug**
The stretch column can become unusable on smaller devices, or if there are a large number of columns.

**To Reproduce**
1. Go to 'http://master-enterprise.demo.design.infor.com/components/datagrid/test-columns-stretch-column.html'
2. Reduce the width of the browser, until you can no longer see the complete column

**Expected behavior**
The stretch column will only expand to fill the remaining space in the datagrid, but then also maintain its minimal width, so that the column is usable. The minimal width should be applied using the calculated width for the column.

**Version**
Master

**Screenshots**
image

**Platform**
 - OS Version: Windows 10
 - Browser Name chrome

marknolaninfor:
Make sure that the stretch column does not become too small, so that it is not usable

Closes #3308 
Closes #3301  

1. Go to 'http://localhost:4000/components/datagrid/test-columns-stretch-column.html'
2. Make the browser width smaller 

1. http://localhost:4000/components/datagrid/test-editable-lookup.html
2. enter a short value in to the first column (width should be maintained)
3. enter a long value in to the first column, longer that current width (width should be expand)


marknolaninfor:
The failure is due to a slight change in how the browser is applying the width's, eg the fill column.

src/components/datagrid/_datagrid.scss
```diff
@@ -688,8 +688,6 @@ $datagrid-short-row-height: 25px;
   border-collapse: separate;
   border-spacing: 0;
   font-size: $theme-size-font-base;
-  //This lets us resize to lower than content width but it causes layout issues with default col sizes
-  table-layout: fixed;
```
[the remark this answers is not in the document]
marknolaninfor: setting the table-layout to fixed, does not allow the min-width style to be used, as it just ignores that value. 

I have changed it so it works on the inital render and resize only.

docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - `[Datagrid]` Fixed an issue where time picker filter trigger icon and text was overlapping. ([#3062](https://github.com/infor-design/enterprise/issues/3062))
 - `[Datagrid]` Fixed a bug where floating point math would cause the grouping sum aggregator to round incorrectly. ([#3233](https://github.com/infor-design/enterprise/issues/3233))
 - `[Datagrid]` Fixed style issues in all theme and theme variants when using the list style including grouped headers and states. ([#3265](https://github.com/infor-design/enterprise/issues/3265))
+- `[Datagrid]` Fixed strech columns minimum width. ([#3308](https://github.com/infor-design/enterprise/issues/3308))
```
```suggestion
- `[Datagrid]` Fixed issues with the stretch columns minimum width. ([#3308](https://github.com/infor-design/enterprise/issues/3308))
```

src/components/datagrid/datagrid.js
```diff
@@ -5870,6 +5907,17 @@ Datagrid.prototype = {
       $(window).on('orientationchange.datagrid', () => {
         this.rerender();
       });
+      $(window).on('resize.datagrid', () => {
```
tmcconechy: Do we only need this event if ` this.settings.stretchColumn` is true? I was trying to avoid resize events in the datagrid but maybe its not possible.

tmcconechy:
Another thought is do we need stretch column anymore at all? I thought of this when doing the latest changes so the grid will now always stretch using the default table layout system since its now only using one table with body and header. Thinking there might not really be the need to have a stretch column at all since it will auto stretch evenly across all columns.

What do you think about deprecating stretchColumn @marknolaninfor @deep7102 ? Or is it needed for some other reason like perhaps we could use min and width on a column to get the same effect?

deep7102:
Exactly as @tmcconechy explained since we use one table layout, no longer need the stretch column anymore at all.

marknolaninfor:
It needs a strech column, as by default the last column would have been padded. many of our grids are code, description, active. The active grid is the a checkbox so when it fills any remaining space the grid does not look nice. Padding on the description looks way better.

tmcconechy:
@marknolaninfor - just to make sure did you try this against the latest? What it does now is evenly dsitribute it. Maybe still not perfect and we probably dont want to break anything so fine leaving this too. Can we just address the above comments and this should be ok.

- wrap the event in a setting so we dont establish un needed handlers if the setting is not set
- also add the teardown to destroy teardown? if needed?
- resize issue.

I can fix the visual regression tests after for your done

src/components/datagrid/datagrid.js
marknolaninfor: The line above does have the condition (this.stretchColumn !== 'last'), as orientationchange.datagrid also only applies if the stretchColumn is set

tmcconechy: Ok i missed that thanks

tmcconechy:
I think the only remaining questions are:

1) Deepaks error on resize? 
2) In general resizing the stretch column
3) Are the tests fails related or just need updates? Especially the one with an error mentioned in it

marknolaninfor:
The error would be outside the context of this issue and a seperate ticket should be created.

The error is in the master brach, http://master-enterprise.demo.design.infor.com/components/datagrid/test-columns-stretch-column.html, as 'dragend.datagrid', and 'drag.datagrid' events are trying to set a style on the next column, but there is not one on the last column

        const nextCol = DOM.getNextSibling(self.currentHeader, ':not(.is-hidden)'); 
        const nextColGroup = DOM.getNextSibling(currentCol, ':not(.is-hidden)'); (undefined)
        const nextColWidth = parseInt(nextColGroup.style.width, 10); (Therfor errors as it does not exist)

There are also odd behaviour on http://master-enterprise.demo.design.infor.com/components/datagrid/test-columns-percent.html, as it uses the percentage as the pixel width and then restyles based on the change in drag.

marknolaninfor:
The last comit corrects the issue with resizing the stretch column, but if the resizing column and the next column are both percentage then the resize will initally jump, just like in the current master.

tmcconechy:
OK, i will fix the tests and then add those notes to https://github.com/infor-design/enterprise/issues/3251 to fix the resize behavior. Thanks for those observations that does help. I'll note those over there.

marknolaninfor:
Added additional fix for when strecthColumnOnChange is set to true.
http://localhost:4000/components/datagrid/test-editable-lookup.html

tmcconechy:
Added fixes for the baseline changes. Regarding your fix for resize. It works great on the first resize but if you.

- resize stretch column
- sort 
- resize again
- then the behavior reverts as before

But will punt that to https://github.com/infor-design/enterprise/issues/3251 tomorrow and roll this in unless you want to keep looking.

PR #3322, opened 2020-01-13
3315 - Fix errors thrown when using Masks on HTMLInputElement number fields

### LINKED ISSUE #3315: Soho Mask: No digits allowed in the field if the first digit is a zero
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Version**

- ids-enterprise: v4.25.0-dev

**Screenshots**
SohoMaskIssue

image

**Platform**
 - All

EdwardCoyle:
This PR fixes a bug in the system that renders correctly-masked values into Mask Input fields with a `type="number"` attribute.  The component was previously attempting to use the `setSelectionRange()` API on these elements, but that API doesn't exist on number fields, and was causing Javascript errors to throw and mask values to be incorrect.

This PR removes the use of this API when Masks are configured against number inputs.

Closes #3315

- Open http://localhost:4000/components/mask/test-input-type-numeric.html
- Try to key in numbers beyond the first three integers, and test using decimals.  All numbers fitting the mask should render properly.
- There should be no console errors.

### DIFF
--- a/app/views/components/mask/test-input-type-numeric.html
+++ b/app/views/components/mask/test-input-type-numeric.html
@@ -9,7 +9,7 @@
   <div class="six columns">
     <div class="field">
       <label for="test-input">Numeric Input</label>
-      <input id="test-input" type="number" class="new-mask" data-options='{ "process": "number", "patternOptions": { "allowThousandsSeparator": true, "allowDecimal": true, "allowNegative": true, "integerLimit": 7, "decimalLimit": 2 } }' placeholder="-#,###,###.00"/>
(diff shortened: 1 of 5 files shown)


EdwardCoyle:
@fitzorama I had a chat with @Sovia about using formatted numbers with `input[type="number"]` fields, and we wanted to check with you about the impact that might be had on Landmark.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Using_number_inputs has this blurb:
> Bear in mind that, logically, you should not be able to enter characters inside a number input other than numbers. There seems to be some disagreement about this among browsers;

The specification seems to be that formatted numbers aren't going to work properly in these fields, since they are apparently built for floating point math only.

If you're using Number masks on these fields, would it impact you greatly to be using text-based inputs instead, if you need number formatting?

PR #3342, opened 2020-01-17
3218 - Add support for tab in and out with popdown [v4.25.x]

### LINKED ISSUE #3218: Popdown: Tab order broken when tabbing into the popdown and back out of it
**Describe the bug**
We have an implementation where we have a form that we want to display inside a popdown. We want to be able to tab into the popdown, tab thru the fields, and on the blur for the last field, close the popdown and focus into the next field from the base form.

I think this may have something to do with the control reparenting the popdown contents div to the body?

**To Reproduce**
1. Copy [test-tab-example.zip](https://github.com/infor-design/enterprise/files/3933562/test-tab-example.zip) into /components/popdown
2. Navigate to /components/popdown/test-tab-example.html
3. Click or tab into the first Date field
4. Press tab (the popdown opens, and the first field is focused)
5. Press tab (the second field in the popdown is focused)
6. Press tab (the popdown is dismissed, but the next field on the base form (Another Field) isn't focused)

**Expected behavior**
Another Field should get focused when the popdown is dismissed.

**Version**
- ids-enterprise: 4.24.0-dev

**Screenshots**
PopdownTabbingIssue

**Platform**
 - All

deep7102:
Added support for tabbing into and exit out of the Popdown.

Closes #3218

http://localhost:4000/components/popdown/test-first-last-tab.html
http://localhost:4000/components/popdown/test-first-last-tab-custom.html
http://localhost:4000/components/popdown/test-first-last-tab-custom-elements.html
- Open above links
- Click or tab into the first Date field
- Press tab (the popdown opens, and the first field should be focused)
- Press tab (the second field in the popdown should be focused)
- Press tab (the popdown should dismiss, and the next field on the base form (Another Field) should be focused)
- And you can go backward by using the `shift + tab`

### DIFF
--- a/app/views/components/popdown/test-first-last-tab-custom-elements.html
+++ b/app/views/components/popdown/test-first-last-tab-custom-elements.html
@@ -1,0 +1,125 @@
+<style>
+  #popdown-example-trigger span:first-child:after {
+    content: '•';
+    color: #5C5C5C;
+    font-size: 18px;
+    padding-left: 5px;
+    margin-right: 5px;
+    position: relative;
+    top: 2px;
+  }
+
+  #popdown-example-trigger {
+    border: 1px black solid;
+    width: 90px;
+    height: 34px;
+    line-height: 27px;
+  }
+
+  #popdown-example-trigger:focus {
+    border: solid 1px #368ac0;
+    box-shadow: 0 0 4px 3px rgba(54, 138, 192, 0.3);
+    outline: none;
+  }
+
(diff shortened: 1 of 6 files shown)


EdwardCoyle:
@deep7102 I like the feature, but I had a similar feeling.  if you're going backwards in tabIndex order with `Shift + Tab`, when the popdown opens, it would be nice to gain focus on the last field in the popdown instead of the first one.

PR #3344, opened 2020-01-17
3315 - Prevent masks on number input fields from allowing double decimal points [v4.25.x]

### LINKED ISSUE #3315: Soho Mask: No digits allowed in the field if the first digit is a zero

EdwardCoyle:
This PR is a Q/A follow up issue for PR #3322, which fixes a couple things:
- The original issue was that the number input field was not allowing the user to enter leading zeroes. This wasn't a bug with the component, but was an issue with the sample not having the `allowLeadingZeroes` setting turned on.  The sample has been updated to allow leading zeroes.
- It was possible to enter double decimal points on `input[type="number"]` fields, due to issues with detection of non-numeric characters specifically on these fields.  The ability to do this has been removed.

Closes #3315 

- Pull this branch, build, and run the demoapp.
- Open http://localhost:4000/components/mask/test-input-type-numeric
- Try to type more than one zero in the field.  It should work correctly.
- Tap out a number and attempt to add decimal points, followed by decimal numbers.  After adding the decimal numbers, try adding a second decimal point.  This should not be allowed. (should fix the problem noted [here](https://github.com/infor-design/enterprise/issues/3315#issuecomment-575065791)

NOTE: I need to raise a followup issue for permanently fixing a problem in Firefox that occurs, where typing too many `0`s at end of a decimal number resets all the decimal places.  For the purposes of generally fixing this case, I haven't done that in this PR.

### DIFF
--- a/app/views/components/mask/test-input-type-numeric.html
+++ b/app/views/components/mask/test-input-type-numeric.html
@@ -9,7 +9,7 @@
   <div class="six columns">
     <div class="field">
       <label for="test-input">Numeric Input</label>
-      <input id="test-input" type="number" class="new-mask" data-options='{ "process": "number", "patternOptions": { "allowThousandsSeparator": false, "allowDecimal": true, "allowNegative": true, "integerLimit": 7, "decimalLimit": 2 } }' placeholder="-#######.00"/>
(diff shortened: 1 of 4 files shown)


EdwardCoyle:
[The new test change](https://github.com/infor-design/enterprise/pull/3344/commits/fc100d78f1b6c319e0dc1046bc03ff06d4ae3338) fixes an issue in a test caused by the changes from #3334.  This issue is present on `4.25.x` and should be merged into any PR that is built from that branch.

Issue #3143, opened 2019-11-19
Pie/Donut: Line labels cut off when something other than value is used


krahs1:
**Describe the bug**
Pie/Donut labels cut off when something other than value is used. The svg node clips the chart as it has a wide of 75%.

**To Reproduce**

- Change the default 'value' that displays on labels to 'label'
- Change the size of the chart until the labels are clipped by the svg node
Expected behavior

Perhaps use an ellipsis on labels at some point. The biggest issue is the labels on the left side of the chart are clipped from the start of the label (ex: instead of 'Apple' you might see 'ple'

**Version**
-- 4.24.0-dev

Screenshots

Pass the options to the pie chart to display with labels as a text value:
![Screenshot 2019-11-19 07 12 18](https://user-images.githubusercontent.com/19823048/69153039-29798c00-0aa3-11ea-90d6-127b2eb1054a.png)

When the screen is wide enough:
![Screenshot 2019-11-19 07 05 10](https://user-images.githubusercontent.com/19823048/69152862-d0115d00-0aa2-11ea-992d-8bfd5b269fce.png)

Here you can see the clipping on left and right side. So part of the text is not displayed:
![Screenshot 2019-11-19 07 05 22](https://user-images.githubusercontent.com/19823048/69152845-c982e580-0aa2-11ea-9409-90778a32f843.png)

This shows the svg node highlighted:
![Screenshot 2019-11-19 07 05 40](https://user-images.githubusercontent.com/19823048/69152832-c4be3180-0aa2-11ea-8c17-b882080c72a1.png)

This shows the graphics node where the chart is painted:
![Screenshot 2019-11-19 07 05 49](https://user-images.githubusercontent.com/19823048/69152823-c0921400-0aa2-11ea-8e76-6cb690207f78.png)

**Platform**
- Chrome/Any

tmcconechy:
Some thoughts on a fix for this.

- Adjust the svg size from 75% to 100% when hideLegend option is used to get more space.
- But dont make the pie too big
- Decrease the line length
- Move the text to top of the line
- Change algorithm to find longest label to a max and adjust the line size accordingly (lines may have different sizes)

krahs1:
This is required for LMCLIENT-28355
https://jira.lawson.com/browse/LMCLIENT-28355

CindyMercadoReyes:
This issue is now resolved. Moving this ticket to the Done column

PR #3362, opened 2020-01-22
3143 - pie line labels cutoff on mobile

ericangeles:
This PR fixes long text/label on mobile view when using name data. `Short name` will be use when the widget container width is under `520`. There's limitation of `420px` minimum width when resizing the browser. This should be working only when the legend isn't visible and lines show is `label` without and formatter.

Closes https://github.com/infor-design/enterprise/issues/3143

- Navigate to http://localhost:4000/components/pie/test-label-long-text.html.
- Resize the browser with the minimum width of 420px or try to log the parent width (widget) to check the size of it. 520 is the maximum width.
- Labels should be change to the short name instead of the name.
- Resize to bigger width, and it should show the name (longer text).


tmcconechy:
This pr needs master merged (after i syn 4.25) I will take care if it shortly.

Issue #3176, opened 2019-11-25
Colour of icons in read-only controls is not consistent


sanx72:
In the File Upload example page, for the read-only field, the icon is a light-gray to suggest it cannot be selected (which I think is the correct look)...

https://design.infor.com/code/ids-enterprise/latest/demo/components/fileupload/example-index

...but in read-only versions of other edit field controls, the icon is dark gray/ black, making it look like it can be selected; e.g.

https://design.infor.com/code/ids-enterprise/latest/demo/components/dropdown/example-readonly
https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/example-index

I think others, like lookup and search, have the same issue but don't have a standard example to link to.

The color picker is a bit different in that the icon is ok until the mouse is moved over the field then it appears to become "active" as both the icon and border change...

https://design.infor.com/code/ids-enterprise/latest/demo/components/colorpicker/test-states

I would expect all teh edit fields that have icons to look and behave the same when they are set to read-only.

Tried in latest Chrome and Firefox versions on Windows 8.1 PC.

janahintal:
This issue is now resolved. Thank you. 
Verified in 
http://localhost:4000/components/colorpicker/test-states
http://localhost:4000/components/datepicker/example-index
http://localhost:4000/components/dropdown/example-readonly
http://localhost:4000/components/colorpicker/test-states
http://localhost:4000/components/datepicker/example-index
http://localhost:4000/components/dropdown/example-readonly

PR #3403, opened 2020-01-29
3176 - Fix color inconsistencies of the icons

ericangeles:
This PR fixes color inconsistencies of the icons when the fields are in readonly state for dropdown, datepicker, and colorpicker. 

Additional fixes
- Hovering on icon color shouldn't look active. It should still be the same.
- Hovering on the field shouldn't change the border color.

Following the icon colors (for all themes) of the file upload example. 
See: https://design.infor.com/code/ids-enterprise/latest/demo/components/fileupload/example-index

Closes https://github.com/infor-design/enterprise/issues/3176

Open the three test pages from colorpicker, datepicker, and dropdown components.
- http://localhost:4000/components/colorpicker/test-states
- http://localhost:4000/components/datepicker/example-index
- http://localhost:4000/components/dropdown/example-readonly
---- 
- For colorpicker (http://localhost:4000/components/colorpicker/test-states) test page, hover the input field (**on the Readonly field**).
- It should not appear any border color and the icon shouldn't change the color.
- Validate and the color if it's the same as the [file upload test page](https://design.infor.com/code/ids-enterprise/latest/demo/components/fileupload/example-index) via inspecting the element.
- Test it to all on different themes and variants.
---
- For datepicker (http://localhost:4000/components/datepicker/example-index) test page, check the second input field, that is in readonly state.
- Validate and compare the color if it's the same as the [file upload test page](https://design.infor.com/code/ids-enterprise/latest/demo/components/fileupload/example-index) via inspecting the element.
---
- For dropdown (http://localhost:4000/components/dropdown/example-readonly) test page, check the second and third input fields that has readonly label.
- Validate and compare the color if it's the same as the [file upload test page](https://design.infor.com/code/ids-enterprise/latest/demo/components/fileupload/example-index) via inspecting the element.
- Also, click the "Remove Readonly Attribute" button repeatedly.
- Icon should be back to it's color when it's in readonly.

### DIFF
--- a/src/core/_config.scss
+++ b/src/core/_config.scss
@@ -234,6 +234,7 @@ $dropdown-menu-border-color: $theme-color-brand-primary-base;
 $dropdown-menu-separator-color: $theme-color-palette-graphite-40;
 $dropdown-menu-separator-border-color: $theme-color-palette-graphite-30;
 $dropdown-box-shadow-top: 0 0 5px $drop-shadow-depth;
+$dropdown-readonly-icon-color: $theme-color-palette-graphite-30;
 
 //List View
 $listview-bg-color: $theme-color-palette-white;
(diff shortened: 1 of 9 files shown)


src/core/_config.scss
```diff
@@ -332,7 +333,7 @@ $datepicker-disabled-icon-color: $input-color-disabled-font;
 $datepicker-disabled-border-color: $input-color-disabled-border;
 $datepicker-disabled-bg-color: $theme-color-palette-graphite-10;
 $datepicker-disabled-color: $input-color-disabled-font;
-$datepicker-readonly-icon-color: $input-color-readonly-font;
```
EdwardCoyle: @tmcconechy if this one wasn't correct, we may need to update this color in the tokens

PR #3405, opened 2020-01-29
3315 - Fix standard inputs not allowing leading zeroes in Number Masks [v4.25.x]

### LINKED ISSUE #3315: Soho Mask: No digits allowed in the field if the first digit is a zero

EdwardCoyle:
This PR fixes a regression in the way Number Masks handle leading zeroes on standard input fields (read: not number input fields).  Previously it was not possible to add more than one zero to the integer part of a Number Mask on an input field that was configured to allow leading zeroes.  It was also not possible to type a zero, then any other number in these same fields.  These issues are now resolved.

Closes #3315 

- Open http://localhost:4000/components/mask/test-number-leading-zeroes
- In the "Text Inputs" section, choose either of the inputs that allow leading zeroes, and attempt to type a number with several leading zeroes.  Multiple zeroes should be allowed.  Additionally, typing a number of zeroes greater than the group size should not cause those zeroes to be formatted with a thousands separator (for example, typing "0000" should not result in "0,000").  However, typing a regular number should format properly in these fields.
- It should also be possible to type numbers like "0001234", and that would format to "0001,234".
- Check that multiple zeroes in the decimal area are also allowed.
- In the fields that allow it, check to make sure negative numbers are working.

There is a new functional test that extensively checks this feature, which should pass.

### DIFF
--- a/app/views/components/mask/test-input-type-numeric.html
+++ b/app/views/components/mask/test-input-type-numeric.html
@@ -9,7 +9,7 @@
   <div class="six columns">
     <div class="field">
       <label for="test-input">Numeric Input</label>
-      <input id="test-input" type="number" class="new-mask" data-options='{ "process": "number", "patternOptions": { "allowLeadingZeroes": true, "allowThousandsSeparator": false, "allowDecimal": true, "allowNegative": true, "integerLimit": 7, "decimalLimit": 2 } }' placeholder="-#######.00" step="0.01"/>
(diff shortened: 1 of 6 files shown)


EdwardCoyle:
Just gonna hold off on a merge of this until @Sovia has a chance to double check against Landmark

tmcconechy:
@Sovia Could you check this today as soon as you can. Would like to get this patch version out today.
Thanks!!!

Issue #3312, opened 2020-01-10
Module Tabs: No selected/opened tab indicator after changing color


sheilamaejornadal:
**Describe the bug**
Pre-condition:
Module tabs should be open.

After changing theme to uplift and then a color, there is no obvious indicator on which of the multiple tabs is selected/opened.

**To Reproduce**
1. Open http://master-enterprise.demo.design.infor.com/patterns/module-tabs.html?colors=db7726&theme=soho&variant=light#order-123456
2. Change the theme to uplift
3. select a color
4. notice all tabs are the same color (and notice the line below the Tab) is not correctly styled

Or simply go to http://master-enterprise.demo.design.infor.com/patterns/module-tabs.html?colors=db7726&theme=soho&variant=light#order-123456

**Expected behavior**
After changing the theme, there should be an indicator in the color of the tabs to identify which is selected/open at the moment.

**Version**
- ids-enterprise: 4.20.1

**Screenshots**
See attached gif file for reference
![tabs-theme-issue](https://user-images.githubusercontent.com/52568990/72136303-7868e100-33c3-11ea-98fe-2da64727c323.gif)

**Platform**
 - OS Version: Windows 10
 - Browser Name: chrome
 - Browser Version: 79.0.3945.88

brianjuan:
Selected tabs are now indicated or have different color from the unselected tabs after changing Color. Tested in http://master-enterprise.demo.design.infor.com/patterns/module-tabs.html.

However, I found a related minor issue. It seems like after changing the Color, unselected tabs remember the previously selected color in the Color menu.
1. Go to http://master-enterprise.demo.design.infor.com/patterns/module-tabs.html
2. Change the Color to Amber
3. Click the unselected tab "Order Entry"
4. Open the ... menu > Colors, See issue: Default is still checked instead of Amber
![image](https://user-images.githubusercontent.com/39508864/73740306-e9ad7100-4782-11ea-9474-5bc33a4fa6d6.png)

5. Change the Color again to Emerald
6. Click the unselected tab "New Tab Builder"
7. Open the ... menu > Colors, See issue: Amber is still checked instead of Emerald
![image](https://user-images.githubusercontent.com/39508864/73740365-09449980-4783-11ea-9400-0cb5500b35ef.png)

Same issue also happens when changing Theme or Variant.

tmcconechy:
@ericangeles - see if its possible to correct that case by @brianjuan 
[...]

PR #3409, opened 2020-01-30
3312 - Fix tab indicator color in uplift theme for all color variations

ericangeles:
This PR fixes color tab indicator and a small space gap below when selected/opened for all color variations in uplift theme.

Closes https://github.com/infor-design/enterprise/issues/3312

- Navigate to http://localhost:4000/patterns/module-tabs.html
- Change the theme to `uplift/vibrant`.
- Change also the color to `Amber`.
- The selected/opened tab should be more lighter or much emphasized than the rest of the tabs.
- Test it also by selecting different tabs.
- Should work the same way.
- It should darken the background color when hovering, and it should not show any gap below of the tab.
- Test this also for the rest of color variations.


tmcconechy:
This looks good. The only (other) could of things we could do to make this a bit better.
- the hover color is VERY dark, can we lower that a bit? should probably be less than the active state.
- can you put one of the color switches on the toolbar on all the tabs, if you switch tab it doesnt have a switcher on it and that threw me off a bit, could do the same to QA

tmcconechy:
Also btw fixed some tests on master for dates. You may need to merge master on if your tests fail still.

ericangeles:
> Also btw fixed some tests on master for dates. You may need to merge master on if your tests fail still.

Yeah, notice that too. Give me a sec.

PR #3415, opened 2020-01-30
3243 - Fix misplaced autocomplete list on some Searchfields

### LINKED ISSUE #3243: Searchfield: Sometimes the result are detached from the searchfield
**Describe the bug**
Bug or By design?

We've found a minor issue with the searchfield. Sometimes, when the search result is displayed above the searchfield the results get detached from the searchfield. 

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/searchfield/example-different-template.html
2. Either make the browser smaller or add some more text using the developer tools to make the searchfield appear at the bottom of the screen
3. Write 'A' in the searchfield. 
4. Notice Alabama is at the top position
5. Write 'labama' after 'A' in the searchfield so that it now says 'Alabama'
6. See error: Alabama is still shown at the same position so the search results are now detached from the searchfield

**Expected behavior**
The searchfield should not be detached and that Alabama (And the "Show all results for..") should be shown right above the searchfield with no space in between

**Version**

- ids-enterprise: v 4.23.0

**Screenshots**
image

image

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 78.0.3904.108 (Official Build) (64-bit)

**Additional context**
-

EdwardCoyle:
This PR fixes a bug where changing filter results while the autocomplete is open may result in the menu being positioned incorrectly.  This was due to an incorrect rendering pipeline.

Closes #3243

- Open http://localhost:4000/components/searchfield/test-place-on-bottom.html
- Shrink/Grow your browser viewport to make the Searchfield nearly aligned to the bottom of the viewport. (for testing puposes I used about 640x480px)
- Type "n" in to the Searchfield.
- Ensure the filtered results list appears above the Searchfield and is connected to its top.
- Finish typing the word "new"
- When the list shrinks, it should re-position itself and remain connected to the top of the searchfield.

This case is also covered by a new visual regression test, so make sure all tests pass.


CindyMercadoReyes:
Moving this ticket to QA Failed. The issue still appears on iPhone pro 11. Please see screen shot attached. 
bs_realios_Mobile_iPhone 11 Pro-13 0

Issue #1643, opened 2019-02-18
Datagrid Multi-line field not expanding while entering text


NathanLouisBarcial:
**Describe the bug**
For multiline fields in Datagrids, as user is entering lines of text, the cell is not expanding thus, some lines of text are being cropped. Although, the cell will expand once user focuses out of the cell.

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/test-long-text.html'
2. Enter some lines of text in the multiline "Activity" column
3. As user is typing several lines, the cell is not expanding cropping out the first lines of text. The cell will expand after user focuses out of the cell.

As User is still typing:
![image](https://user-images.githubusercontent.com/41028784/52980802-ea84b000-3416-11e9-8726-5ecfb9dc98e3.png)

After focusing out:
![image](https://user-images.githubusercontent.com/41028784/52980847-1d2ea880-3417-11e9-84b3-cf78ea691d36.png)

**Version**
Version: 4.16.0-dev

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome

tmcconechy:
What editor did you use? Probably this wont be easily fixable. I think it's acceptable if the cell expands when its done typing. And scrolls while editing initially.

NathanLouisBarcial:
Hi @tmcconechy the editors used is Editors.Textarea

Yes, the scroll while editing might be okay, but one issue we're encountering with this is once we focus out of the cell and try to edit again a cell with multiple lines, the height of the row is still retained and the scroll bar is cropped:

![image](https://user-images.githubusercontent.com/41028784/53056504-64d03580-34e6-11e9-8d1a-de954a68d2d1.png)

tmcconechy:
Gotcha - yeah this seems like something we should address.

NathanLouisBarcial:
Alright, thanks Tim.

tmcconechy:
Think this is now completed

PR #3418, opened 2020-01-31
3417-textarea row height

### LINKED ISSUE #3417: The textarea formatter is not the expected height when the row height is set to 'Short' or 'Medium''
**Describe the bug**
The textarea formatter is not the expected height when the row hieght is not set to 'Normal', eg. Short or Medium

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/test-editable-long-text.html'
2. Set the row height to 'Short'
3. Edit the Product Id in the second row 

**Expected behavior**
Single lined textara formatters should be the correct size that matches the row height. This works as expected when the row height is set to normal. 

**Version**
v4.26.x

**Screenshots**
image

**Platform**
 - OS Version: Windows 10 
 - Browser Name: chrome

**Additional context**
There are other issues with the textarea editor, but they should be done as part of #1643

marknolaninfor:
Correct the row height when the textarea formatter is used.

Closes #3417 3417

Go to 'http://localhost:4000/components/datagrid/test-editable-long-text.html'
Set the row height to 'Short'
Edit the Product Id in the second row


marknolaninfor:
'Datagrid Align Header Text Toggle Tests Should not visual regress' fails as it contains a textarea formatter that is viewed with the Short row height.

tmcconechy:
Do you think that the problem that when you click in the comment (Textarea) cell it collapses can be fixed? Its not really part of https://github.com/infor-design/enterprise/issues/1643 or is it?

Feel like we should just fix up this editor entirely. Let me know if you want to take a stab at it? If not i can help. And i will fix the images when this is ready 100%

marknolaninfor:
The textarea editor, now displays as an ids text area, and allows for the text editor options to be set. The activity column will show the remaining characters, and the status column is just a standard text area.

The autogrow does not work correctly as the heights are not calculated correctly. This ticket is a start on #1643 , but I was not able to implement.

marknolaninfor:
@tmcconechy 

The focus issue caused by the click has been corrected, and the character count will not display by default, (different from the textarea control)

The editor uses the textarea ids control, over just a standard textarea, so you get all the additional functionality of the textarea control, eg autogrow.

I think the behaviour should be the same to be consistant, similar to the other editors like the datepicker.
http://localhost:4000/components/textarea/example-index.html

Issue #3392, opened 2020-01-27
Datagrid Component: Uplift Dark and High Contrast Multiple Issues


CindyMercadoReyes:
**Describe the bug**
Datagrid Component: Uplift Dark and High Contrast Multiple Issues

**To Reproduce**
1. Go to url: http://master-enterprise.demo.design.infor.com/components/datagrid/example-index?theme=uplift&variant=dark
2. Click on uplift theme and then also click on Dark contrast
3. Please see my screen shot for the multiple issues that need to be correct on Dark contrast

**Expected behavior**
The colors and text on the data grid need to updated. Spoke to Tim and Tim will work with the Design team. 

**Version**

- ids-enterprise: [e.g. v4.9.0 or v4.10.0]

**Screenshots**
<img width="1252" alt="Datagrid Dark Contrast Multiple Issues" src="https://user-images.githubusercontent.com/39101346/73207701-2938e500-4113-11ea-99d9-fdef0e41732b.png">

**Platform**
 - All Browsers 

**Additional context**
None

tmcconechy:
@ericangeles 
Forgot I actually fixed this already in 4.25. Just try and test it some more and i think we can just move to done

ericangeles:
@tmcconechy it seems hovering the icons does not make it darker. That's the only issue that am seeing now.

ericangeles:
@tmcconechy am not certain about the issue that I saw. Should we add hover effect the sort up and down icons? It seems that clicking the desc icon isn't functioning the descending order but it always in ascending. If we add the hovering effect, it should follow the icon that you hover and click.

tmcconechy:
Its looking ok for me i think? If i go to:

http://master-enterprise.demo.design.infor.com/components/datagrid/example-index?theme=uplift&variant=dark

Are you just suggesting that we make the hover color on the header dark instead of light? If so that could look better. Feel free to give it a try and we can review

ericangeles:
> Its looking ok for me i think? If i go to:
> 
> http://master-enterprise.demo.design.infor.com/components/datagrid/example-index?theme=uplift&variant=dark
> 
> Are you just suggesting that we make the hover color on the header dark instead of light? If so that could look better. Feel free to give it a try and we can review

So in that case, this will be reflectes to for all the themes and contrast?

tmcconechy:
Ok, maybe im confused about what your saying? Can you describe the issue you see? It seems not bad in all themes with what i did to me.

[...]

PR #3440, opened 2020-02-05
3392 - Add hover color effect on sort icons

ericangeles:
This PR adds hover effect on sort icons.

Closes https://github.com/infor-design/enterprise/issues/3392

- Navigate to http://localhost:4000/components/datagrid/example-index
- Hover on the header, then hover the icon when visible
- It should change the color of the icon when hovering
- Test this on all themes and different variants


src/components/datagrid/_datagrid.scss
```diff
@@ -2424,6 +2424,10 @@ $datagrid-short-row-height: 25px;
 
         .icon {
           opacity: 1;
+
+          &:hover {
```
tmcconechy: I think we can remove this then its perfect. There isnt really a need to have a state on the icon. The state is for the whole column and happens when you hove the column. Since clicking the icon does nothing different than clicking the header

ericangeles:
@tmcconechy got it. Will close this PR then move the ticket to Ready for QA.

PR #3443, opened 2020-02-05
2553 - Add a fix to filter example and fix frozen filter editors

### LINKED ISSUE #2553: Datagrid: Filter is returning wrong result
**Describe the bug**
Datagrid: Filter is returning wrong result.

Affected pages:
- http://master-enterprise.demo.design.infor.com/components/datagrid/example-filter.html
- http://master-enterprise.demo.design.infor.com/components/datagrid/example-custom-filter-conditions.html
- All datagrid pages with filter

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/datagrid/example-custom-filter-conditions.html
2. Select "Contains" filter in ID column, then type 5test
3. See issue: Datagrid will return rows with ID 5
4. Select "Equals" filter in ID column, then type 5test
5. See issue: Datagrid will return rows with ID 5
6. This is also happening in other columns except for Product Name

**Expected behavior**
Filter should return correct result

**Version**
- ids-enterprise: v4.11.0 to v4.20.0-beta1

**Screenshots**
image

**Platform**
 -All devices, OS, and browsers.

**Additional context**
No additional context.

tmcconechy:

I thought that #2553 was fixed so included it in this sprint when testing filter stuff in datagrid. But it failed, in effort to fix it I found a worse problem. When the filter fields are in the frozen section of the datagrid the editor was not being initialized, so the mask didnt work.

Fixes #2553 

- go to http://localhost:4000/components/datagrid/example-filter.html
- notice the "Product Id" column now has an editor
- try and type "24test" in that filter field 
- notice you cant because it has a mask for only numbers that works now 👍 


janahintal:
This issue is now resolved. Thank you.
Verified in http://master-enterprise.demo.design.infor.com/components/datagrid/example-filter.html?theme=uplift&variant=light

Issue #3473, opened 2020-02-11
Datepicker: Chinese cannot be used in datepicker with time


SofiK:
**Describe the bug**
When I select a time in the datepicker (Or click Today) and then unfocus the field the field becomes invalid.

**To Reproduce**
1. Go to 'https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/example-timeformat.html?locale=zh-CN' or 'https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/example-timeformat.html?locale=zh-TW'
2. Click on the calendar icon in the second field
3. Click Today
4. Click outside the field (not in another field, for some reason it doesn't validate then)
5. Notice that it becomes invalid.

**Expected behavior**
The field should keep the format (or have the correct format which I do not know) and it should not be invalid

**Version**

- ids-enterprise: 4.23 but reproducible in 4.25.2

**Screenshots**
![image](https://user-images.githubusercontent.com/8664975/74255706-8d060500-4cf2-11ea-9f3e-7c981409efd1.png)

**Platform**
 - OS Version: Win 10
 - Browser Name: Chrome
 - Browser Version: Version 79.0.3945.130 (Official Build) (64-bit)

**Additional context**
This issue can be related to https://github.com/infor-design/enterprise/issues/3470
All these 3 language codes (ko-KR, zh-Ch and zh-TW) gets the same error.

jbrcna:
**QA PASSED with observation**
The issue is now resolved. v4.27.0-dev
verified in https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html?locale=zh-TW
In `Safari` Browser the font in menu looks different.
![image](https://user-images.githubusercontent.com/56722184/76412653-edaa5f80-63ce-11ea-9aeb-5a58da9ce02f.png)

tmcconechy:
This is correct @jbrcna for your info. Some fonts like chinese need to be in a special font to show all the characters so we make that adjustment.

Issue #3470, opened 2020-02-11
Datepicker: Korean doesn't work in datepicker


SofiK:
**Describe the bug**
When I select a time in the datepicker (Or click Today) and then unfocus the field the field becomes invalid.

**To Reproduce**
1. Go to 'https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/example-timeformat.html?locale=ko-KR'
2. Click on the calendar icon in the second field
3. Click Today
4. Click outside the field (not in another field, for some reason it doesn't validate then)
5. Notice that it becomes invalid.

**Expected behavior**
The field should keep the format (or have the correct format which I do not know) and it should not be invalid

**Version**

- ids-enterprise: 4.23 but reproducible in 4.25.2

**Screenshots**
![image](https://user-images.githubusercontent.com/8664975/74251772-ec611680-4cec-11ea-8f7f-65125a95be28.png)

**Platform**
 - OS Version: Win 10
 - Browser Name: Chrome
 - Browser Version: Version 79.0.3945.130 (Official Build) (64-bit)

**Additional context**
-

janahintal:
QA Passed. Korean is now working and verified in https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html?theme=uplift&variant=contrast&locale=ko-KR across all platform. Thank you.

Issue #3474, opened 2020-02-11
Datepicker: AM/PM doesn't work correctly for locale hi-IN


SofiK:
**Describe the bug**
When I try to switch between AM and PM in Hindi it changes back to a value dependent on the time. 

**To Reproduce**
1. Go to 'https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/example-timeformat.html?locale=hi-IN'
2. Click on the calendar icon for the second datepicker that will use the time 12:00
3. For the am/pm dropdown select the first option
4. Click The apply button
5. Notice that the time is still shown correctly with the correct am/pm option
6. Click outside of the field
7. Notice that the am/pm value changes (Expected result: The value is not changed)
8.  Click on the calendar icon for the third datepicker that will use the current time
9. For the am/pm dropdown select the second option
10. Click the apply button
11. Notice that the time is still shown correctly with the correct am/pm option
12. Click outside of the field
13. Notice that the am/pm value changes (This could perhaps depend on if it is currently am or pm time? I'm in pm for my time zone when testing?) (Expected result: The value is not changed)

**Expected behavior**
The am/pm value should not change in step 7 and 13

**Version**

- ids-enterprise: Found in v 4.25.2

**Screenshots**
![Testing datepicker hi-IN](https://user-images.githubusercontent.com/8664975/74259406-21bf3180-4cf8-11ea-9103-198a47b45472.gif)

**Platform**
 - OS Version: Win 10
 - Browser Name: Chrome
 - Browser Version: Version 79.0.3945.130 (Official Build) (64-bit)

**Additional context**
-

jbrcna:
**QA PASSED**
The issue is now resolved. v4.27.0-dev
verified in https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html?locale=hi-IN

Issue #3202, opened 2019-12-03
Datepicker: Today is changing PM/AM value


SofiK:
**Describe the bug**
Hi! 
We want a way to allow the user to set the time they want and then select todays date by clicking Today if that is wanted. This works fine for some time periods (especially in he morning) but not for 12:00 - 12:59 **AM** when it changes to 12:XX PM And for 1:00-11:59 **PM** which changes to XX:XX **AM**. 

**If the user click the current date, then it works fine. It is just when clicking Today that this issue occur.**

**Note** that we do not want it to choose current time as in this issue https://github.com/infor-design/enterprise/issues/2575. 

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/datepicker/test-12hr-time.html
2. Click on the datepicker
3. Change the time to "11:55 AM"
4. Click Today
5. Note that the time remains 11:55 (which is what we want)
6. Click on the datepicker icon again 
7. Change the time to "12:10 AM"
8. Click Today
=> Time changes to 12:10 **PM**
9. Click on the datepicker
10. Change the time to "6:55 PM"
11. Click Today
=> Time changes to "6:55 **AM**"

**Expected behavior**
I think this could be handled differently depending on the value of useCurrentTime option,  If useCurrentTime is false I do not think current time should be used while clicking Today (If useCurrentTime is true then Today should change to current time). 

When "useCurrentTime" setting is false, the Time should not change at all when clicking Today (just as when you click todays date). So the "12:XX AM" should stay "12:XX **AM**" after clicking Today and PM should stay PM for all afternoon values.

**Version**

- ids-enterprise: v. 4.23.0

**Screenshots**
![image](https://user-images.githubusercontent.com/8664975/70044282-5e690100-15c2-11ea-9014-0c484492f741.png)

![image](https://user-images.githubusercontent.com/8664975/70044298-6759d280-15c2-11ea-9226-0830e25a8d46.png)

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version:Version 78.0.3904.108 (Official Build) (64-bit)

**Additional context**
Currently the sample in the steps to reproduce does not have the "useCurrentTime" option set, but I think it should have been false by design according to spec? I've also tried to add the "useCurrentTime" option and set it to false. But still have issues with AM/PM changes.

SofiK:
[...]

PR #3502, opened 2020-02-18
3470/3473 - Fix korean and chinese locale date format

ericangeles:
This PR fixes korean and chinese locale format. It makes it invalid after selecting date and out focusing in the field. It seems that the current date format for both locale doesn't work when parsing the date.

This is still WIP as I bump into an issue.
- Selecting PM will not work, it will still choose AM.

Closes 
- https://github.com/infor-design/enterprise/issues/3470
- https://github.com/infor-design/enterprise/issues/3473

- Navigate to http://localhost:4000/components/datepicker/example-timeformat.html?locale=ko-KR for Korean
- Then http://localhost:4000/components/datepicker/example-timeformat.html?locale=zh-CN for Chinese
- Click the calendar icon
- Select today or any date
- Click apply or click outside to focus out the field
- It shouldn't invalid the value
- Play all the fields to make sure it's working properly

### DIFF
--- a/src/components/locale/cultures/ko-KR.js
+++ b/src/components/locale/cultures/ko-KR.js
@@ -11,19 +11,19 @@ Soho.Locale.addCulture('ko-KR', {
     name: 'gregorian',
     // ca-gregorian/main/dates/calendars/gregorian/dateFormats/
     dateFormat: {
-      separator: '. ', // Infered
+      separator: '-', // Infered
       timeSeparator: ':',
-      short: 'yyyy-MM-dd', // Uses custom format
-      medium: 'yyyy-MM-dd',
+      short: 'MM-d-yyyy', // Uses custom format
+      medium: 'MM-d-yyyy',
       long: 'yyyy년 M월 d일',
       full: 'yyyy년 M월 d일 EEEE',
       month: 'M월 d일',
       year: 'yyyy년 M월',
       timestamp: 'a h:mm:ss',
-      hour: 'a h:mm',
-      datetime: 'yyyy-MM-dd a h:mm',
-      timezone: 'yyyy-MM-dd a h:mm zz',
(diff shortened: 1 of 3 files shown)


tmcconechy:
Issue looks similar to https://github.com/infor-design/enterprise/issues/3474 and https://github.com/infor-design/enterprise/issues/3202 - dont have a solution for you but noting...

src/components/locale/cultures/ko-KR.js
```diff
@@ -11,19 +11,19 @@ Soho.Locale.addCulture('ko-KR', {
     name: 'gregorian',
     // ca-gregorian/main/dates/calendars/gregorian/dateFormats/
     dateFormat: {
-      separator: '. ', // Infered
+      separator: '-', // Infered
       timeSeparator: ':',
-      short: 'yyyy-MM-dd', // Uses custom format
-      medium: 'yyyy-MM-dd',
+      short: 'MM-d-yyyy', // Uses custom format
```
tmcconechy: I think this format was correct https://en.wikipedia.org/wiki/Date_format_by_country so i dont think we should change it. Not sure about the a in time if thats the problem we can remove that but shouldnt generally change the format here.

ericangeles: It seems that position of time period is the problem here. Reverting back to the original date format.

ericangeles:
Will probably add the related issues #3474 & #3202

tmcconechy:
Getting anywhere with this? Or need any help debugging?

ericangeles:
> @ericangeles Getting anywhere with this? Or need any help debugging?

Probably need a help on this. Kinda tricky.

tmcconechy:
Have an alternate fix coming soon on a new PR.

PR #3549, opened 2020-02-26
2399 - Fix a focus trap in datagrid

### LINKED ISSUE #2399: Datagrid/Modal: Focus Trap Issue
**Describe the bug**
There is a focus trap issue involving Datagrid components inside of Modals. Inside an active Modal, once keyboard focus is on an element inside of the Datagrid, it is no longer possible to navigate outside of the Datagrid's rows/columns. 

This doesn't appear to affect Lookup components.  Specifically, it affects a full-fledged Datagrid used inside of a Modal.

**Steps to reproduce**
- Open http://localhost:4000/components/modal/test-toolbar.html
- Click "Open Modal"
- Press `Tab` key once to move focus to the grid
- Press `Shift + Tab` to try and move focus back to the button.  See that it doesn't move focus out of the grid.
- The same occurs when tabbing all the way through all grid cells.  It's not possible to then press `Tab` to get out of the grid and into the Pager bar, or beyond.

**Version**
Found while testing v4.19.x, may have been present before since the example page was new.

**Additional Context**
Originally found while testing [#2381](https://github.com/infor-design/enterprise/pull/2381#issuecomment-502838269)

tmcconechy:

When using actionableable mode in datagrid (tabs instead of arrow keys to navigate). There was an issue that the tabs would not release out of the datagrid - thus a focus trap. Made the tab key move up and down rows. Once you get to the end or start of the grid (tab or shift tab) then you will tab to the next page focusable item.

Fixes #2399 

- go to http://localhost:4000/components/datagrid/test-editable-actionable-mode.html
- tab down the page - notice tabs at the end of the row will move to next row - and then ultimately down to the pager
- shift+tab down the page - will do the reverse
- go to http://localhost:4000/components/datagrid/example-index
- tab into the grid, then shift + tab out. in this mode tabs will not work to move cells only arrow keys
- Open http://localhost:4000/components/modal/test-toolbar.html
- Click "Open Modal"
- Press Tab key once to move focus to the grid
- Press Shift + Tab to try and move focus back to the button. See that it does focus out of the grid now

### DIFF
--- a/app/views/components/datagrid/test-expandable-row-multiselect.html
+++ b/app/views/components/datagrid/test-expandable-row-multiselect.html
@@ -1,0 +1,63 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+{{={{{ }}}=}}
+
+<script>
+  $('body').one('initialized', function () {
+
+      var grid,
+        columns = [],
+        data = [];
+
+      // Some Sample Data
+      data.push({ id: 1, productId: 2142201, sku: '<b>SKU #9000001-237</b>', productName: 'Compressor', activity:  'Assemble Paint', quantity: 1, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action'});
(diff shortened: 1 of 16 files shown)


deep7102:
Like the tab feature to navigate the grid this way, Looks way nicer to move to next/prev rows.
One thing saw there is no way to navigate pager items with keyboad tab-or-arrows

PR #3582, opened 2020-03-05
ng#467 - Add API to control Buttons inside Modals

EdwardCoyle:
This PR creates a "ButtonSet" API that can be used within Modals to control an entire row of Modal Buttons at once. To add this functionality, this PR:
- Fleshes out many aspects of the existing Button API to be more programmatic and controlled with Javascript API.
- Takes the previous syntax for defining a Button from the Modal component's `buttons` setting and translates it to a new component, `ButtonSet`, for adding/removing buttons by using the element, API reference, or an ID.
- Replaces the existing method of dealing with Modal Buttons with the new ButtonSet component, allowing it to be controlled from the Modal API.

Related to infor-design/enterprise-ng#467

Pull this branch, build, and run the demoapp.  Then:

_To test the ButtonSet and new Button settings:_
- Open http://localhost:4000/components/button/test-buttonset-gauntlet.html
- Use all the various settings to test adding buttons.
- Use `reset` to clear the ButtonSet.
- Use the "Default/Modal" switch to toggle styles.  Make sure the background color disappears on Primary buttons when using 'modal` mode.

_To test Modal integration:_
- Basically [smoke test all Modal samples](http://localhost:4000/components/modal/list) and make sure they match what's currently on `master`.  These are all using the new component, but their general look/feel/actions should be the same.
- Also check that all [Contextual Action Panel samples](http://localhost:4000/components/contextualactionpanel/list) work as expected.  These are not using the new component and continue to use either standard Toolbar/Flex Toolbar for now.  Nothing should have changed here.
- There are also many new functional tests for Button, ButtonSet, and Modal that cover the button definition cases described in the ticket.


EdwardCoyle:
@tmcconechy I fixed pretty much everything there except the placement of the validation messages, and the (X) button tooltip.  We should make a new issue for those -- this issue's pretty big and those wouldn't have been affected by this -- seems like changing them affects quite a few visual regressions too.

PR #3591, opened 2020-03-09
3341 - Fix position was not aligning with busy indicator

### LINKED ISSUE #3341: Busy Indicator: Scroll in parent not accounted for at initialization
**Describe the bug**
When the Busy Indicator is located inside a scrollable container, the scroll is not accounted for at initialization. The Busy Indicator will always be initialized at the center of the parent container, but depending on how much the user has scrolled this might be out of view. It is not until the user performs a scroll that the Busy Indicator will be centered again, since it has an event listener that triggers when scroll is performed.

A fix for this could be to calculate the offset of the busy indicator in regards to the parent container scroll, like it is done in the event listener https://github.com/infor-design/enterprise/blob/5a497e5d9dd2afd485fde33f72426801e68fec76/src/components/busyindicator/busyindicator.js#L361 and use that as the top property in the css here: https://github.com/infor-design/enterprise/blob/5a497e5d9dd2afd485fde33f72426801e68fec76/src/components/busyindicator/busyindicator.js#L356

**To Reproduce**
1. Go to sample /busyindicator/example-non-blocking
2. Edit sample to force scroll on parent container:
2.1. Wrap busy indicator div inside a new div. Set styles "height: 500; overflow: auto" on wrapper.
2.2 Set "height: 2000" on busy indicator div. 
2. Scroll to the bottom of the container
4. Click "Start"
5. The busy overlay is shown, but not the indicator.

**Expected behavior**
The busy indicator should be shown. The user has to scroll in the wrapper div for it to appear.

**Version**
- ids-enterprise: v.4.24.0

**Screenshots**
busy-indicator-scroll

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 79.0.3945.130 (Official Build) (64-bit)

deep7102:
Fixed position was not aligning with Busy Indicator.

Closes #3341

- Navigate to http://localhost:4000/components/busyindicator/test-scroll-in-parent.html
- Scroll inside the card area
- Use `Start/Stop` buttons to see the busy indicator
- See the position for busy indicator should not misaligned 


ericangeles:
Saw a weird bug when testing this @deep7102. 

See this captured video.

busy-indicator


I followed the steps above. After the busy indicator appear, I was scrolling while loading and then, this happened.

tmcconechy:
Also is the busy indicator supposed to go away on scroll?

deep7102:
@ericangeles - updated now to fix the point you see.
@tmcconechy - its updated now, but `No` busy indicator should not go away on scroll

PR #3593, opened 2020-03-09
3423 - Fix double change events in datepicker

### LINKED ISSUE #3423: Datepicker (ng):  triggering change event twice
**Describe the bug**
Date picker is calling the change event twice.

**To Reproduce**
1. Angular sample application.
2. Go to date picker example.
3. Do not set focus on a DP and click on the icon to open the DP and then select a date.
4. Click outside to loose the focus
5. You can see that the console message is displayed twice, one when you select a date and one when the DP input looses the focus.

**Expected behavior**
Change event should only trigger once.

**Version**
Soho ng 6.4.0

**Platform**
Windows 10 Firefox

tmcconechy:

Change fired twice in datepicker. This fixes..

Fixes #3423 

**Test One**
- go to http://localhost:4000/components/datepicker/test-change-event.html
- click in first field
- then hit tab
- no toast should show (no event should fire)

**Test Two**
- go to http://localhost:4000/components/datepicker/test-change-event.html
- click in second field
- then hit tab
- no toast should show (no event should fire)

**Test Three**
- click in first field
- type 5/5/2020
- then hit tab
- one toast should show (event should fire)
- reload
- click in first field
- select a value in the picker
- focus the input
- then hit tab
- one toast should show (event fired on selection not tab)

**Test Four**
- click in second field
- type over the value and make it 5/5/2020
- then hit tab
- one toast should show (event should fire)

Essentially only one toast should show per change.


tmcconechy:
Found a new issue here - still testing

deep7102:
All looks good, saw one thing. I guess `no toast should show (no event should fire)` if steps:
- click in second field
- don't change the value in the picker
- just click on button `Apply`
- see toast showing

ericangeles:
Am also seeing what @deep7102 mentioned. It seems that even if you didn't tab out in the field, the toast trigger immediately.

I do
- Click the first field
- Select the value in the datepicker via clicking apply button or clicking a date
- Toast immediately trigger

@tmcconechy should toast immediately trigger upon selecting date?

EdwardCoyle:
@ericangeles as long as the date changes to something other than the current value, it's valid for the change event to fire.  I think the problem @deep7102 was pointing out is when clicking "Apply" and the selected date is already in the field, it probably shouldn't fire the change event because the date wouldn't change.

BTW I'm seeing that too @tmcconechy.  Everything else looked good to me.

tmcconechy:
Ok have another look guys. I fixed the issue that if you press apply and it didnt change it sends an event.
Good catch! @EdwardCoyle @deep7102 @ericangeles

CindyMercadoReyes:
Moving this ticket to QA failed. When conducting the steps below. Two toast pop up on Win 10 IE 11. Please see screen shot attached. 
Test Three

    click in first field
    type 5/5/2020
    then hit tab
    one toast should show (event should fire)
    reload
    click in first field
    select a value in the picker
    focus the input
    then hit tab
    one toast should show (event fired on selection not tab)
bs_win10_IE_11 0 (3)

Issue #3482, opened 2020-02-12
Modal: Dropdown field in Modal is misaligned after expanding


janahintal:
**Describe the bug**
Modal: Dropdown field in Modal is misaligned after expanding

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/modal/example-validation.html
2.  Tab in all fields
3. Click Type dropdown 

**Expected behavior**
The dropdown should expand without getting misaligned 

**Version**
IDS Beta 4.251 Dev

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/74315951-825f7480-4db3-11ea-8b58-9fde0a05cfce.png)

**Platform**
All Browser, OS, and Devices

**Additional context**
N/a

tmcconechy:
I noticed another case of this. 
1. go to http://localhost:4000/components/datepicker/example-timeformat.html?locale=zh-TW
2. browser is about 1408x870 (html size)
3. Open the second datepicker
4. click the hours dropdown
5. see when it moves up its positioned wrong.

Im sure the problem is within https://github.com/infor-design/enterprise/blob/master/src/components/dropdown/dropdown.js#L1995 someplace

![Screen Shot 2020-03-04 at 9 30 55 AM](https://user-images.githubusercontent.com/814283/75889414-e5648a00-5dfa-11ea-859f-d4e2dfadab79.png)

janahintal:
The issue is now resolved. Thank you. 
Verified in https://master-enterprise.demo.design.infor.com/components/modal/example-validation.html?theme=uplift&variant=light &  https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html?theme=uplift&variant=light

PR #3596, opened 2020-03-10
3482 - Fix dropdown position when the field is required inside of modal

ericangeles:
Fixes dropdown list position when opening and if the field is required inside of modal. It also includes the datepicker issue when the list will be on top.

Closes https://github.com/infor-design/enterprise/issues/3482

- Pull this branch, run, and build the app
- Navigate to http://localhost:4000/components/modal/example-validation.html
- Tab all fields
- Open the dropdown field
- It should not overlap the `Type` label
- Also test the http://localhost:4000/components/datepicker/example-timeformat.html
- Resize the browser to 1408x870
- Open the second field datepicker
- Click the hours dropdown
- Dropdown list should stay on it's position


tmcconechy:
Still seeing the issue but i think the fix area is on the right track.

ericangeles:
@tmcconechy, I suspected that the modal causes the misalignment. I tried typing in the dropdown and it goes back to it's position. Same goes when deleting the text after searching. I pushed some changes but still it has a bug. Still figuring out some solution.

tmcconechy:
Ok, let me know if your stuck and i can have a look too @ericangeles

tmcconechy:
Need help with this @ericangeles or figured it out?

ericangeles:
> Need help with this @ericangeles or figured it out?

For sure, @tmcconechy. I've pushed some changes. You can check it out.

tmcconechy:
@ericangeles was testing this out and i found the issue. the validation is hiding the message and it makes the form move. I'm going to make a new PR with part of yours and a fix as i saw. Coming shortly....

ericangeles:
@tmcconechy got it! My first fix before was to retain the validation text so that the modal doesn't shrink. That causes the dropdown to misaligned.

PR #3602, opened 2020-03-11
NG#321 - Allow click events attached to accordion headers to propagate

EdwardCoyle:
This PR removes some code from Accordion headers that was stopping external click event handlers from propagating and ultimately running in most cases.  This was causing issues when developers attempted to run their own click handlers on accordion headers outside of the component, which was a problem in Angular environments.

Closes infor-design/enterprise-ng#321

- Pull this branch, build, and run the demoapp
- Open http://localhost:4000/components/applicationmenu/test-click-event-propagation.html
- Click any of the accordion headers.  They should correctly fire a toast containing their text contents.  The toast is driven by a simple click listener on their elements.

This test is also covered via a new e2e test.  Additionally, for further testing, you can set up the example provided by @anhallbe on the original ticket in his repo here: https://github.com/anhallbe/ids-repro/tree/accordion-click, and NPM link / symbolic link your Enterprise repo with this branch to see the changes working.

### DIFF
--- a/app/views/components/applicationmenu/test-click-event-propagation.html
+++ b/app/views/components/applicationmenu/test-click-event-propagation.html
@@ -1,0 +1,84 @@
+{{> includes/head}}
+
+<body class="no-scroll">
+  <a href="#maincontent" class="skip-link" data-translate="text">SkipToMain</a>
+  {{> includes/svg-inline-refs}}
+  <nav id="application-menu" class="application-menu is-personalizable" data-options="{ 'openOnLarge': true }">
+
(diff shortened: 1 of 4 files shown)


tmcconechy:
Looks ok, but we did have issues that clicking menu items in Angular would redirect. Thats why we had this sort of thing? WIll that occur?

EdwardCoyle:
I didn't see that occuring in my testing.  The events at the Accordion Header level are still using `preventDefault()`, which should stop any unintended routing.

tmcconechy:
ok i kind of forget the circumstances

PR #3633, opened 2020-03-18
2287 - Fix focus traps on Modals within iframe elements

### LINKED ISSUE #2287: Modal: Keyboard focus can be set outside of modal content?
I'm not entirely sure this if this is a bug. More of a question.

**Description**
When opening a modal / contextual action panel, it's possible to focus elements outside of the modal content area. However, this only seems to be possible if the app / page is within an iframe. Is this some kind of known limitation?

**To Reproduce**
* Open https://design.infor.com/code/ids-enterprise/latest/modal
* Click Show Modal button
* Mouse click outside of the modal on the overlay
* Tab a few times
Result: The Show Modal button can be focused

However, it's not possible here (non-iframed)
https://design.infor.com/code/ids-enterprise/latest/demo/components/modal/example-index

Since our app is iframed, it becomes an issue with all modals. Content outside can reached by tabbing and actions can be performed that shouldn't be possible with the modal opened.

Just wondering if you have any idea about this, or if something can be done to solve this for iframed content as well?

EdwardCoyle:
This PR fixes a focus trap bug on Modals.  Previously, when an open Modal within an iframe element lost focus, it was possible to focus an element outside of the Modal, but within the iframe, which is not supposed to occur.  This PR fixes that issue.

Closes #2287 

- Pull this branch, build, run the demoapp. The demoapp needs to be rerun to test this, since we added a new layout.
- Open http://localhost:4000/components/modal/test-iframe.html
- Click in the overlay area to defocus the Modal's input fields
- Press the "Tab" key.  Focus should occur on the first input field in the modal, and not on the "Show Modal" button beneath the overlay.

This is also covered by an e2e test.

**NOTE:** The `.page-container` section in this sample is an `<iframe>`.  Due to limitations related to sandboxing, we can't control focus outside of an `<iframe>` from within that `<iframe>`.  This PR only fixes the focus traps within the frame.


EdwardCoyle:
paging @volante007 @tralarn

EdwardCoyle:
Marking as WIP for a bit... the code I've written for retaining focus doesn't seem to work well with Modals containing components that render "focusable" elements outside the Modal boundaries.  I have to work the logic out for this a bit before the tests will pass.

Issue #3630, opened 2020-03-18
ContextualActionPanel: search icon were misaligned in Uplift


jbrcna:
**Describe the bug**
the `search` and `x` icon were misaligned when using the `uplift` theme in Firefox. 

**To Reproduce**
1. Go to https://master-enterprise.demo.design.infor.com/components/contextualactionpanel/example-markup.html?theme=uplift&variant=light
Ensure that you are using `Uplift` Theme
2. Click on Button to open CAP
3. See error. Notice that the `search` icon was misaligned, a little bit more bottom than the `x` icon

**Expected behavior**
icons should be aligned in any state

**Version**
v4.27.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/56722184/76962785-cd444d00-695a-11ea-80e2-74d0bde8ed7a.png)

**Platform**
Firefox 

**Additional context**
Uplift Theme

jbrcna:
The issue is now resolved. v4.34.0-beta0
https://4340-beta0-enterprise.demo.design.infor.com/components/contextualactionpanel/example-markup.html?theme=uplift&variant=light
<img width="1207" alt="image" src="https://user-images.githubusercontent.com/56722184/98105747-54870680-1ed3-11eb-9b15-0befedf2c0c4.png">

Issue #3586, opened 2020-03-06
Contextual Action Panel: Close button is not fully visible in mobile


jbrcna:
**Describe the bug**
the close button in the modal is partially visible

**To Reproduce**
1. Open mobile browser
2. Go to https://master-enterprise.demo.design.infor.com/components/contextualactionpanel/example-jquery.html
3. Click button to open CAP

**Expected behavior**
close button and its text should be visible

**Version**
4.27.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/56722184/76076392-bc4e2000-5fd9-11ea-9bb2-559ec7711eed.png)

**Platform**
iOS and Android

**Additional context**
In android, the issue only persists in subtle theme

PR #3641, opened 2020-03-20
3586 - Fix CAP icons issue on mobile

ericangeles:
Fixed related issues on contextual action panel button icons on mobile.

Issues:
- Misaligned search icon in uplift theme.
- Close button cuts off on mobile.

Fix:
#3630 
- Adjust margin on search icon.

#3586
- Removing the `Close` text when reaching 460px width.
- Adjust some elements to make sure that the close icon will fit inside of the buttonset.

Additional fix (Optional):
- Making sure that the button (the one will trigger the modal) will fit in smaller/mobile view, and add text ellipsis too.

Closes 
- https://github.com/infor-design/enterprise/issues/3586
- https://github.com/infor-design/enterprise/issues/3630

#3630 
- Navigate to http://localhost:4000/components/contextualactionpanel/example-markup.html?theme=uplift&variant=light
- Click `Open Markup-Based Contextual Action Panel` button
- All icons should align properly

#3586 
- Navigate to http://localhost:4000/components/contextualactionpanel/example-jquery.html
- Click `Contextual Action Panel (Uses jQuery Settings)` button
- Resize the browser down to 320px
- Play it around from 320px min to 460px max
- Notice that the text `close` will not display when reaching 460px
- `X` icon should be more visible on 320px 
- Test this also in uplift theme


app/views/components/contextualactionpanel/example-markup.html
```diff
@@ -7,7 +7,7 @@
 <div class="row top-padding">
   <div class="six columns">
     <div class="field">
-      <button type="button" id="trigger-1" class="btn-secondary">
+      <button type="button" id="trigger-1" class="btn-secondary btn text-overflow">
```
tmcconechy: This seems to be working a lot better. But wondering if there is a way to do it without adding a new class (since no one will now to do this). Maybe it works if we us last-child? Or do we think this is specific to this example?

ericangeles: Was anticipating the possibility of button that has a long text, and not really thinking to this specific example. I guess there's a less chance that this will be happen. Let me try another fix for this.

ericangeles: Thinking out loud, I guess we can use the class `.contextual-action-panel-trigger` instead creating another special class? Not sure about last-child selector as this could probably affect other buttons. What do you think?

ericangeles: I have another option. Give me a sec.

src/components/button/_button.scss
```diff
@@ -30,6 +30,7 @@ button {
   min-width: 100px;
   overflow: hidden;
   position: relative;
+  text-overflow: ellipsis;
```
tmcconechy:
```suggestion
  text-overflow: ellipsis;
```

This all looks good now. But maybe we can add this under .contextualactionpanel because this effects all buttons there could be some add side effects?

tmcconechy: And i think this is why these tests fail. https://travis-ci.com/github/infor-design/enterprise/jobs/300968163#L1373-L1378 so i would make this more specific to CAP Toolbar buttons by adding a rule there

tmcconechy:
@ericangeles noticed a couple new things now.

1) on http://localhost:4000/components/contextualactionpanel/example-jquery.html the text is now small. This is because you used the class full-width on the button. I dont think that that will work (this class is for form rows). Also would like to avoid adding new classes someone has to know to add to fix an issue.

2) On http://localhost:4000/components/contextualactionpanel/example-markup.html?theme=uplift&variant=light if you open the modal. Then close it the overlay sticks around. Compare to same page on master)

3) The text on the CAP is very small. compare the title text on the CAP on http://localhost:4000/components/contextualactionpanel/example-index.html with http://localhost:4000/components/contextualactionpanel/example-markup.html?theme=uplift&variant=light

ericangeles:
> 1. rows

I saw the full-width class used on example page here: http://localhost:4000/components/button/example-100-percent.html

But will push some updates in a bit.

ericangeles:
Noticed another issue on modal. The textarea gets cut off on a smaller size (guessing on a larger mobile device). I added this fix on the recent changes. Probably another thing to test it.

tmcconechy:
Ok i would suggest to change http://localhost:4000/components/button/example-100-percent.html as well. The font is wrong. Maybe we need special class for "full-width" buttons or just to reset the font size.

ericangeles:
Agree on adding a special case for a full width button. Something like `full-width-button` class?

tmcconechy:
Sure, yeah `button-full-width` or `button-100` or `button-responsive` we have used all three :(

Issue #3635, opened 2020-03-18
Lookup: No title header when opened inside a datagrid


sharmainerivera:
**Describe the bug**
When a lookup is opened inside a datagrid, it doesn't have a title header.

**To Reproduce**
1. Open a lookup inside a datagrid by clicking the magnifying glass 
(reference: https://master-enterprise.demo.design.infor.com/components/datagrid/test-editable-lookup.html)

**Expected behavior**
There should be a title header for a datagrid lookup just the same as a standalone lookup has. 
(reference: https://master-enterprise.demo.design.infor.com/components/lookup/example-clientside-paging-with-filtering.html)

**Version**
- ids-enterprise: v4.27.0-dev

**Screenshots**
Standalone lookup opened has title header:
![image](https://user-images.githubusercontent.com/61222084/77015911-2d65de00-69b1-11ea-875a-7025e6295c19.png)

Datagrid lookup opened has no title header:
![image](https://user-images.githubusercontent.com/61222084/77015731-97ca4e80-69b0-11ea-96fc-ffdc74ca188f.png)

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 80.0.3987.132 (Official Build) (64-bit)

**Additional Context**
Having title header for datagrid lookup will help the user in identifying the current lookup he/she is looking into especially if there are multiple lookups inside a datagrid that are editable.

tmcconechy:
- Fix the layout issue (title should be centered)
- Make the dialog pick up the editorOptions lookup title setting or default it to the column name if not provided.

janahintal:
Thi issue is now resolved. Thank you. 
Verified in https://master-enterprise.demo.design.infor.com/components/datagrid/test-editable-lookup.html?theme=soho&variant=dark across all platform. 
![image](https://user-images.githubusercontent.com/45757016/79118774-5b2c2000-7dc1-11ea-96fe-4f92a81cad0f.png)

PR #3675, opened 2020-03-30
3635 - Fix lookup modal title

ericangeles:
Currently, the `toolbar title` is not attached in the lookup modal title. Also, the position of the title and the toolbar more is not properly align.

Fix
1. Getting the toolbar title and prepend it  in the lookup modal title.
2. Adjust the position of the title to make it centered.

Closes https://github.com/infor-design/enterprise/issues/3635

- Navigate to http://localhost:4000/components/datagrid/test-editable-lookup.html
- Click the lookup icon under `Product Id` column
- You should be able to see now the `Option Title` which is 'Product' text before the number of results
- You should also noticed that the modal title and the toolbar more align properly
- If no title in toolbar, the column name will be used as title by default instead of grid toolbar title


src/components/datagrid/datagrid.js
```diff
@@ -6601,6 +6601,11 @@ Datagrid.prototype = {
       });
     }
 
+    if (this.settings.toolbar.title && $('.lookup-modal .modal-title')) {
+      const optionTitle = this.settings.toolbar.title;
```
tmcconechy:
```suggestion
      const optionTitle = this.settings.toolbar.title;
```
This looks good but a couple concerns.

1) how can we guarantee this wont effect other lookups in the page? We might need to find a way to populate this within the lookup formatter to be sure. I think its worth adding another lookup on a button the test page to test
2) Can we default the title to the column name if no toolbar title is used? 
3) i think the title should be on the column as this is the general grid tooltbar title i.e. `editorOptions: {title: 'Products'}` (default to column name otherwise)
4) slight layout issue on the results text

ericangeles: Should we use the title outside of the modal instead the column name? I just noticed that in the lookup, it uses it if there's no title in the formatter.

tmcconechy: What do you mean by title outside the modal? Was thinking it would use either the editorOptions: {title: 'something'} on the column or the column "name" if thats not there.

Its likely going to be selecting values for that column so its tied to the column data not the grid title. If that answers the question?

ericangeles: Gotcha. Yeah, I couldn't get the value of the grids title either.

tmcconechy:
Just take off the WIP when you think this is ready to review

ericangeles:
@tmcconechy I didn't see your push for baseline. I've committed the baseline and pushed but it seems it didn't go through?

tmcconechy:
Hmm wierd i dont remember doing that...

ericangeles:
Weird, I think that was me, but why it's connected to your git account?

tmcconechy:
Oh ! I know - i guess its something with the VM since i created it. We probably should "figure that out"

Well just check if its ok and if it is we can review this as a whole if its ready? (See the fits in there still) 

Also wierd is that sav file https://github.com/infor-design/enterprise/pull/3675/commits/eadf25b70f2346a9c37adf9e45857406dacc07d7

PR #3691, opened 2020-03-31
3462 - Add support for disabled callback with datepicker

### LINKED ISSUE #3462: Datepicker: disable a year within the min/max date range
The datepicker enables a min and max date to be defined, plus additional days, of week, and/or dates to be added, so they are not selectable. 

We have a case where the year 2018 and 2020 are allowed, but not 2019. The current control requires that the min and max are defined to allow 2018 and 2020, but then every day for 2019 has to be added as a disabled date.
```
disable: {
    dates: [],
    minDate: '',
    maxDate: '',
    dayOfWeek: [],
    isEnable: false,
    restrictMonths: false
  },
```

Could year(s) be disabed by adding a years array?

Adding every single date for a year to disable it seams like way too many dates.

**Additional context**
Our system requires that a user defines what years the calendar is avaliable, and not what years are disabled. An enable list may work better with our solution.

deep7102:
Added support for custom api callback to disable passed dates with Datepicker.

Closes #3462

- Navigate to http://localhost:4000/components/datepicker/example-disabled-dates.html
- Use bottom two date pickers
- Try to select any date
- Since year `2018` and `2019` are disabled
- All the dates for those years should be unavailable


ericangeles:
Looks good, but I've noticed one thing. Do we have a default value when the dates are all disabled after hitting apply button? It gives `NaN/NaN/NaN` value inside the input field. Can we disable the apply button if passed dates/years in the datepicker instead?

deep7102:
Hi @ericangeles updated now to handle invalid dates.

Issue #3070, opened 2019-11-04
ApplicationMenu: '+' icons on application menu displays as '|' when screen is zoomed out and on IE11.


vonnyw:
**Describe the bug**
'+' icon changed to '|' in application menu.

**To Reproduce**
1. Go to 'http://localhost:4000/components/applicationmenu/test-six-levels.html?theme=uplift&variant=light&colors=0563C2'
2. Expand Level 1 and a few of its children
3. Ctrl -
4. See a few of the '+' icons turned to '|'

This also happens sporadically on IE11.

**Expected behavior**

**Version**

- ids-enterprise: v4.23.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/39130590/68161109-c02e4080-ff1a-11e9-82dd-0d531bae0b71.png)

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome and IE11
 - Browser Version: Version 77.0.3865.120 (Official Build) (64-bit) and 11.1069.171340.0

tmcconechy:
@vonnyw Can you retest this on 4.24 against the new add menu changes and see if its still an issue?

https://4230-rc0-enterprise.demo.design.infor.com/components/applicationmenu/test-six-levels.html?theme=uplift&variant=light&colors=0563C2

Having some issues reproducing this and think it may be fixed.

vonnyw:
@tmcconechy This is no longer an issue in master as of today.

vonnyw:
This issue seems to resurface sporadically, please see attachment. ![image](https://user-images.githubusercontent.com/39130590/77334395-b737e300-6cf2-11ea-944d-cdb5b7fd44bd.png)

tmcconechy:
Any idea how to reproduce? Are we still supporting IE 11? I read today it was out of the supported browser list but unsure if thats official.

vonnyw:
@lipetzan  Can you please comment on IE11 support? 

Reproduce on https://master-enterprise.demo.design.infor.com/components/applicationmenu/test-six-levels.html?theme=uplift&variant=light&colors=0563C2 - click Level 1 - click Level 2 (no children with icons), notices the Level 2 with content link icon changed to "|".

No error was observed.

tmcconechy:
![Screen Shot 2020-03-23 at 12 26 16 PM](https://user-images.githubusercontent.com/814283/77338990-827f5800-6d01-11ea-86a1-b128ba31d57e.png)

[...]

PR #3737, opened 2020-04-13
3070 - Fix applicationmenu icons breaks apart when zoomed out in IE11

ericangeles:
The `plus-minus` icon in application menu breaks apart when zooming out the browser in IE11, uplift theme.

When zooming in or out, you'll encounter issues because of round and text rendering. Making sure that the elements and or layouts must survive a bit of stretching without breaking down.

Add up a `1px` on icon's height to get through when zooming out in IE11 only.

Closes https://github.com/infor-design/enterprise/issues/3070

- Navigate to http://localhost:4000/components/applicationmenu/test-six-levels.html?theme=uplift&variant=light and open in IE11.
- Expand Level 1 and it's children
- Hit `Ctrl -` to zoom out
- `+` icon should not turn into `|`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### v4.28.0 Fixes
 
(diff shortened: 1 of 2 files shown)


tmcconechy:
I think this works for me since the change is just IE

PR #3762, opened 2020-04-16
3759 - Add Fix for recursion in deepCopy and mergeUtils

### LINKED ISSUE #3759: Datagrid: Datagrid dies with Maximum call stack size exceeded
**Describe the bug**
After upgrading when passing our column object to the datagrid the datagrid now crashes. We use GWT and it has a recursive object in it that seems to be failing on the newer deepCopy util
 
**To Reproduce**
1. Copy this zipped example to datagrid/test-recursive-object
1. http://localhost:4000/components/datagrid/test-recursive-object
1. Run page
1. See error
1. can also be reproduced if loading any page and running this odd script
```
var x = {}
x.y = x;
Soho.utils.mergeSettings(window, y, {})
```

**Expected behavior**
No errors. DeepCopy should somehow not loop on this.

**Version**
- ids-enterprise: [v4.27]

**Platform**
 - Any

**Additional context**
Ms Teams
Infor SunSystems

**Example Code**
```

<div class="row">
  <div class="twelve columns">
    <div id="datagrid" class="datagrid">
    </div>
  </div>
</div>

<script id="datagrid-script">
  $('body').one('initialized', function () {

      var columns = [];

      $.getJSON('{{basepath}}api/datagrid-sample-data', function(res) {

        //Define Columns for the Grid.
        columns.push({ id: 'productId', hideable: false, name: 'Id', field: 'productId', formatter: Formatters.Text, width: 100});
        columns.push({ id: 'productName', name: 'Product Name', field: 'productName', formatter: Formatters.Hyperlink, width: 300, minWidth: 100});
        columns.push({ id: 'activity', name: 'Activity', field: 'activity'});
        columns.push({ id: 'hidden', hidden: true, name: 'Hidden', field: 'hidden'});
        columns.push({ id: 'price', align: 'right', name: 'Actual Price', field: 'price', formatter: Formatters.Decimal, numberFormat: {minimumFractionDigits: 0, maximumFractionDigits: 0, style: 'currency', currencySign: '$'}});
        columns.push({ id: 'percent', align: 'right', name: 'Actual %', field: 'percent', formatter: Formatters.Decimal, numberFormat: {minimumFractionDigits: 0, maximumFractionDigits: 0, style: 'percent'}});
        columns.push({ id: 'orderDate', name: 'Order Date', field: 'orderDate', formatter: Formatters.Date, dateFormat: 'M/d/yyyy'});
        columns.push({ id: 'phone', name: 'Phone', field: 'phone', formatter: Formatters.Text});


        columns.yk = columns;

        $('#datagrid').datagrid({
          columns: columns,
          dataset: res,
          saveColumns: false,
          toolbar: {title: 'Compressors', results: true, actions: true, rowHeight: true, personalize: true}
        });
      });
 });
</script>
```

tmcconechy:

Sunsystems has data with a big bunch of recursition. Which can be reproduced by doing this https://github.com/infor-design/enterprise/pull/3762/files#diff-78376856d7f8428d78eca7db0f2a0cbaR27.

When upgrading datagrid broke with stack heap errors. This fixes the issue by changing extend to use spread operator (more risky) and changing deepCopy to keep a map of recursive checks (not risky).

This issue should be tested thoroughly

Fixes #3759

- smoke test as many pages as you can
- try it in NG a bit
- try the following pages that represent the types of merges: 

http://localhost:4000/components/datagrid/example-frozen-columns.html
http://localhost:4000/components/applicationmenu/test-click-event-propagation
http://localhost:4000/components/datagrid/test-frozen-columns-with-colspan
http://localhost:4000/components/datagrid/test-recursive-object.html
http://localhost:4000/components/popover/example-index


tmcconechy:
Seems to be giving a lot of errors i wasnt expecting. Looking into it

src/utils/utils.js
```diff
@@ -645,12 +647,80 @@ DOM.focusableElems = function focusableElems(el, additionalSelectors = [], ignor
...
+  if (!proto) {
+    return true;
+  }
+
+  return obj !== null && typeof (obj) === 'object' && Object.getPrototypeOf(obj) === Object.prototype;
+};
+
+/**
+ * Object deep copy and merge. Replaces jQuery.extend without the true option.
+ * @param {boolean|object} deep If true, the merge becomes recursive (aka. deep copy). Passing false for this argument is not supported.
```
EdwardCoyle, line 670: @tmcconechy the `true` here actually does work, correct?  Need the docs to match either way.

src/utils/utils.js
```diff
@@ -645,12 +647,80 @@ DOM.focusableElems = function focusableElems(el, additionalSelectors = [], ignor
...
+utils.extend = function extend() {
+  // Variables
+  let extended = arguments[0] || {};
+  let deep = false;
+  let i = 0;
+
+  // Check if a deep merge
+  if (Object.prototype.toString.call(arguments[0]) === '[object Boolean]') {
+    deep = arguments[0];
+    extended = {};
```
EdwardCoyle, line 685: Does `extended` need to change to take `arguments[1]` like [jQuery's takes](https://github.com/jquery/jquery/blob/master/src/core.js#L130)?  Just wondering if setting `true` would actually cause the next argument to be the base object (could probably write a test for this)

src/utils/utils.js
```diff
@@ -1108,12 +1178,22 @@ utils.getScrollbarWidth = function () {
  * @returns {array|object} The copied array or object.
  */
 utils.deepCopy = function (arrayOrObject) {
+  const references = new Map();
```
EdwardCoyle, line 1181: Will this be OK in IE11 for now? I tested all the links provided in IE11 and it seemed to all work https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility

tmcconechy: Yeah IE seems ok to me. I think since we are just using the constructor. To really make sure this runs just try http://localhost:4000/components/datagrid/test-recursive-object.html in IE 11 (and it seems ok

src/utils/utils.js
tmcconechy: Yes, it works i just did it a different way by extending {} as it runs through, i added a few tests check below.

src/utils/utils.js
tmcconechy: Fixed!

tmcconechy:
@EdwardCoyle added some tests and fixed the comments up a bit

Issue #3479, opened 2020-02-11
Modal: A large title can scroll the modal buttons off the page, depending on screen size


Sovia:
**Describe the bug**
A modal with a large title will scroll the modal buttons off depending on the size of the window.

**To Reproduce**
1. Copy [test-long-title-large-content.zip](https://github.com/infor-design/enterprise/files/4188259/test-long-title-large-content.zip) into /components/modal/
2. Click the Show Modal button
3. Change the window size to various sizes, use the iPhone 5 emulator

**Expected behavior**
There should be a max height that the title can take up, and then be ellipses. 

**Version**

- ids-enterprise: v4.26.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/22107636/74278229-81a8de80-4cde-11ea-9120-f6f28f4c7895.png)
![image](https://user-images.githubusercontent.com/22107636/74278262-8f5e6400-4cde-11ea-9f1f-0835c8194417.png)
![image](https://user-images.githubusercontent.com/22107636/74278303-a0a77080-4cde-11ea-9cfb-cf56595fd338.png)
![image](https://user-images.githubusercontent.com/22107636/74278365-b87ef480-4cde-11ea-838c-aba0a6691b06.png)

**Platform**
 - Any window with limited space

tmcconechy:
- might be able to update the function, that sets the height there. that is JS in modal.js
- also maybe make a max text to where after it goes to ellipsis (if this is possible)

janahintal:
The issue is still failing in Windows 10 IE 11.  

The tooltip is not working and the title as 1 line only. 
https://master-enterprise.demo.design.infor.com/components/modal/test-long-title-large-content.html?theme=uplift&variant=light
![image](https://user-images.githubusercontent.com/45757016/79452460-c58bcd00-801a-11ea-9eb1-0705f8419556.png)

jbrcna:
The issue is now resolved. moving this ticket to Done. Thank you
v4.28.0-dev
https://master-enterprise.demo.design.infor.com/components/modal/test-long-title-large-content.html?theme=uplift&variant=light
**IE11**
![image](https://user-images.githubusercontent.com/56722184/79978241-2f501f00-84d2-11ea-9c02-a588590ca12d.png)

PR #3763, opened 2020-04-17
3479 - Fix multiline ellipsis in IE

ericangeles:
This is a follow up fix for https://github.com/infor-design/enterprise/issues/3479. The `-webkit-line-clamp` is not supported in IE11, so I created a pure CSS that will fix this issue. 

Closes https://github.com/infor-design/enterprise/issues/3479

- Open in IE11 browser (via browserstack or windows) and navigate to http://localhost:4000/components/modal/test-long-title-large-content.html
- Click `Show Modal` button
- The title should be multi line with ellipsis
- Tooltip should also work

### DIFF
--- a/src/components/modal/_modal.scss
+++ b/src/components/modal/_modal.scss
@@ -542,6 +542,43 @@ html.ie {
   }
 }
 
+// Solution for multi line with ellipsis
+.ie,
+.ie11 {
+  .modal {
+    &:not(.display-fullsize):not(.about) {
+      .modal-content {
+        h1 {
+          line-height: 1.2em;
+          margin-right: -1em;
+          max-height: 2.4em;
+          overflow: hidden;
+          padding-right: 1em;
+          position: relative;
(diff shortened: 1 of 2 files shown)


src/components/modal/_modal.scss
```diff
@@ -542,6 +542,43 @@ html.ie {
   }
 }
 
+// Solution for multi line with ellipsis
+.ie,
+.ie11 {
```
tmcconechy, line 547: Do you need this just for IE11 or Edge too? Is so maybe just need one rule here

ericangeles: @tmcconechy, not sure if the Edge is attaching the `.ie` class. I was under impression that the `.ie` is a lower version (IE10 and below) so I included it. You want me to remove it?

tmcconechy: Yeah i think you just need ie11

PR #3786, opened 2020-04-22
281 - Add new empty message design in modal

ericangeles:
Provided a new design for error empty message inside of modal. This design will probably use for error pages like HTTP 500 or HTTP 400.

Closes https://github.com/infor-design/design-system/issues/281

- Navigate to http://localhost:4000/components/emptymessage/example-modal-empty-message.html
- Click `Modal Empty State Message` button
- You'll see an error empty message design


tmcconechy:
@ericangeles this is looking pretty good but a couple suggestions.

- i was thinking this would be a new component in components folder called "error-page" but it doesnt seem to be logically a emptymessage 
- in an example there (example-index) just show the page with the error on it we dont need to animate in a modal the page should be static
- so presumably this can be all css component
- i don't think we should add a new primary button style (different colors and radius 8px) - so just use the normal primary button and not introduce any new styles
- im also getting a scroll bar on the dialog 
- icons seem a bit large we reduced this on another ticket... may want to tweak that a bit.

ericangeles:
> @ericangeles this is looking pretty good but a couple suggestions.
> 
> * i was thinking this would be a new component in components folder called "error-page" but it doesnt seem to be logically a emptymessage
> * in an example there (example-index) just show the page with the error on it we dont need to animate in a modal the page should be static
> * so presumably this can be all css component
> * i don't think we should add a new primary button style (different colors and radius 8px) - so just use the normal primary button and not introduce any new styles
> * im also getting a scroll bar on the dialog
> * icons seem a bit large we reduced this on another ticket... may want to tweak that a bit.

This is all noted, @tmcconechy ! Thank you.

I was following the design that was provided, but yeah, will work this tomorrow. This is a big help! :) I wasn't 100% sure at first where to put this design so I come up putting in the empty message. I'll overhaul the whole design and code.

ericangeles:
@tmcconechy maybe a new PR will do?

tmcconechy:
Yeah i think the main thing we cant change the consistency of the buttons so we fit that into our system was probably a mistake in the design

ericangeles:
Gotcha. Will close this now.

tmcconechy:
ok look forward to the next one.

PR #3787, opened 2020-04-22
3779 - Fix datagrid filter conditions order and multiple selection

### LINKED ISSUE #3779: Datagrid: Filter conditions Leaves Two selected if you just reorder.
**Describe the bug**
If you rearrange the items in filterConditions and leave the default IN you get two selected in the list visually.

**To Reproduce**
1. Go to http://localhost:4000/components/datagrid/example-custom-filter-conditions.html
2. Edit that page so one column has `filterConditions: ['equals', 'does-not-equal', 'contains', 'does-not-contain', 'is-empty', 'is-not-empty', 'end-with', 'does-not-end-with', 'start-with', 'does-not-start-with']`
3. open filter list (not that it is showing equals correctly but)
4. See error - two are selected

**Expected behavior**
Only equals is selected

**Version**
- ids-enterprise: [e.g. v4.27.0]

**Screenshots**
Screen Shot 2020-04-20 at 2 19 07 PM

**Platform**
 - Any

jmacaluso711:
Updates the `filterButtonHtml` function to loop through the `filterConditions` if they are present. Also adds a function (`formatFilterText`) to format the filter text in order properly generate the filter popupmenu list.

closes #3779 

- pull branch
- go to http://localhost:4000/components/datagrid/example-custom-filter-conditions.html
- open second column filter list and make sure the list follows the order of the filterCondition array  `filterConditions: ['equals', 'does-not-equal', 'contains', 'does-not-contain', 'is-empty', 'is-not-empty', 'end-with', 'does-not-end-with', 'start-with', 'does-not-start-with']`
- make sure only the first filter condition is set to true.


tmcconechy:
@jmacaluso711 i think what would be better here instead of having to manually change it to test it is to just update the example to have it in it.

1. the test is on the first column only now so just add that full list for the second column inside the example so we dont have to change it
2. then make a second similar to test to https://github.com/infor-design/enterprise/blob/master/test/components/datagrid/datagrid.e2e-spec.js#L141 so we have a test for this working

PR #3793, opened 2020-04-23
801 - Fix nested caps and memory leaks

### LINKED ISSUE #3785: Popupmenu: "attachToBody" cause memory leak when used in a Modal / CAP
**Describe the bug**
The default value for `attachToBody` for popupmenu was changed from false to true. With this, we've seen some memory leaks when popupmenus are used in Modals / CAPs. If destroying the modal when closed, some popupmenus remain in DOM and increase +1 each time the modal is re-opened.

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/contextualactionpanel/test-destroy.html
2. Repeat "Show me potato salad" -> "Destroy CAP" -> "Invoke CAP" a few times

**Expected behavior**
When CAP is destroyed, all popupmenus within should be destroyed & removed from DOM

**Actual behavior**
No. of popupmenus increase +1 on body each time CAP is re-invoked and opened. In this example, it seems it's the toolbar more menu that doesn't get destroyed.

**Version**
- ids-enterprise: v4.27.3

**Additional context**
In older versions, when`attachToBody` was false by default, any menus within a CAP / modal got destroyed since the whole modal is removed from DOM.
4.24.0
https://design.infor.com/code/ids-enterprise/latest/demo/components/contextualactionpanel/test-destroy.html

Perhaps there are other scenarios where a popupmenu fails to destroy, but for now this is the one we've seen.

tmcconechy:

CAP did not work with nested CAPs anymore. I went back to check if this worked all the way to 4.18 on the new example here and could not find any evidence it ever worked. But maybe NG worked as it destroyed better.

Fixes #3785
Fixes infor-design/enterprise-ng#801

**Test 1**
- go to http://localhost:4000/components/contextualactionpanel/example-index.html
- this example and all examples in http://localhost:4000/components/contextualactionpanel should work ok still as before

**Test 2**
- go to http://localhost:4000/components/contextualactionpanel/test-destroy.html
- repeatedly open the cap and then destroy it
- then check the DOM - should not be an set of popupmenu wrappers left

**Test 3**
- go to http://localhost:4000/components/contextualactionpanel/test-nested.html
- click button one then close the cap
- click button two -> should be a different CAP
- click button one then the nested cap buttons and make sure nesting works

**Test 4**
- bring this branch over to enterprise-ng (Master has a change)
- go to http://localhost:4200/ids-enterprise-ng-demo/contextual-action-panel
- click panel
- click nested CAP -> should open a new cap -> close it
- click panel -> click save - should open a new cap as well

### DIFF
--- a/app/views/components/contextualactionpanel/test-destroy.html
+++ b/app/views/components/contextualactionpanel/test-destroy.html
@@ -13,38 +13,38 @@
 
     <div class="field">
       <button type="button" id="contextual-panel-trigger" class="btn-secondary contextual-action-panel-trigger">
-        <span>Show Me Potato Salad</span>
+        <span>Show CAP</span>
       </button>
       <div class="contextual-action-panel">
-  	    <div class="row">
-    			<div class="twelve columns">
-    				<div class="field">
-    				  <label for="cool-dropdown">Cool Dropdown</label>
-    				  <select class="dropdown" id="cool-dropdown">
-    					<option value="1">One</option>
-    					<option value="2">Two</option>
-    					<option value="3">Three</option>
-    					<option value="4">Four</option>
-    				  </select>
-    				</div>
+        <div class="row">
+        <div class="twelve columns">
+        <div class="field">
+        <label for="cool-dropdown">Cool Dropdown</label>
+        <select class="dropdown" id="cool-dropdown">
+        <option value="1">One</option>
+        <option value="2">Two</option>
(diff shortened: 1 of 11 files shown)


tmcconechy:
@EdwardCoyle Fixed the error. That line wasnt needed at all. Just a copy paste issue/carryover

PR #3748, opened 2020-04-14
3702 - Add support to ellipsis for yaxis labels with bar and line chart

deep7102:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
Added support to ellipsis for yaxis labels with Bar and Line Chart.

**Related github/jira issue (required)**:
Closes #3702

**Steps necessary to review your pull request (required)**:
- Pull this branch and build/run the demo app
- Navigate to:
- http://localhost:4000/components/bar/test-axis-y.html
- http://localhost:4000/components/line/test-axis-y.html
- Resize browser window to small viewport
- See the y-axis 1st label, should show `...` three dots ellipsis
- Hover on it, tooltip should show with full content for this label
- Triple click on `...` three dots, should select the text
- Right click then copy, or use keys `cmd + c` to copy
- Paste copied content to somewhere
- See it should be full content for this label

FYI:
- Copy/paste feature is not supported in IE11

**Included in this Pull Request**:
- [x] A note to the change log.

Issue #2663, opened 2019-08-19
Custom Builds:  Improve Code Splitting in SASS files


EdwardCoyle:
This issue is dependent on #1038 being complete.

**Is your feature request related to a problem? Please describe.**
To better ensure the integrity of a custom IDS build, we need to revisit our SASS source code and label all IDS components with a dependency structure. This structure is similar to how we currently enable dependency in the Javascript parts of the components, by importing the dependencies before they are used.

All component SASS files should be reviewed to add dependencies on lower-level components where they make sense.  As an example, the Dropdown component implements the styles for the Input component.  Therefore, the `_dropdown.scss` file should include an `@import` statement at the top referencing `_input.scss`.

**Describe the solution you'd like**
- [ ] Review the CSS styles as a whole and move source code to separate files where it makes sense.
- [ ] In mid-to-complex components, `@import` statements for lower level component dependencies should be included at the top of the file wherever necessary (see the [Listview](https://github.com/infor-design/enterprise/pull/2661/files#diff-553b4ea2f951aa47a5d0cb1c1f8501a0) for an example of how this works).  For example, the Popdown component [overrides existing Card/Widget styles](https://github.com/infor-design/enterprise/blob/master/src/components/popdown/_popdown.scss#L79-L97), so it should include a dependency for the original Card/Widget styles.
- [ ] Explore better splitting of variables and mixins (may include removing extra ones).

Feel free to comment on this issue for any other code-splitting ideas.

**Additional context**
This type of SASS code splitting is enabled via a Node-Sass plugin we've recently added called [`node-sass-once-importer`](https://github.com/maoberlehner/node-sass-magic-importer/tree/master/packages/node-sass-once-importer).  See that Github repo for an explanation of how it works for a better understanding of how to tackle this issue.

tmcconechy:
Is this different than https://github.com/infor-design/enterprise/pull/2661 or are they just not linked?

EdwardCoyle:
It's the follow up task to #2661.  Doing this code splitting is a pretty big task on its own.

tmcconechy:
This wont be needed with the new project, so descoping for now

PR #3797, opened 2020-04-23
3784 - Fix failing Custom Builds when including Datagrid

### LINKED ISSUE #3784: General: Error building datagrid
**Describe the bug**
Including datagrid in a custom build is giving an error.

**To Reproduce**
1. Pull master
2. run `npm run build -- --components=button,busyindicator,datagrid`
3. or `npm run build -- --components=datagrid`
4. See 2 errors `Error: no mixin named required-indicator
    at Object.module.exports.renderSync (/Users/tmcconechy/Dev/master4/node_modules/node-sass/lib/index.js:441:16)` and `Error: 'DatagridEditors' is not exported by src/components/datagrid/datagrid.editors.js, imported by temp/components.js`

**Expected behavior**
Should build with datagrid. We can rename exports if needed.

**Version**

- ids-enterprise: [e.g. v4.2.7]

**Platform**
 - Mac and windows

**Additional context**
Infor S3

EdwardCoyle:
This PR fixes an issue with custom builds including the Datagrid, where the builds would not complete because of dependencies not correctly being included and written as exports in the wrong places.  The build script has been adjusted, and inter-library dependencies have been corrected, which should now allow correct custom builds of the Datagrid.

Additionally, I moved some of the ESLint rules around in the app/scripts/test code area to make more sense (some of the errors I was seeing in the scripts folder didn't make sense for a Node environment, such as IE11 compatibility).  The linter should pass this PR before we merge.

Closes #3784

- Pull this branch
- In the terminal, run `npm run build -- --components=datagrid --dry-run`
- When the build completes, open the `enterprise/temp` folder.
- Find `components.txt` - it should only contain "datagrid"
- Find `source.txt` - In the "JS Source Code" section, all `datagrid.*.js` files should be present (the jQuery one sits below in the "jQuery Source Code" section.
- Find `components.js` - It should only export ONE file, which is `Datagrid` in the "Complex" section.
- Find `rules.js` - This should be exporting all the extra Datagrid files.
- Now, run `npm run build -- --components=datagrid` In the terminal
- When the build completes, launch the demoapp with `npm run quickstart` (no library builds)
- Navigate to http://localhost:4000/components/datagrid/example-index.html?layout=nofrills.  Test that all Datagrid features work.  Test the toolbars, popmenus, all Datagrid Editors/Formatters, etc.
- Check Uplift theme for the same things: http://localhost:4000/components/datagrid/example-index.html?layout=nofrills&theme=uplift

NOTE: Use the `nofrills` layout when testing all pages because the Datagrid being built by itself doesn't include the Header component or the Skiplink Component, so pages will be broken if you use the standard layout.


EdwardCoyle:
Note that some of the work in this PR is actually related to #2663, but I would say not to close that issue just yet, since we need to include the SASS dependencies throughout, not just on Datagrid.

EdwardCoyle:
@tmcconechy Maybe we should look at this one again, per the build issues on #3748?

EdwardCoyle:
@tmcconechy yeah the temp folder has all kinds of "loose" stuff inside that gets cleaned up when a build "really" completes.  This is where all the parsing happens at build time.  `--dry-run` just lets you see what's in there before a build occurs.  If you see an error in a compiled custom build, you can usually check all the automated import/export creation here to double-check that everything's right.

If you see any weird errors in the component(s) included in that custom particular, then there may be an issue, but otherwise I wouldn't worry about these too much.

PR #3821, opened 2020-04-28
804 - Fix change event with autocomplete

### LINKED ISSUE #804: Autocomplete doesn't fire change event when a value is selected
**Describe the bug**
When binding to the change event of an autocomplete the change is not fired if the user selects a value from the dropdown list.

**To Reproduce**
1. Go to http://localhost:4000/components/autocomplete/example-index
2. Set up a change handler `$("#autocomplete-default").on('change', function() { alert($(this).val())})`
3. Type in a partial match to the value 'a'
3. Arrow down to select, notice change fires with the typed in value
4. Select a value from the list

**Expected behavior**
change event fires

**Platform**
 - Windows 7
 - All Browsers

### LINKED ISSUE #3548: FileUploadAdvanced: Need an event for the delete of an attachment
**Describe the bug**
Need an event added to SohoFileUploadAdvancedComponent when the x is clicked to remove an attachment.

**To Reproduce**
1. Select an attachment(s).
2. There is an x on the left if attachment is in progress.    There is an x on the right if complete.
3. When the x is clicked need an event that is available.
4. Expose this in Angular too

**Version**
4.13

deep7102:
Fix change event was not firing while navigating list with Autocomplete.

Closes #804 

- Navigate to http://localhost:4000/components/autocomplete/test-on-change.html
- Open developer tools
- Type `a` and an list with available option will open
- Press arrow down/up keys to navigate list
- See change event in console while navigating list (First time it looks fire twice, but its not -it fire one for search field leave and one for content change)

### DIFF
--- a/app/views/components/autocomplete/test-on-change.html
+++ b/app/views/components/autocomplete/test-on-change.html
@@ -1,0 +1,18 @@
+<div class="row">
+  <div class="twelve columns">
+    <div class="field">
+      <label for="autocomplete-default">States</label>
+      <input type="text" autocomplete="off" class="autocomplete" data-options="{source: '{{basepath}}api/states?term='}" placeholder="Type to Search" id="autocomplete-default"/>
+    </div>
+  </div>
+</div>
+
+<script>
+  $('#autocomplete-default')
+    .on('selected', function (e) {
+      console.log('selected:', $(this).val());
+    })
+    .on('change', function (e) {
+      console.log('change:', $(this).val());
+    });
+</script>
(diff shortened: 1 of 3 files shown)


PilotBob:
Shouldn't it only fire change when a value is selected from the list, vs just moving through them?

tmcconechy:
When you move through them it does change the field value so thats why it needs to fire in that case.

PilotBob:
> When you move through them it does change the field value so thats why it fires as well.

i c. Well, as long as it fires if you select from the mouse too. We've worked around this by adding our own change tracking thus far.

tmcconechy:
Yes both the change event and the selected event fire (BTW we noted you maybe could have used the selected event as well for list selection).

PR #3834, opened 2020-04-29
3826 - Fix Accordion lifecycle to allow individual header updates without a full teardown [v4.28.x]

### LINKED ISSUE #3826: Accordion: Unable to dynamically add accordion items
**Describe the bug**
Adding a new accordion header/panel then calling .updated($headers) no longer works. This has worked since 4.x came out, but I think it broke with commit 312ec60995e2fe1e2ecc42b15e3a12e1ac8e50ba in February.

**To Reproduce**
1. Add this file to the app/views/components/accordion directory: https://gist.github.com/awbuboltz/fc642f679f10c27df861fd2086d4bc99
2. Click the "Add new favorite" button
3. See the console error

**Expected behavior**
The newly added favorite menu item should be properly initialized as a menu item, including key handling and CSS.

**Version**

- ids-enterprise: 4.28

**Platform**
 - OS Version:. Windows 10
 - Browser Name: Chrome

**Additional context**
When passing new headers to .updated(), eventually teardown() gets called. Teardown now does a delete on this.headers, this.anchors, this.panes, and this.contentAreas. Right after teardown(), init() is called, which tries to access those variables.. which were just deleted.

EdwardCoyle:
This PR fixes a regression in the Accordion that was preventing consumers from updating individual accordion headers without doing a full teardown of the entire accordion.  Script errors were being thrown when attempting this previously.

Additionally, there is now a demo for this test case, along with new e2e/functional tests.

Closes #3826

- Pull this branch, build, and run the demoapp
- Open http://localhost:4000/components/accordion/test-add-dynamically.html?layout=nofrills
- Expand the Accordion Header labeled "Favorites"
- Click the "Add new favorite" button several times.
- See that new accordion headers should populate underneath the Favorites header.  If you open a developer tools console, there should be no Javascript errors.
- Try keyboard navigation on the new headers - it should be possible to use the Arrow keys and Tab/Shift+Tab on the new ones.

### DIFF
--- a/app/views/components/accordion/test-add-dynamically.html
+++ b/app/views/components/accordion/test-add-dynamically.html
@@ -1,0 +1,64 @@
+<div class="row">
+  <div class="six columns">
+    <p>Github Issue: <a class="hyperlink" href="https://github.com/infor-design/enterprise/issues/3826" target="_blank">#3826</a></p>
+    <p>In this test, clicking the "Add to favorites" button will add a new menu item in the Favorites content panel.</p>
+  </div>
+</div>
+
+<div class="row top-padding">
+  <div class="six columns">
+    <div class="field">
+      <button class="btn" id="addFavs">Add new favorite</button>
+    </div>
+
+    <div data-init="false" id="test-accordion" class="accordion panel inverse">
+
+      <!-- general -->
+      <div class="accordion-header">
+        <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
+          <use href="#icon-user"></use>
+        </svg>
+        <a href="#"><span>General</span></a>
+      </div>
+      <div class="accordion-pane">
+        <div class="accordion-header">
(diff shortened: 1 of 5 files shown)


EdwardCoyle:
@awbuboltz @marknolaninfor if possible, would you mind checking this branch out in your dev environments and let me know if this fixes the issues you were seeing?

marknolaninfor:
@EdwardCoyle 

Thanks for your prompt response.

This may be a better example to show the change to the source example, and how it can be used to limit the teardowns and improve the performance of large accordions. 

    const sourceContainer = {
      source: (ui, response) => {
        ui.pane.empty().append($('<div class="accordion-content is-disabled"><p>Loading</p></div>'));
        callback(true);
        setTimeout(() => {
          ui.pane.empty().append(html);
          callback(false, ui.pane.find('.accordion-header'));
        }, 300);
      }
    };

Issue #2695, opened 2019-08-27
Tabs: If info icon is used rather than error icon , it is misaligned


jeffuycoque:
**Describe the bug**
On our application, if there is a field that has an info message, the tab would have an info icon right next to it. However the info icon is misaligned when it is rendered 

**To Reproduce**
1. Go to http://master-enterprise.demo.design.infor.com/components/tabs/test-modal-error.html
2. Click Add Context
3. Trigger Validation By inputting a blank value on Last Name Field
4. Right Click Error Icon on the Tab and Click Inspect
![image](https://user-images.githubusercontent.com/49975141/63737964-c10f7a00-c8ba-11e9-8d72-7fb99bdfb83d.png)
5. Change icon-error to icon-info
icon-error:
![image](https://user-images.githubusercontent.com/49975141/63737995-e2706600-c8ba-11e9-87f7-ce62b7fe25bc.png)

icon-info
![image](https://user-images.githubusercontent.com/49975141/63738046-121f6e00-c8bb-11e9-8d3e-5d3d9b207b60.png)

**Expected behavior:**
Info Icon would be aligned

**Actual behavior:**
![image](https://user-images.githubusercontent.com/49975141/63738017-fcaa4400-c8ba-11e9-8b52-186529e85250.png)

It seems that the info and error icons have different sizes, the error icon seems to be a little smaller which may be the reason why the issue is not reproducible when tab has an error.

Platform
OS Version: Windows 10
Browser Name: Google Chrome

jeffuycoque:
@tmcconechy this is also reproducible using alert icon
![image](https://user-images.githubusercontent.com/49975141/64302145-94431d00-cfb4-11e9-898a-28e56e09e6d2.png)

![image](https://user-images.githubusercontent.com/49975141/64302173-ad4bce00-cfb4-11e9-920c-bd2e4b03185e.png)

jbrcna:
**QA Passed**
v4.29.0-dev
https://master-enterprise.demo.design.infor.com/components/tabs/test-modal-error.html

![image](https://user-images.githubusercontent.com/56722184/81406314-e9e74f00-916b-11ea-9fb5-d4af44065f2a.png)

PR #3860, opened 2020-05-04
2695 - Fix info and alert icons on tabs and inside of modal

ericangeles:
The info and alert icons on tabs and inside of modal doesn't align properly the way the error icon does. Some CSS rules inside of the tabs affects the icons that was not in modal. Moving some css rules to the modal as a parent instead.

Closes https://github.com/infor-design/enterprise/issues/2695

1.)
- Navigate to http://localhost:4000/components/tabs/test-modal-error.html
- Click `Add Context` button
- Tab out in input field to trigger the validation error
- Error icon will display on tab section
- Inspect the element of the icon
- Change the class `icon-error` to `icon-info`, then after `icon-alert`
- Both `icon-info` and `icon alert` should align properly
- Test also on uplift theme

2.)
- Navigate to http://localhost:4000/components/tabs/example-dropdown-tabs.html
- Click on `Paper` tab, then click the `Plates` to show the data that has input field
- Trigger the validation error by tabbing out
- Error icon will show on tab
- Inspect the element of the icon
- Change the class `icon-error` to `icon-info`, then after `icon-alert`
- Both `icon-info` and `icon alert` should align properly
- Try also to trigger the focus of tab
- Icon should have more space and not close to the border tab

Additional Test at 2.)
- Inspect the element of sub list of tab that has icon when validation trigger. See image below
err-icon

- Change the class `icon-error` to `icon-info`, then after `icon-alert`
- Both `icon-info` and `icon alert` should align properly
- Test this also on uplift theme


tmcconechy:
I agree with @jmacaluso711 on the list icon. Also i think the setup on the dropdown tabs could use a bit of adjustment.

Screen Shot 2020-05-04 at 2 04 52 PM

- as John said icon in the list should be pushed down
- the arrow should be more centered
- the icon should be on the other side of the arrow and could be smaller. This is it on the dropdown component..

Screen Shot 2020-05-04 at 2 06 38 PM

Can live with these but minor improvements would be nice

ericangeles:
All good now, @jmacaluso711 @tmcconechy. Thank you.

PR #3890, opened 2020-05-06
1458 - Splitter drag handle icon

### LINKED ISSUE #1458: Splitter: Character sometimes renders incorrectly
**Describe the bug**
Very occasionally the character used for the splitter drag handle renders incorrectly.

**To Reproduce**
Cannot consistently reproduce, happens intermittently, perhaps latency dependent? 

**Expected behavior**
Ideally, this character should be replaced with an SVG icon that wouldn't be susceptible to character encoding problems or other text-related oddities. 

**Version**
Has happened in all versions

**Screenshots**
image

**Platform**
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 71.0.3578.98 (Official Build) (64-bit)

jmacaluso711:
Fixes an issue where the drag handle characters sometime render incorrectly. Changes the drag handle to an svg icon.

closes #1458 

- pull and build branch
- go to http://localhost:4000/components/splitter and check the examples. Make sure the drag handle is rendering properly.

### DIFF
--- a/src/components/splitter/splitter.js
+++ b/src/components/splitter/splitter.js
@@ -66,6 +66,7 @@ Splitter.prototype = {
     const direction = s.axis === 'x' ? 'left' : 'top';
     const thisSide = parent.is('.content') ? parent.parent() : parent;
     const defaultOffset = 299;
+    const dragHandle = $('<div class="splitter-drag-handle"><svg class="icon"><use xlink:href="#icon-drag"/></svg></div>');
     let w = parent.width();
     let parentHeight;
 
@@ -77,6 +78,7 @@ Splitter.prototype = {
     this.isSplitterRightSide = splitter.is('.splitter-right') || (s.axis === 'x' && s.side === 'right');
     this.isSplitterHorizontal = splitter.is('.splitter-horizontal') || s.axis === 'y';
     s.uniqueId = utils.uniqueId(this.element, 'splitter');
+    dragHandle.appendTo(splitter);
 
     if (this.isSplitterRightSide) {
       const thisPrev = thisSide.prev();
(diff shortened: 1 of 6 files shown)


src/components/splitter/splitter.js
```diff
@@ -66,6 +66,7 @@ Splitter.prototype = {
     const direction = s.axis === 'x' ? 'left' : 'top';
     const thisSide = parent.is('.content') ? parent.parent() : parent;
     const defaultOffset = 299;
+    const dragHandle = $('<div class="splitter-drag-handle"><svg class="icon"><use xlink:href="#icon-drag"/></svg></div>');
```
tmcconechy, line 69: I think just change this to use the `$.createIcon(` function. This will fix a few issues with just plain SVG's like basetag in the head and some IE append bugs. So we filter all icon creation through that function

ericangeles:
All good to me. Just noticed a minor color issue when you hover, on high contrast variant uplift theme. The icon is really hard to see. This issue occurs to all pages.

draggable

tmcconechy:
yeah i think just dial back the colors in uplift high contrast so its the same or similar to:

http://localhost:4000/components/splitter/example-index.html?theme=soho&variant=contrast

Its actually the opposite of higher contrast... I would say turn the backcolor in soho contrast down one on the drag state then use the same colors in uplift contrast

jmacaluso711:
Using the tokens.. so they are slightly different between soho and uplift, but should be pretty close

tmcconechy:
The only thing i still see is on http://localhost:4000/components/splitter/example-index.html?theme=uplift&variant=light (uplift theme). The icon is a lot smaller than it was before on master.

Its a little off center too there but i think if we make it closer to the same size as before it would be centered?

PR #3913, opened 2020-05-12
1423 - Update handle on touch devices with swaplist

deep7102:
Updated to use whole `LI` item instead of tree dot icon `handle` on touch devices with swaplist

1423

- Use iOS device
- Navigate to: http://localhost:4000/components/swaplist/example-index.html
- Select multiple items from one of the lists
- Use whole LI item and drag them over to the next one
- Items will move
- Select moved items again and drag them back to same list
- Try again as many times, drag and drop should work fine

### DIFF
--- a/src/components/swaplist/swaplist.js
+++ b/src/components/swaplist/swaplist.js
@@ -438,8 +438,8 @@ SwapList.prototype = {
     if (this.isDragAndDropSupports) {
       // Use Handle if available
       this.handle = ul.first().attr('data-swap-handle');
-      this.handle = $(this.handle, ul).length > 0 ? this.handle : null;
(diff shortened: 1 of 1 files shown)


src/components/swaplist/swaplist.js
```diff
@@ -438,8 +438,8 @@ SwapList.prototype = {
     if (this.isDragAndDropSupports) {
       // Use Handle if available
       this.handle = ul.first().attr('data-swap-handle');
-      this.handle = $(this.handle, ul).length > 0 ? this.handle : null;
-      // this.handle = (!this.isTouch && $(this.handle, ul).length > 0) ? this.handle : null;
+      // this.handle = $(this.handle, ul).length > 0 ? this.handle : null;
```
tmcconechy, line 441: Just remove this?

Issue #3897, opened 2020-05-07
Datagrid: display issue with modal dialog containing a datagrid and search form


claudenbach:
Describe the bug
A datagrid doesn't expand to fill the table width when the visibility of a search form within the modal dialog is toggled to show and hide.

To Reproduce
1. Using the attached test-modal-datagrid-search-form.html ([Git-3897.zip](https://github.com/infor-design/enterprise/files/4596205/Git-3897.zip))
2. Click the filter icon on the toolbar 
3. Width of the modal dialog grows - the grid table doesn't expand
4. Click the filter icon on the toolbar

Version
v4.29.0-dev

Screenshots

![image](https://user-images.githubusercontent.com/39804038/81352701-cac5c000-908c-11ea-8095-15cc0f234797.png)

![image](https://user-images.githubusercontent.com/39804038/81352721-d2856480-908c-11ea-9ac8-2b2c8a090ff8.png)

claudenbach:
[Git-3897.zip](https://github.com/infor-design/enterprise/files/4596205/Git-3897.zip)

claudenbach:
This is how it looks in Landmark...

![image](https://user-images.githubusercontent.com/39804038/81353070-c0f08c80-908d-11ea-84c8-fa274400b474.png)

![image](https://user-images.githubusercontent.com/39804038/81353095-cfd73f00-908d-11ea-98f9-9d3489a1cc74.png)

tmcconechy:
Have to test `-webkit-fill-available` out thats a neat idea i hadn't considered.

jamie-norman:
Looks like the IDS team is moving this into the current sprint. We continue to hear more noise and bugs written against this (in HCM and in Landmark), so I appreciate the focus on this issue.

claudenbach:
Seeing noise on the rendering of a lookup select list in a modal dialog were the last column doesn't stretch to fill the space.  A separate issue is when the modal dialog containing a lookup select has a search form.

jbrcna:
**QA Passed** with observation
v4.30.0-dev
Tested in all major browsers (Chrome, Firefox, Safari, Edge etc)
https://master-enterprise.demo.design.infor.com/components/datagrid/test-modal-datagrid-search-form.html
![image](https://user-images.githubusercontent.com/56722184/85718549-33dac300-b721-11ea-94e9-05bdd48ca76e.png)

**Observation**
In `Firefox`, the modal looks big
[...]

Issue #3879, opened 2020-05-05
Datagrid: modal dialog containing a datagrid issue with horizontal scrollbar


claudenbach:
Describe the bug
When opening a modal dialog containing a datagrid, if the browser's window is resized smaller a horizontal scrollbar appears on the modal body.  Want the horizontal scrollbar to appear on the datagrid itself.

To Reproduce
1. Go to 'test-modal-datagrid-two-columns.html'
2. Open the browser's debugger (in this example align debugger on the right)
3. Narrow the browser window until the horizontal scrollbar appear on the modal dialog

Expected behavior
Horizontal scrollbar appears on the datagrid itself.

Version
4.29.0-dev

Additional context**
if width:100% is added to the div element with a class of modal-body the scrollbar appears on the datagrid vs the modal-body.

claudenbach:
![image](https://user-images.githubusercontent.com/39804038/81086052-9ea81480-8ebd-11ea-8303-0755d8853cb5.png)

claudenbach:
![Git-3879](https://user-images.githubusercontent.com/39804038/81086552-445b8380-8ebe-11ea-93dd-19a79e1f99e5.gif)

EdwardCoyle:
@claudenbach you mentioned this:
> if width:100% is added to the div element with a class of modal-body the scrollbar appears on the datagrid vs the modal-body.

I believe that is our recommended way of handling scrolling issues within the Modal... Any thoughts @tmcconechy?

tmcconechy:
Yes it is. These data grid on modal issues are challenging because there is no widths set on the table. We have a way to do that with a default class now. Which might work but i think here the real ask was to be able to scroll in the grid itself not the modal body.

I know we discussed that somewhere @claudenbach ? Is this the issue around having a setting to scroll the grid not the modal? Or was that another issue im not finding?

claudenbach:
Yes, this is the issue to scroll the datagrid.  We discussed last week with Phillip in Teams.

claudenbach:
On that default class issue, when adding the class of 'datagrid-default-modal-width' using .cssClass on sohoModalDialogService, the class gets added to the div element with class of 'modal' but its not working.  Snippet of our code....

```
      this.dialog = this.sohoModalDialogService
      .modal(LMListLookupDialogComponent, this.modalPlaceholder)
      .buttons(buttons)
      .title(title)
      .autoFocus(false)
      .cssClass('datagrid-default-modal-width')
      .open()
[...]

PR #4009, opened 2020-06-04
3897/3879 - Datagrid: Fix table width and horizontal scrollbar inside of modal

ericangeles:
It seems that there's a missing class `datagrid-default-modal-width` that will fix the width of the table inside of the modal in #3897. There should be no changes on width being  set to `auto`.

By adding this class, the test page will set the `isInModal` to `true`, where it calculates and stretched the table to fill the datagrid container.

See this line https://github.com/infor-design/enterprise/blob/master/src/components/datagrid/datagrid.js#L4471

In #3879, I've set a width `auto` when the browser viewport is below `565` so horizontal scrollbar will not appear. (Fixes also when resizing it)

Closes
- https://github.com/infor-design/enterprise/issues/3897
- https://github.com/infor-design/enterprise/issues/3879

1.) #3897

- Navigate to http://localhost:4000/components/datagrid/test-modal-datagrid-search-form.html
- Click `Open Modal` button
- The table should filled the datagrid wrapper
- Toggle the filter button
- It should remain filled the datagrid wrapper
- Test this in all major browsers (Chrome, Firefox, Safari, Edge etc)

2.) #3879 
- Navigate to http://localhost:4000/components/datagrid/test-modal-datagrid-two-columns.html
- Click `Open Modal` button
- Resize the browser up to smaller (mobile) view
- There should be no horizontal scrollbar
- Try it also when the view is already on mobile
- Should work as well

3) Also test any pages on http://localhost:4000/components/datagrid with "modal" mentioned


tmcconechy:
Actually this did effect a few test pages in the tests. Maybe have a look. Im also wondering if we can get rid of `datagrid-default-modal-width` with this setting. There are three examples that use that that seem to be failing.

If you can remove that class from those examples and they work still then we can get rid of that as that was a much clunkier solution to this.

ericangeles:
Labeling this as WIP. I will also add the fix in #3879. Seems related to each other.

ericangeles:
Open to review now, @tmcconechy

src/components/datagrid/datagrid.js
```diff
@@ -4470,7 +4470,7 @@ Datagrid.prototype = {
 
       if (!this.elemWidth && this.element.closest('.datagrid-default-modal-width').length > 0) { // handle on invisible modal
         this.elemWidth = this.settings.paging ? 466 : 300; // Default a size for when on modals
-        this.element.css('width', this.elemWidth);
+        this.element.css('width', (!this.isInModal && window.innerWidth > 565) ? this.elemWidth : 'auto');
```
tmcconechy, line 4473: @ericangeles can you explain the logic on this? Im not sure i follow what its doing? Can you try with 3 columns? 4 ect? I was hoping we could handle this with `-webkit-fill-available` . Or is it actually using that.

The fix works in the steps but just wanted to verify

src/components/datagrid/datagrid.js
```diff
@@ -5963,6 +5963,10 @@ Datagrid.prototype = {
       });
     }
 
+    $(window).on('resize.modal', () => {
```
tmcconechy, line 5966: Do we need to add something to destroy? And is this needed if the data grid is not on  a modal? Maybe we can put isInModal around it?

docs/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 - `[Calendar]` Fixed a bug where the accordion sections would show a line on hover in high contrast mode. ([#2779](https://github.com/infor-design/enterprise/issues/2779))
 - `[Calendar]` Fixed a bug where the days would be out of alignment if the end and starts dates intersect. ([#1725](https://github.com/infor-design/enterprise/issues/1725))
 - `[Lookup]` Fixed an issue in the min width examples that showed up in Safari only. ([#3949](https://github.com/infor-design/enterprise/issues/3949))
+- `[Accordion]` Fixed an issue where the chevron icon is not poperly centered in Safari. ([#2161](https://github.com/infor-design/enterprise/issues/2161))
```
tmcconechy, line 18: Can you check the alhoabetic order here. Also this accordion entry is duplicated.

tmcconechy:
Any updates here @ericangeles ? Maybe you missed my comments?

tmcconechy:
I added another fix , removing the JS in favor of fully using webkit-fill available. With fallbacks to fixed with on IE and Edge

PR #4010, opened 2020-06-04
3771 - Allow Custom Builds with ES Module output

### LINKED ISSUE #3771: General: Add default export so IDS-enterprise can be bundled
Currently using IDS-enterprise require you to include the assets in `<script>` tags.
And it will be available globally afterwards. There's no way to include it via ES6 import/syntax. Therefore you cannot use bundler such as Webpack to include IDS-enterprise

Add default export, visible exports so components/functionalities from IDS-enterprise can be imported instead. 

In my modified `sohoxi.js` I just added a default export, and it was enough for Webpack to recognize and bundle IDS-enterprise.

EdwardCoyle:
This PR enables the creation of ES Module-style bundles with Rollup via the IDS custom builder.  Rather than being used through a script tag, this allows for the consumption of IDS as a direct import:

```js
import { * as Soho } from 'sohoxi.esm.js';
```
This PR also adds documentation about this feature to `docs/DEVELOPER.md`, and upgrades the Rollup library to its most current version.

Closes #3771

- Pull this branch
- Run `npm i` to upgrade Rollup.
- Try running a standard build with `npm run build`.  The build should complete as expected, and produce a standard bundle in `<project root>/dist/js/sohoxi.js`
- Try running an ES Module build with `npm run build -- --types=es`.  The build should complete and produce an ES Module at `<project root>/dist/js/sohoxi.esm.js`.  Note that this will not produce a corresponding source map on purpose (source maps would ideally be generated by the consuming application)
- Try running a build that produces both at the same time with `npm run build -- --types=iife,es`.  It should create both files. (the CI/release process will run this).


EdwardCoyle:
@guoliang would you be able to test this in your environment to see if the ES Module is formed as expected, and that all parts you need are accessible?

guoliang:
> @guoliang would you be able to test this in your environment to see if the ES Module is formed as expected, and that all parts you need are accessible?

tested briefly this build, and it works.
is it possible to add generation of the typescript type definitions or corresponding too?

EdwardCoyle:
@infor-design/eng-ids calling this one good to go

PR #4033, opened 2020-06-11
1725 - Add follow up calendar fix

tmcconechy:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
QA found a follow up issue changing and clearing events with #1725 so this is the fix.

**Related github/jira issue (required)**:
Fixes #1725 

**Steps necessary to review your pull request (required)**:
- go to http://localhost:4000/components/calendar/example-index.html
- click on any event (for example "More Meetings") and then extend the date one or two days.
- everything should better adjust now

PR #4040, opened 2020-06-15
fixes issue with accordion reflow

nbcp:
When dealing with a huge list of accordions plus a couple of expensive methods forcing reflow as mentioned here: https://gist.github.com/paulirish/5d52fb081b3570c81e3a, the whole application performance suffers. Anything you do is unbearably slow even just ticking a checkbox.

The proposed style changes here is not going to affect how the accordion would look. Doing the `display:none` when the panel is not expanded just helps the browser deal with the force reflows more easily.

I am unable to create a reproducible example in this repo since the problem only occurs when using the library in an angular project where those methods mentioned above are called during component modifications done in AfterViewChecks. 

What I can show is how it affects the performance in landmark:

accordion-reflow
This gif above shows the comparison with and without the proposed changes and how it affects the animations like expanding the menus and ticking the checkboxes.

### DIFF
--- a/src/components/accordion/_accordion.scss
+++ b/src/components/accordion/_accordion.scss
@@ -662,8 +662,8 @@
 
   // Visible accordion panes force a border to show up on the next accordion header for visual separation
   &.is-expanded {
-    padding: 0;
     display: block;
+    padding: 0;
 
     + .accordion-header,
     + .accordion-content {
--- a/src/components/accordion/_accordion.scss
+++ b/src/components/accordion/_accordion.scss
@@ -652,7 +652,7 @@
 .accordion-pane {
   @include transition(padding 300ms cubic-bezier(0.17, 0.04, 0.03, 0.94));
 
-  display: block;
+  display: none;
   overflow: hidden;
   padding: 0;
 
@@ -663,6 +663,7 @@
   // Visible accordion panes force a border to show up on the next accordion header for visual separation
   &.is-expanded {
     padding: 0;
+    display: block;
 
     + .accordion-header,
     + .accordion-content {
--- a/src/components/accordion/accordion.js
+++ b/src/components/accordion/accordion.js
@@ -722,7 +722,7 @@ Accordion.prototype = {
       return;
     }
 
-    return header.children('a').attr('aria-expanded') === 'true';
+    return header.children('a').attr('aria-expanded') === 'true' || header.hasClass('is-expanded');
   },
 
   /**
--- a/src/components/calendar/calendar.js
+++ b/src/components/calendar/calendar.js
@@ -485,7 +485,6 @@ Calendar.prototype = {
     }
 
     $('.accordion-pane.no-transition', this.element).removeClass('no-transition');
-    $(this.eventDetailsContainer).find('.is-expanded').removeClass('is-expanded');
   },
 
   /**
--- a/test/components/calendar/calendar.e2e-spec.js
+++ b/test/components/calendar/calendar.e2e-spec.js
@@ -373,7 +373,7 @@ describe('Calendar allow one pane tests', () => {
   it('Should only allow one pane open at a time', async () => {
     await utils.checkForErrors();
 
-    expect(await element.all(by.css('.calendar-event-details .accordion-pane.is-expanded')).count()).toEqual(0);
+    expect(await element.all(by.css('.calendar-event-details .accordion-pane.is-expanded')).count()).toEqual(1);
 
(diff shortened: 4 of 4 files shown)


tmcconechy:
Seems like this broke one test https://travis-ci.com/github/infor-design/enterprise/jobs/349083288#L785

Also you have some lint issues.

The Locale tests are fixed on another PR for your info https://github.com/infor-design/enterprise/pull/4033/files#diff-0b7cb3e3f9fe0b2c01599c6dc8107496L84

pwpatton:
@tmcconechy  We need this in 4.28.x and 7.1.x. I'm going to pull down the code and verify shortly.

pwpatton:
don't understand why a simple css change would break the build?! I'm thinking we can override this css in landmark for our imminent upcoming CU and fix this in 4.30. As well as fixing some of the pager stuff we found.

pwpatton:
**************************************************
1) Calendar specific month tests Should show events on click
  - Expected '' to be truthy.
2) Calendar allow one pane tests Should only allow one pane open at a time
  - Failed: Wait timed out after 3008ms

tmcconechy:
It looks like he set the accordion to closed by default. So now the calendar when using the accordion has it closed.

Something definately broke here. Seems like maybe the accordion will never default as open now.

As well as the linter errors which wont let me merge. I can take a look if you not looking at it already.

pwpatton:
interesting, the changes shouldn't cause a programmatic expanding of a menu, just that if it's not expanded then the accordion pane is set the display none. which is a great way to stop it from being processed in a reflow and makes sense because it's not actually displayed on the screen anyway.

tmcconechy:
It makes sense i think just needs a bit of tweaking i think.

pwpatton:
yea, and that gives us some things to verify on our side as well with the calendar.

tmcconechy:
Maybe calendar just needs the is-expanded class. Although that is sort of a breaking change in that you have to add that or it wont be expanded by default

pwpatton:
I don't think this is imminent as we can override in landmark for now. Ben can take a look when he's on next unless you think you have something obvious that can be done.

pwpatton:
strange, that is-expanded wouldn't be required to have it expanded.

tmcconechy:
Seems like it was already required http://localhost:4000/components/accordion/test-expand-on-init.html so that confuses me a little the other thing is the height code for animation. Maybe thats calculating wrong now that its invisible

pwpatton:
@EdwardCoyle we weren't seeing much in an EP example that @tmcconechy  made. Perhaps we didn't have the right amount of nested menus or something else that was causing us so much trouble. I do know that we seem to have a lot of dynamism and dom nodes in our apps that seems to cause a lot of reflow and tree layer updates. We can make this change locally to get by our current issue. I think we'll have to help illustrate the problem with a better example.

tmcconechy:
If you need it http://localhost:4000/components/applicationmenu/test-performance.html could be expanded on as its the example i tried

pwpatton:
@nbcp I think we need to make a sample that simulates the depth and breadth of some of these large landmark menus. I think perhaps the first place to do it is in ids-enterprise-ng since that's closes to what landmark uses. I think it will need simulate more of the dynamic nature of lmclient to get a real world idea of the "forced reflows" we are seeing.

tmcconechy:
Or, Maybe this isnt so far off:

1) Maybe we can make the animation routines https://github.com/infor-design/enterprise/blob/master/src/utils/animations.js add the height once its clicked so it still animates.
1) And fix the calendar issue by adding is-expanded class in the code/examples
1) And the lint issues

pwpatton:
It didn't have the animation problem when expanding. Just when collapsing. I think the animation on the collapse was lost because the animation started and finished after the `display: none` was applied (removing the `is-expanded` class would cause that). So perhaps a way to not set `display: none` until after the animation is complete would do (maybe an intermediate class like `is-collapsing`).

nbcp:
@tmcconechy @pwpatton I've added a new commit to address the calendar allow one pane test. The `allowOnePane:true` in accordion should now work as expected and only have is-expanded class only where it is needed. 

Unfortunately I was not yet successful in producing an example in ep and ep-ng that simulates the performance issue that worsens as the menu increases. I have also not yet found an alternative for the `display:none` that would not affect performance and produce a smooth animation when collapsing .

nbcp:
Also fixed the linter btw not sure why the build is failing on functional TimePicker API test

tmcconechy:
I'll reraise this on a branch so we can contribute.

PR #4052, opened 2020-06-17
NA - Add fixes to speed up accordion

tmcconechy:
When dealing with a huge list of accordions plus a couple of expensive methods forcing reflow as mentioned here: https://gist.github.com/paulirish/5d52fb081b3570c81e3a, the whole application performance suffers. Anything you do is unbearably slow even just ticking a checkbox.

The proposed style changes here is not going to affect how the accordion would look. Doing the `display:none` when the panel is not expanded just helps the browser deal with the force reflows more easily.

I am unable to create a reproducible example in this repo since the problem only occurs when using the library in an angular project where those methods mentioned above are called during component modifications done in AfterViewChecks. 

What I can show is how it affects the performance in landmark:
accordion-reflow
This gif above shows the comparison with and without the proposed changes and how it affects the animations like expanding the menus and ticking the checkboxes.

http://localhost:4000/components/applicationmenu
http://localhost:4000/components/calendar
and
http://localhost:4000/components/accordion

### DIFF
--- a/src/components/accordion/_accordion.scss
+++ b/src/components/accordion/_accordion.scss
@@ -652,7 +652,7 @@
 .accordion-pane {
   @include transition(padding 300ms cubic-bezier(0.17, 0.04, 0.03, 0.94));
 
-  display: block;
+  display: none;
   overflow: hidden;
   padding: 0;
 
@@ -662,6 +662,7 @@
 
   // Visible accordion panes force a border to show up on the next accordion header for visual separation
   &.is-expanded {
+    display: block;
     padding: 0;
 
     + .accordion-header,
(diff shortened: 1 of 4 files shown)


tmcconechy:
cc @nbcp I think the time tests should pass. The only issue i see is lack of animation when closing then i think this could be ok. Having a quick look as well.

EdwardCoyle:
http://localhost:4000/components/applicationmenu/test-performance.html is a really good test of this.  Seems to work pretty well!

nbcp:
thanks @tmcconechy, will try these changes out in landmark today

nbcp:
tested it, looks good

Issue #3666, opened 2020-03-26
Widget Count: Needs uplift fixes


tmcconechy:
**Describe the bug**
Update the widget counts to encorporate new designs https://www.figma.com/file/UTwbDbeB1Xv34Uz6js36L4/ids-widget-counts?node-id=0%3A1 

Modify this for uplift and soho theme.

**To Reproduce**
1. Go to 'https://master-enterprise.demo.design.infor.com/components/counts/example-widget-count.html
2. Compare to https://master-enterprise.demo.design.infor.com/components/counts/example-widget-count.html?theme=uplift&variant=light
3. Compare to https://www.figma.com/file/UTwbDbeB1Xv34Uz6js36L4/ids-widget-counts?node-id=0%3A1
4. Add new aspects of the design

**Expected behavior**
Looks like new designs

**Version**
- ids-enterprise: [e.g. v4.27]

**Screenshots**
![78367228-5b693600-75df-11ea-9e91-9502088d8e06](https://user-images.githubusercontent.com/814283/78562305-1f082680-77e7-11ea-88ed-b15bd1dcc857.jpg)

**Platform**
- Any

tmcconechy:
cc @kevinwhitedesign what do you think?

kevinwhitedesign:
Yeah these changes generally make sense to me. Any more background on this? Specifics?

This is what I'm seeing btw:
https://www.dropbox.com/s/yjggvyenuvu01ws/Screenshot%202020-03-31%2018.24.37.png?dl=0

tmcconechy:
Ok sounds good. Im not sure about the background. I mean in general this is just a "dashboard" of counts of certain types. The new design was something manilla provided to a team and we noticed a mismatch.

The only problem with it is that the text especially the yellow is not visible. Maybe we can come up with something with:
- the color
- an alert icon
- text that is readable with no contrast issues

kevinwhitedesign:
Oh gotcha, thanks for that background. I remember doing a few of these in the subtle-era. 

Maybe this would work. What did you mean by alert icon?
https://www.figma.com/file/UTwbDbeB1Xv34Uz6js36L4/ids-widget-counts?node-id=0%3A1

tmcconechy:
Yeah thats a nice treatment. Idea being just show the icon "somewhere" in case the use is color blind and cant determine the states by color alone. That could definitely work.

dhumalkishor:
[...]

PR #4054, opened 2020-06-18
3666 - Update the design of  widget counts in uplift

ericangeles:
Implementing the new design for widget count in uplift. I've also tweaked some positions in non uplift.

This is not limiting to the four colors that was in the design, we can use all the color palettes that we have, even adding other colors in the future. 

Closes https://github.com/infor-design/enterprise/issues/3666

- Navigate to http://localhost:4000/components/counts/example-widget-count.html?theme=uplift&variant=light&colors=0066D4
- Compare it to the old design in uplift https://master-enterprise.demo.design.infor.com/components/counts/example-widget-count.html?colors=0066D4
- Should show the new design for uplift
- Also check the non uplift, made some adjustments
- Test this to all variants


tmcconechy:
Calling this WIP as it needs a few more things.

ericangeles:
@tmcconechy you can take a look now. The CI will probably failed because of baseline image. But will fix it after.

tmcconechy:
Looking great. Maybe just two things now

1) On the 4th example: User example let's change the alert to another icon (like success)
2) On the 3rd example: Can we show icons on all 4 (error, orange alert, yellow alert , info?)
3) Can we add two more with six , two, three ect and see if they work ok?

tmcconechy:
Looks like you got it all covered now 😄  Maybe just fix the test and we call it done.
If you want to perhaps some of these can go on seperate test pages. And fix the test that way. Up to you either way.

ericangeles:
@tmcconechy this should be good to go now. ☕

PR #4076, opened 2020-06-23
853 - Add fixes for datepicker change event

### LINKED ISSUE #4075: Datepicker: Change event not trigering when selecting a single date in Date Range component
**Describe the bug**

**To Reproduce**
1. Go To Date Range component at http://localhost:4000/components/datepicker/example-range.html
2. click on the button in "Range - no initial value"
	select 6/3/2020 - 6/3/2020 (same day)'....'
3. We are getting 1 console logs for above action
First -  >     6/3/2020

**Expected behavior**
We should get 2 console logs for above action
First -  >     6/3/2020
Second  --->  6/3/2020 - 6/3/2020  

**Version**

- IDS Enterprise 4.28.0/ IDS Enterprise NG 7.1.0.

**Additional context**
We are getting two logs when we select separate dates in Date range component.

tmcconechy:

In a couple cases the change event didnt fire right.

Fixes infor-design/enterprisee-ng#853
Fixes #4075

- go to http://localhost:4000/components/datepicker/test-change-event.html
- open the picker and click today -> toast will fire
- blank out the field -> toast will fire
- open the picker and click today again -> toast will fire (this wasnt working before)
- go to http://localhost:4000/components/datepicker/example-range.html
- in the "Range - No Initial value" field open the picker
- click June 3
- click June 3rd again
- check the console and there should be two events one with one value and one with the same day range

### DIFF
--- a/app/views/components/datepicker/test-change-event.html
+++ b/app/views/components/datepicker/test-change-event.html
@@ -17,6 +17,7 @@
   $('#date-field-1, #date-field-2').on('change', function () {
       $('body').toast({
       title: 'Change Event Fired',
+      timeout: 20000,
       message: 'Value is now: '+ $(this).val() +''
     });
   });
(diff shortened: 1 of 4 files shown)


fitzorama:
@tmcconechy is it going to be an issue where in one spot lastValue is set to null and the other it is set to an empty string?

tmcconechy:
@fitzorama dont think it will matter from my tests but just in case i made them both null

Issue #3597, opened 2020-03-10
Hierarchy: menu selection lost after expanding hierarchy chart


bemariepedragosa:
**Describe the bug**
Menu selection is lost after expanding the hierarchy org chart

**To Reproduce**
1. Go to https://codepen.io/bemuuchan/pen/abOEQzg
2. Expand the hierarchy chart

**Expected behavior**
Menu is still selected

**Proposed solution:**
File > enterprise/src/components/hierarchy/hierarchy.js

selectLeaf function
   >> add .leaf class on selector
         $('.leaf.is-selected').removeClass('is-selected');

handleEvents function >> mouseup
   >> add .leaf class on selector
         $('.leaf.is-selected').removeClass('is-selected');

**Version**

- ids-enterprise: 4.25.2

**Platform**
- Chrome

tmcconechy:
cc @krishollenbeck ? what do you think?

bemariepedragosa:
Hi @tmcconechy and @krishollenbeck ,
Any updates on this issue? Thank you!

tmcconechy:
No one has had time for this yet on our end. But if you have an idea you could try a pull request. To clarify do you mean that the tab is loosing the selected (blue) indicator when you click the hierarchy to expand?

bemariepedragosa:
Yes @tmcconechy. You can test it here https://codepen.io/bemuuchan/pen/abOEQzg.
I have a proposed solution above and @krishollenbeck agreed to it based on his response above. 

**Proposed solution:**
**File > enterprise/src/components/hierarchy/hierarchy.js**

selectLeaf function

add .leaf class on selector
$('.leaf.is-selected').removeClass('is-selected');

**handleEvents function >> mouseup**

add .leaf class on selector
$('.leaf.is-selected').removeClass('is-selected');

Thank you!

tmcconechy:
cc @krishollenbeck what do you think about this one?

krishollenbeck:
@tmcconechy  The codepen example doesn't seem to work anymore. I could make on in ids-enterprise project though. If I recall the issue was that tab component and hierarchy both have the same class names for selected state. Seems like an easy fix. I can create a PR for it if you would like me to do that.

tmcconechy:
Sure thanks that would be great

jbrcna:
**QA Passed**
v4.30.0-beta0
https://4300-beta0-enterprise.demo.design.infor.com/components/hierarchy/test-vertical-tabs.html
[...]

PR #4100, opened 2020-06-26
Fixed hierarchy selection issue when nested inside of tabs.

krishollenbeck:
Hierarchy selection causes tab selection to be removed. This happens because both tabs control and hierarchy control have the same class name for selection state.

https://github.com/infor-design/enterprise/issues/3597

1. Pull PR
2. Launch ids-enterprise project
3. Navigate to  http://localhost:4000/components/hierarchy/test-vertical-tabs.html
4. Verify that selection of tabs is no longer lost when expanding a hierarchy leaf

hierarchy-inside-vertical-tabs-selection


ericangeles:
Looks good! Perhaps add a changelog?

PR #4108, opened 2020-07-01
3910 - Revert nested Modal behavior to stacking instead of hiding

### LINKED ISSUE #3910: Modals: Allow for some modals being opened above other modals
We have noticed that with the new modals when a modal is opened within another modal the first modal is temporary closed. This seems to work pretty well but gets a little weird when it comes to messages, such as confirm messages. 

Our request is to add the possibility to open some modals above others, especially message modals but if possible also other modals (for cases where it just looks weird or becomes to much hide/show ). Perhaps there could be an option parameter to keep below modal in background or something?

I've also considered if the approach to not hide the "base" modal should always be used for message modals, but I think it's better if there is an option for it.

I have also considered having the option parameter to keep in background on the "base modal", but in some cases both alternatives might need to be used within the same modal which I guess would not work with that approach.  

**Additional context**
-

And what application do you work on? Homepages

EdwardCoyle:
This PR changes the visual behavior of Nested Modals back to stacking, instead of hiding all the inactive modals by default.  The visual stacking behavior is slightly different than before 4.28.x, and shares a common overlay among the modals.  Inactive, open modals still appear on the page, but sit behind the overlay.

Additionally, a setting called `hideUnderneath` has also been added to Modals/Messages that allows the current behavior to be used, which will hide the modal window when another one is stacked on top.

Closes #3910

Pull this branch, build, run the app. Then:

_Test the standard behavior:_
- Open http://localhost:4000/components/modal/test-nested
- Click "Show Nested Modal" to open the modal
- Click "Show Another Modal" to open a second one on top
- Click "Show One More Modal" to open a third
- Notice the modals are now "stacked" like before, without multiple overlays.  The active modal is on top of the overlay, and the inactive ones are underneath.
- Close each modal.  When the top one disappears, the one beneath becomes active and is displayed above the overlay.

_Test the one-at-a-time behavior:_
- Open http://localhost:4000/components/modal/test-nested-no-underneath
- Click "Show Nested Modal" to open the modal
- Click "Show Another Modal" to open a second modal
- Click "Show One More Modal" to open a third modal
- Notice only one modal is visually present, and it sits above the overlay.
- Close each modal.  When the top one disappears, the one beneath becomes active and visible again.


EdwardCoyle:
@pwpatton @SofiK if you're able to, please test this in your environments

PR #4110, opened 2020-07-02
1060 - Fix search example focus states

### LINKED ISSUE #1060: Focus state needed on search examples
**Describe the bug**
The following pages show examples of search fields, that could use a focus state.

**To Reproduce**
http://localhost:4000/components/searchfield/example-header-large.html - should have a focus state
http://localhost:4000/components/toolbar/example-non-collapsible-searchfield.html - should have a focus state

**Expected behavior**
We should add the border change and dropshadow effect like other inputs.

jmacaluso711:
Fixed an issue where some of the searchfield examples did not have focus states.

closes #1060 

- pull and run branch
- go to http://localhost:4000/components/header/example-searchfield-large and focus the input. Make sure there is a visible focus state.
- got to http://localhost:4000/components/toolbar/example-non-collapsible-searchfield.html and focus the search field. Make sure there is a visible focus state.


tmcconechy:
Was reviewing this. Maybe it needs a bit more...

1) For the BIG search field. We in the past did a double box shadow. First pixel is white border around it and second is the blue border again. Something similar to http://localhost:4000/components/header/example-searchfield-full.html with the "go" button but with a bit of box shadow blur?
2) could also have focus state the same on http://localhost:4000/components/header/example-searchfield-full.html
3) For http://localhost:4000/components/toolbar/example-non-collapsible-searchfield.html im not as sure? Maybe a blue border all around it but but the box shadow on the bottom line?

tmcconechy:
Also the failing date test is fixed on https://github.com/infor-design/enterprise/pull/4118

jmacaluso711:
@tmcconechy what color were you thinking for the box shadow blur? I think part of the problem is you can't see the box shadow when the background is Azure.

tmcconechy:
Maybe white at .30 opacity? Might have to play around with it and see what looks good. Also notice the focus on the ... button on the header but the issue there is it has a azure background already.

Closest example i could find was that "Go" button on http://localhost:4000/components/header/example-searchfield-full.html

I think the box shadow part is going to be less visible/important to it

jmacaluso711:
@tmcconechy made some changes.. will need to update the baselines. Not sure if we want the lookup input to have the same focus state

tmcconechy:
@jmacaluso711 i think that works for these two cases. I like it 👍 
But i would say lets not change the entire lookup focus state or any other ones as they had been working ok. If you do want to make it consistent can try the box shadow around the whole thing on http://localhost:4000/components/toolbar/example-non-collapsible-searchfield.html I was thinking with the bottom border on that one just having it on the bottom would be a good idea. But i think either could work?

PR #4128, opened 2020-07-08
4085 - Misc Accessibility fixes

### LINKED ISSUE #4085: Accessibility: Miscellaneous Fixes
**Describe the bug**
A few quick hits on accessibility

- [x] Datagrid has a role="datagrid" in one place TJM: Could not see this in EP
- [x] Landmark is using quick edit mode (LM issue) TJM: Could not see this in EP
- [x] All links need hrefs to be focused TJM: Checked around and this seems ok on our end
- [x] Field Options -> lacking roles cannot tab to it but i think this is a LM only issue TJM: Need guidance or this is a LM issue
- [x] Datagrid -> need to be able to activate the tooltips in cells TJM: Added a fix
- [x] Buttons -> aria-pressed on toggles that don't toggle TJM: This seems ok in IDS toggle buttons http://localhost:4000/components/button/example-toggle-button.html
- [x] Context menu-> Focus goes to top of page on ESC? TJM: Fixed this in datagrids, it was jumping there but seems ok on http://localhost:4000/components/contextmenu/example-index.html 
- [x] Contextmenu/Popupmenu -> Roles are in wrong places TJM: Added a fix
- [x] Datagrid Checkboxes -> Has aria-selected in addition to aria-checked and it shouldn't TJM: Added a fix
- [x] Datagrid checkbox -> Cannot tab to header checkbox in row TJM: Added a fix
- [x] List Filter -> Dropdown Wrong Roles, not announced TJM: Need more info here...
- [x] Datepicker -> year and month selector -> focus on hitting apply goes back to field https://master-enterprise.demo.design.infor.com/components/datepicker/example-index.html TJM: Added a fix
- [x] App Menu -> Focus on hidden elements (Fixed previously)
- [x] Triggers -> button should get focus on tab, add a setting to turn button tabindex off (non default). Other libs like material https://stackblitz.com/angular/nknyovevygv?file=src%2Fapp%2Fdate-range-picker-overview-example.html do make it tabbable TJM: Skip for now and will discuss...
- [x] Filter Conditions -> Need aria label that you are selecting TJM: Ensure what this refers to
- [x] Editor -> Need aria labels TJM: Ensure what this refers to
- [x] App menu  -> Toolbar Icon buttons missing labels in examples? TJM: this was fixed, LM should update their examples
- [x] Vertical Tabs -> Tab panel role not being implemented correctly (tab panel content) TJM: Could not locate any issue, need more info.

**Version**
- ids-enterprise: [e.g. v4.29.0]

**Additional context**
LM...

tmcconechy:

Fixes several issues and looked into all issues on #4085 and was unable to reproduce all. Will circle back when we start working with those teams formally. Each fix should have a test now and lead to possible QA test cases. See steps for details.

Fixes #4085

**Issue 1**
- go to http://localhost:4000/components/applicationmenu/example-personalized-role-switcher.html
- enable chome vox
- tab down the app menu until you hit the buttons on the button
- labels added

**Issue 2**
- go to http://localhost:4000/components/datagrid/test-contextmenu.html
- right click any cell
- hit ESC -> focus should stay in the cell not shoot up

**Issue 3**
- go to http://localhost:4000/components/datagrid/test-ellipsis.html
- resize any text column so ... shows
- click the cell before it
- arrow to the cell with ... in it
- tooltip should show via the keyboard

**Issue 4**
- go to http://localhost:4000/components/datagrid/example-multiselect
- tab until you hit the multiselect check in the header -> this is now possible
- hit space

**Issue 5**
- go to http://localhost:4000/components/datepicker/example-index.html
- open the month/year pane
- tab to the apply button and hit enter -> focus now doesnt go to the field

**Issue 6**
- go to http://localhost:4000/components/tooltip/example-index.html
- tab to the button -> tooltip should show via the keyboard


EdwardCoyle:
This looks good but I had two questions:

- On issue #5, I saw that when using "Apply", focus is returned to the date field... that's supposed to be the behavior, correct?
- On issue #6, when the tooltip is displayed, should the contents of the tooltip be read by the screen reader?  I was hearing Vox read the contents of the button itself, but not the tooltip contents.

tmcconechy:
@EdwardCoyle 

On issue 5: On first apply with the month/year view open it returns back to the modal (not the field) that was the bug. But on the second apply with out month/year pane it goes back to the field when it closed.

On issue 6: I guess this could be optional so they could add that contents in the audible text or not if they want. Like for example if its showing an overflow value should be read as it ignores the ellipsis. The complaint was about keyboard users for this case.

PR #4131, opened 2020-07-09
4120 - Revert in string to starts with change

### LINKED ISSUE #4120: Regression: PR #3582 added unnecessary stylings to any classes which includes "btn"
**Describe the bug**

**This is a regression**.

In an application where there are lots of classes which contains `btn`, PR #3582 added unnecessary stylings to these classes.
The commit https://github.com/infor-design/enterprise/commit/70c44dfa80107984e88f9e31167bbf34713e6778#diff-3d5f7c545f1d12e63ab9f6c95f5e04d5 changed the selector from `[class=^'btn']` to `[class*='btn']` which causes the issue. 

Moreover, the PR #3582 **was not** listed in `CHANGELOG.md` .

**To Reproduce**
1. Have an example app, create an element in HTML as `<div class="foo-btn-bar"></div>`
2. Use Chrome Dev Tools to inspect this element
3. Observe that styling from `.btn` was added to this div

**Expected behavior**
Stylings from `.btn` shouldn't be added to the div above because the class name doesn't start with `btn`.

**Version**

- ids-enterprise: 4.27 -> latest

The issue doesn't happen to versions below 4.27

**Screenshots**
**N.A.**

**Platform**
 - Infor Application/Team Name: Infor ION
 - Device: all devices
 - OS Version: both Windows and MacOS
 - Browser Name: all browsers
 - Browser Version: **N.A.**

**Additional context**

The selector `[class*='btn']` can cause lots of troubles for any applications which have high level of customizations. This bug leads to the increasing amount of migration work for the team to upgrade to any `ids-enterprise` versions above 4.27.

**Workaround**
Change all class names to avoid them to contain `btn` or monkey patching `ids-enterprise` to change the selector.

tmcconechy:

As per #4120 the change we made for the btn class to be in string select vs starts with selector caused issues. After an analysis decided the best course was to reverse this change for now. In WC refactor we will make sure each button has its own classes.

Only thing i found wrong was it broke two buttons in tags. So i renamed those to start with btn.

Also for everyones info...Distinct button classes:

btn-menu
btn-tertiary
btn-primary
btn-secondary
btn-actions
btn
btn-icon
btn-close 
btn-default 
bin-toggle 
btn-modal-primary
btn-actions and btn-menu 
application-menu-trigger
bin-split-menu 
btn-moveto***
btn-expander 
btn-secondary-border 
btn-filter
btn-link 
btn-destructive 
btn-editor 
btn-month-year-pane 
btn-next 
btn-previous
btn-back 
btn-toggle-steps?
js-btn**

Number of ".btn" rules in all css -> 496

Fixes #4120

- Smoke test buttons http://localhost:4000/


ahnpnl:
hi, the issue is not completed resolved as this https://github.com/infor-design/enterprise/blob/c7c20a1b2a9ee1241dc36c5cb44d4016331890d8/src/components/toolbar-flex/_toolbar-flex.scss#L88 still causes issues with my custom stylings. Would you please check again ?

ahnpnl:
Hi, the issue is not resolved in **4.30.1** as https://github.com/infor-design/enterprise/blob/18baf6b19bc440979aa11409b45fc893a20ff751/src/components/toolbar-flex/_toolbar-flex.scss#L88 still causes the issue.

Master branch contains the correct fix.

I saw `CHANGELOG.md` mentioned that this PR is included in **4.30.1**.

ahnpnl:
I see the fix is in 4.31.0 so I guess it will be available soon as I see RC version is on going ?

tmcconechy:
@ahnpnl yes seems this fix was in 4.31.0. There is now an RC available as of yesterday. 7.5.0-rc.1 (NG) which contains 4.31.0-rc.0
You can give that a try and confirm.

ahnpnl:
yes, I confirmed 4.31.0-rc.0 in 7.5.0-rc.1(NG) fixed the problem.

PR #4162, opened 2020-07-20
4020 - Fix Dropdown backspace special character bug

### LINKED ISSUE #4020: Datagrid: Backspace issue with Dropdown lists on Grids
See a "口" display in the cell when pressing the "Backspace" key having selected an entry form a dropdown list on a grid

1. Open Example grid https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/example-editable.html
2. Press S key twice, then press Enter key, to select the Shipped option.
3. Press Backspace key twice
Result: "口" (square character) displayed in the cell

**Expected behavior**
No "口" (square character) displayed in the cell

**Version**
latest (4.28.2) *Screenshots**
image

**Platform**
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Chrome
 - 81.0.4044.138 (64 bit)

jmacaluso711:
Fixes an issue where pressing `backspace` in Windows 10 and `fn + delete` in Mac OS was inserting a special character in the input field. Using `String.fromCodePoint` to check for the special char in windows. Also checking for IE11 as this is not supported, but the bug was actually not occurring in IE11 so I think it's safe.

closes #4020 

- pull and run branch
- Go to http://localhost:4000/components/datagrid/example-editable.html click in an editable cell in the Action column.
- Select a new value, "Shipped" for example.
- In Windows press `backspace`  / In Mac OS press `fn + delete`
- The cell should now clear and not display a special character.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -27,6 +27,7 @@
 - `[Datepicker]` Fixed a bug where two dates may appear selected when moving forward/back in the picker dialog. ([#4018](https://github.com/infor-design/enterprise/issues/4018))
 - `[Datepicker]` Fixed a bug where an error may occur if using the gregorian calendar on ar-SA locale. ([#4130](https://github.com/infor-design/enterprise/issues/4130))
 - `[Dropdown]` Fixed an issue where "phraseStartsWith" does not filter the list after deleting a character. ([#4047](https://github.com/infor-design/enterprise/issues/4047))
+- `[Dropdown]` Fixed a bug when backspacing in windows or fn + delete in Mac OS would render a ascii character in the input field. ([#4020](https://github.com/infor-design/enterprise/issues/4020))
 - `[Editor]` Fixed a number of translation issues in the editor component. ([#4049](https://github.com/infor-design/enterprise/issues/4049))
 - `[Locale]` The Added placeholder for missing Thai `Locale` translation. ([#4041](https://github.com/infor-design/enterprise/issues/4041))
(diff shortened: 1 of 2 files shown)


tmcconechy:
Maybe getting my guys confused but:

- if running on a browser stack (windows running on a mac) and i do fn + delete ? Is that the same as back space key on a windows device?
- if so should it do the same thing as delete key?

ie:
- type s in the dropdown to open
- hit backspace (fn + delete ?) (nothing happens)

jmacaluso711:
Yeah that's what I assumed was the correct behavior. If you're on Mac and you hit delete nothing happens so fn+delete in windows should be the same.

tmcconechy:
Well on mac now delete (moves back in the list). So you have to type to filter then you can backspace what you typed.
fn+delete on mac i think isnt important there.

But what happens on PC (not able to test it on browser stack that accurately). Does backspace work or delete? Sounds from the issue they expect backspace to remove a character (not add the wierd char).

jmacaluso711:
Here's what I'm seeing when I test:

Mac OS: Press `delete`, nothing happens (this was the previous behavior)
Mac OS: Press `fn + delete`, the selection is deleted and no special char is inserted
Windows (browserstack): Press `backspace`, the selection is deleted and no special char is inserted
Windows (browserstack): `fn + delete`, nothing happens.

deep7102:
I tested with VM on my machine win10 and use onscreen keyboard (virtual keyboard comes with windows)

PR #4175, opened 2020-07-22
758 - Fix duplicate values when using paging

ericangeles:

Fixes duplicated values when paging is activated in `Lookup`. Changing the `value` from string to array so it can easily filter the duplicate values.

Closes https://github.com/infor-design/enterprise-ng/issues/758

- Navigate to http://localhost:4000/components/lookup/test-paging-with-count.html
- Open the lookup
- Go to page 2
- Select any `Product ID`
- Then click `Apply` button
- Open again the lookup, then click `Apply`
- It should not duplicate the values in the input field
- Test also on multiple values
- Open lookup, go to any pages except page 1, then select more than 1 item
- Click `Apply` button
- It should show the original values, and should not duplicate any of those.
- Count should correct also
 

ericangeles:
@tmcconechy I guess it fixed now. It seems some of ES6/ES7 features doesn't cover up in our app. Like for instance, `spread syntax` and `Array.from`. Not sure if we need to add some dependencies in polyfill to work these features.

EdwardCoyle:
@ericangeles Spread syntax should work, odd that it's not being transpiled by Babel here.  Also, we have a [utility method here](https://github.com/infor-design/enterprise/blob/master/src/utils/utils.js#L1146-L1156) which usually covers the cases for `Array.from()`

Issue #1924, opened 2019-04-08
RTL: Datagrid column resize direction is reversed


jmacaluso711:
**Describe the bug**
In RTL datagrid the columns are resizing in the opposite direction that you drag the resize-handle.

**To Reproduce**
1. Go to http://localhost:4000/components/datagrid/example-fixed-header.html?locale=he-IL#
2. Drag column resize-handle to the right and the column will resize to the left.

**Expected behavior**
The direction you drag and the column resizing should be going in the same direction.

tmcconechy:
This was missing a project so i added it, this was added into 4.18

janahintal:
Hi @tmcconechy, tested this ticket in the latest build. http://4180-rc3-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html?locale=he-IL# 

Issue still exist. Please see image below. Thanks
<img width="1419" alt="Screen Shot 2019-05-03 at 2 25 51 PM" src="https://user-images.githubusercontent.com/45757016/57122362-7cae1a80-6daf-11e9-91ba-a7a758e2ec68.png">

tmcconechy:
Did this issue maybe go in 4.19 I wasn’t sure but thought it was in 4.18 by the date @jmacaluso711

clepore:
@tmcconechy This went to master before we cut the `4.18.x` branch, so it should be in there. I'm going to go ahead and reopen this and move it to the DG milestone so you can address it with all the DG stuff for the `4.19` May sprint.

SofiK:
The wrong separator is also dragged when dragging.

![The_wrong_column_is_resized_ids](https://user-images.githubusercontent.com/8664975/69011373-07302300-096a-11ea-81cd-b9ace1b3e94c.gif)

tmcconechy:
Works better with recent fixes by needs more work for next sprint

janahintal:
QA Passed. This issue is now resolved. Verified in https://4260-beta0-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html?theme=uplift&variant=light&locale=he-IL across all platform. Thank you.

PR #4185, opened 2020-07-23
Misc - Datagrid Column Size Fixes and Features

### LINKED ISSUE #4016: Datagrid: Default column width renders too wide
**Describe the bug**
One pushback we are getting in landmark for 4.x is that there is too much horizontal scrolling for datagrids compared to the 3.5. 

**To Reproduce**
- See me for how URL/login and credentials.

**Expected behavior**
Want to columns to be greedy about reclaiming white space.

**Version**
7.1.1 of ng/ 2.28 of ep
**Screenshots**
In this case many columns seem to take more space than is needed. Some have their filter fields taking up more width than their cell.
image
In 3.5 this was much less of an issue because in landmark we set the size to fit on the columns and had a much smaller font.
image
A more aggressive reclaim of white space could be of at least marginal improvement. 
image

**Platform**

 - PC
 - Window10
 - Latest Chrome

**Additional context**

### LINKED ISSUE #4017: Datagrid: Add favor data setting over the column header in a datagrid when sizing column
One pushback we are getting in landmark for 4.x is that there is too much horizontal scrolling for datagrids compared to the 3.5.

A new setting to allow for cell data to be favored over the column headers when computing width.

A full auto fit all column where datagrid would never horizontally scroll; like in 3.5.

**Additional context**
With a few changes this could reclaim a lot of space and keep us from horizontally scrolling as much as we have to now. 
- change filter icon to be something much narrower perhaps just a down arrow
- have a minimum size for filter fields to they at least fit in the column header: sizes would be different based on the component. 
- Keep data always in view but allow the header text to have an ellipsis.

Notice in the before and after of the datagrid it's much small and it still shows all data. This is ofcource a stress test and when the browser is this wide it wouldn't need to have all the columns so narrow but I wanted to illustrate the difference that could be made up by doing the things listed above. 

image

### LINKED ISSUE #4032: Datagrid: Make the select column consistent size
When datagrid allows select/multi-select, the column size seems to be trying to factor into the column sizing. However, this column doesn't need to change widths as it will always contain the same size content. It can be very inconsistent when in a multi-list situation. 

Create a standard, fixed column size for the select column

-
**Additional context**
image

### LINKED ISSUE #4154: Datagrid: Spacing Inconsistencies
**Describe the bug**
I was looking at this https://master-enterprise.demo.design.infor.com/components/datagrid/example-index.html?theme=uplift&variant=light&colors=0066D4

Discovered that cell padding was 21px on large size. We should move this onto the 8pt grid and use 24px for this padding.
Screenshot 2020-07-15 12 07 40

I also found that medium was using 16px cell padding, and in places where text was right aligned - 21px padding.
Screenshot 2020-07-15 12 07 49

The small datagrid was using 11px, should we switch to 16px to be on the 8pt grid? (I guess 8px padding is another option here)
Screenshot 2020-07-15 15 15 28

The extra small datagrid was using 10px cell padding, but probably could be using 8px for the same reasons.
Screenshot 2020-07-15 16 32 24

My designs on the left:
https://www.figma.com/file/4CtBGz8dwvSa6JsBcdtVZQ/Data-Grid?node-id=29%3A2

tmcconechy:

This pr fixes 4 bugs and features all related to column sizing.
- Added a spaceColumn setting (no stretching)
- Added a columnSizing setting (data over header)
- Fixed bugs in the auto column width (mainly it wasn't taking the left and right padding into account)
- Fixed QA bug so that the left and right padding of cells and headers is in multiples of 4 (this will be the norm going foward). Large/Medium: 16 Small/Extra Small: 8

Impact: When reviewing try and smoke test a few example datagrid examples on http://localhost:4000/components/ particular with columns/column sizes, RTL themes and row height.

Note: Caused EVERY baseline to need an update 😢 

Fixes #4032 
Fixes #4154 
Fixes #4016 
Fixes #4017 

**Issue 4032**
- go to http://localhost:4000/components/datagrid/example-spacer-column.html
- notice the spacer column filling the right end
- hide/show columns in the .. menu
- the columns (particularly the selection column) will stay the same size
- try various column related tests

**Issue 4154**
- go to http://localhost:4000/components/datagrid/example-filter.html
- test themes and rowheights
- left and right padding has changed and is now Large/Medium: 16 Small/Extra Small: 8

**Issue 4016**
- go to http://localhost:4000/components/datagrid/example-books.html
- compare each to https://master-enterprise.demo.design.infor.com/components/datagrid/example-books.html
- notice some size reduction....
- try a few themes / row heights

**Issue 4016**
- go to http://localhost:4000/components/datagrid/example-columnsizing
- notice some size reduction and that the headers are cut off, its sizing to data alone
- go to http://localhost:4000/components/datagrid/example-books.html (this is the same but without that feature)
- try a few themes / row heights

**bonus fix**
- go to http://localhost:4000/components/datagrid/test-fixed-header-nested-grids.html
- expand the grid to reveal the nested grid
- scroll up and the headers should go under the other headers now


tmcconechy:
@EdwardCoyle which example is that video. That sounds like a separate issue which @jmacaluso711 and I?  fixed here: https://github.com/infor-design/enterprise/issues/1924

But wondering if i broke something as it Looks Good Here:
https://4260-beta0-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html?theme=uplift&variant=light&locale=he-IL

So would like to compare

EdwardCoyle:
I was able to make this occur on http://localhost:4000/components/datagrid/example-spacer-column.html?locale=ar-EG -- possibly due to the spacer column?

I tried on http://localhost:4000/components/datagrid/example-fixed-header.html?theme=uplift&variant=light&locale=he-IL and it doesn't seem to occur unless you:
- make sure the grid is scrolled all the way to the left
- grab the second or third to last column resize handles and slowly drag to the right

Seems like it's ok for a short distance and it accelerates once it passes a certain point, making it feel like you're dragging the wrong way?  Not sure.

IDS-Enterprise-_9_

jmacaluso711:
@EdwardCoyle @tmcconechy I'm seeing this behavior here as well: https://4260-beta0-enterprise.demo.design.infor.com/components/datagrid/example-fixed-header.html?theme=uplift&variant=light&locale=he-IL

If you keep dragging past the point where the columns on the right are at their min-width it continues to resize the column on the left. If you go really slowly you can see it, but if you continue dragging at the same pace it accelerates like in @EdwardCoyle's video.

tmcconechy:
OK i played around with this bit but it seems like a separate issue. Hey @deep7102 do you have any thoughts?

deep7102:
Yeah this looks funky, if you try to drag while holding shift key that's loos fine which isotherm option i see. I think this broken case while we dragging the column we are not adjusting whole table width which causing all the columns shifting. so might be if try to adjust table width accordingly as column wider/shorter might fix. not sure just thought might need to play around.

tmcconechy:
ok lets push that to later then - it wasnt really 100% related to this fix as evident it was there in 4.26 as well

deep7102:
Let's fix in separate issue.

tmcconechy:
@EdwardCoyle @jmacaluso711 skipping this for now made a new issue https://github.com/infor-design/enterprise/issues/4193

Issue #4152, opened 2020-07-14
Datepicker: Inconsistent CSS in read only mode


StaciGreen:
**Describe the bug**
The datepicker icon is greyed out in readonly mode when not showing time, but is black when showing time.

**To Reproduce**
1. Edit datepicker.demo.html
2. Add the following input at bottom (copy of standard date with timepicker, setting readonly="true"
```
<div class="four columns">
    <div class="field">
      <label for="datetime" class="label">Standard Date with Timepicker (Current Time) Readonly</label>
      <input soho-datepicker id="datetime" name="datetime" readonly="true" dateFormat="MM/dd/yyyy" timeFormat="HH:mm:ss" showTime="true" useCurrentTime="true" mode="standard" placeholder="MM/dd/yyyy HH:mm:ss" [(ngModel)]="model.datetime" (change)="onChange($event)" class="input-mm"/>
    </div>
  </div>
```
3. Run demo app, and goto Date Picker demo
4. See difference between read only inputs with & without show time

**Expected behavior**
The css for the date picker icon should be consistent read-only = true, regardless of showTime.

**Version**
- ids-enterprise-ng: 7.4.0

**Screenshots**
![image](https://user-images.githubusercontent.com/50175666/87468103-f8783980-c5e6-11ea-9a3e-d0f09484c969.png)

**Platform**
 - Device (if applicable) Windows PC
 - OS Version: Windows 10
 - Browser Name Chrome
 - Browser Version Version 83.0.4103.116 (Official Build) (64-bit)

**Additional context**

tmcconechy:
This issue is the same as https://github.com/infor-design/enterprise/issues/3176 however it seems timepicker was missed. Its still a different color on 

https://master-enterprise.demo.design.infor.com/components/timepicker/test-states.html
vs
https://master-enterprise.demo.design.infor.com/components/datepicker/example-index

As described...

jbrcna:
**QA Passed**
will retest once beta is deployed
v4.32.0-dev
https://master-enterprise.demo.design.infor.com/components/form/example-readonly-inputs.html
![image](https://user-images.githubusercontent.com/56722184/90127169-55f2d800-dd97-11ea-8eb4-ed320f2e2b29.png)

jbrcna:
Retest Done. Test Passed
v4.32.0-beta0
https://4320-beta0-enterprise.demo.design.infor.com/components/form/example-readonly-inputs.html

PR #4215, opened 2020-07-30
4152 - Fix inconsistencies readonly styling

ericangeles:

Fixes the inconsistencies of readonly styling on all input-related components.

Closes https://github.com/infor-design/enterprise/issues/4152

- Navigate to http://localhost:4000/components/form/example-readonly-inputs.html
- All input components should be consistent
- Test this to all themes and variants


EdwardCoyle:
To me it seems the icon colors for the Readonly state on both Dark variants, and the Vibrant Contrast variant, blend in a bit too much:

**Dark Variants:**
Screen Shot 2020-07-30 at 1 10 27 PM
Screen Shot 2020-07-30 at 1 11 23 PM

**Contrast Variants:**
Screen Shot 2020-07-30 at 1 10 57 PM
Screen Shot 2020-07-30 at 1 11 09 PM

... however, I'm not sure if I should mark Changes Requested here because it matches everything else (I tested against the Datepicker and the icon colors are the same there).  I think as long as it's easily changed, we should be OK if they want to change icon colors for Readonly/Disabled states overall in the future.

tmcconechy:
I suggest that we bump those up/down a notch while we are "in here". As they are rather faint to the point it looks like a mistake @EdwardCoyle

ericangeles:
Does these will work for all the input components?

tmcconechy:
@ericangeles yes for consistency all input components should be similar.

PR #4231, opened 2020-08-05
4101 - Add fifth column option with homepage

### LINKED ISSUE #4213: Datagrid Tree view fails to display rows sometimes when updating dynamically.
**Describe the bug**
Datagrid fails to show rows sometimes when updating dataset for tree via updateDataset() function. I created a new branch with demo illustrating the issue. This is a pretty close comparison as to how we update data in our application, using the updateDataset() function.

**To Reproduce**
1. Get code from example branch "**_datagrid-tree-updateDataset-issue_**"
2. Go to "http://localhost:4000/components/datagrid/test-tree-update-data.html"
3. Expand all rows.
4. Last row fails to show data. 

**Expected behavior**
All data should display

When expecting the DOM you can see that the rows are there. But they both have `is-hidden` class on them. After digging around for a while. It appears as though this code is the culprit: (**_datagrid.js_**)

```
rowHtml() {
...
if (parentNode && parentNode.node.expanded !== undefined && !parentNode.node.expanded) {
     isHidden = true; 
} else {
    isHidden = rowData._isFilteredOut;
}
...
```

**Potential fix:**
I think maybe this can be fixed by also checking if parentNode has children. For example.. (_this seems to fix the issue in my example_)
```
if (parentNode && parentNode.node.children && parentNode.node.expanded !== undefined && !parentNode.node.expanded) {
     isHidden = true; 
} else {
    isHidden = rowData._isFilteredOut;
}
```

`isHidden` appears to get set to true when a row with no children exists. Which then appears to cause the flag to continue to be true late on.

**Version**
4.32.0-dev

**Screenshots**

**_Example broken:_**
tree-not-expanding

**_Example with potential fix:_**
tree-expanding-with-fix

**Platform**
 - Windows 10
- Chrome

**Additional context**

deep7102:
Added support to fifth column option with Homepage.

Closes #4213

- Navigate to: http://localhost:4000/components/homepage/example-five-column.html
- Resize the browser window to different sizes
- On large view port should render five columns


tmcconechy:
@deep7102 is this ready or no? Saw its passing now

deep7102:
Ok, Updated now and this PR is now ready to review.

deep7102:
@tmcconechy Updated now, to use as `quintuple-width`

PR #4274, opened 2020-08-18
4145 - Fix Accordion nested header display

### LINKED ISSUE #4145: Accordion: Tree view display as black point when theme is Vibrant
**Describe the bug**
Tree view display as black point when theme is Vibrant.

**To Reproduce**
1. Change theme to Vibrant
2. Tree view display as black point

**Expected behavior**
Node name should list. The display is correct when theme is Subtle. See the screenshots.
Changing the font size of _.accordion-pane .accordion-header.list-item::before {...}_  from 1.6rem to 1.2rem in theme-uplift-*.css can solve this problem

**Version**

- ids-enterprise: [v4.26.2]

**Screenshots**
Display as black point when theme is Vibrant:
image

Display is correct when theme is Subtle:
image

**Platform**
 - OS Version: [Windows 10]
 - All Browser

**Additional context**
None.

jmacaluso711:
This PR fixes a bug in the vibrant theme where nested header text was not showing because the width was pushing it to the next line.

closes #4145 

- Pull and run branch
- Go to http://localhost:4000/components/accordion/test-six-levels-deep.html?theme=uplift&variant=light&colors=0066D4
- Navigate a few levels deep and you should now see the text in headers with bullet points.


EdwardCoyle:
@tmcconechy Not sure exactly what the use case is but AFAIK we've had it for some time and there are teams using it.

@jmacaluso711 I'd have to look through a bit to confirm this, but I want to say we've previously adjust all of those list item headers using that same `left-align-cascade-styles-header` mixin.  Instead of adding a new one it may just be a matter of tweaking the values (IIRC one of those four values specifically targets the list-item header)

jmacaluso711:
@tmcconechy @EdwardCoyle Seems as thought the combination of mine and @ericangeles changes have resolved the issues

PR #4379, opened 2020-09-16
1590 - Add support to disable column buttons with datagrid

### LINKED ISSUE #1590: DataGrid (Angular): Would Like to be able to disable buttons in the grid
yes, we're using buttons in datagrid for user operations such as delete & download. There are conditions where we dont want users to download, delete etc. There doesnt appear to be a way in the SOHO configuration to disable a button within a datagrid.

A clear and concise description of what the problem is. [e.g. {Component Name} has a problem when
An example of our datagrid buttons below. We'd like a way to be able to disable such buttons based on some business logic.
```
{ id: 'button-formatter', sortable: false, icon: 'download', align: 'center', width: "100",
        formatter: Soho.Formatters.Button, click: (e, args) => this.onDownloadCredential(args)
      }
```
A way to enable & disable individual buttons at runtime based on data held within each row. Ie could have a mix of disabled & enabled buttons in the rows.

we're using the Formatter to just now show the button. This is ok, but we prefer to disable such features rather than hide them away from the user

**Additional context**
none

deep7102:
Added support to disable column buttons with Datagrid.

Closes #1590

- Navigate to: http://localhost:4000/components/datagrid/test-icon-buttons.html
- See in 2nd and 3rd row for column `Disable Buttons`, `Disable Buttons (as link)` and `Product Desc (disabled)`
- The buttons and hyperlink should be disabled


tmcconechy (changes requested):
This looks good but

1) can we show the disable link in the test page?
2) can we name the setting `disabled` not `disableButton` ?

deep7102:
@tmcconechy updated now, to fix both cases as you found -thanks

tmcconechy (approved):

jmacaluso711 (approved):

PR #4408, opened 2020-09-22
4079 - Enable independent editable sections in Date Masks

### LINKED ISSUE #4079: Datepicker/Mask: Editing dates results in odd modifications
**Describe the bug**
Editing a date causes a shift of all numbers into the left most value ignoring the date separator and makes editing dates confusing to the end user.

**To Reproduce**
1. Go to 'https://master-enterprise.demo.design.infor.com/components/mask/test-date-with-validation.html'
2. Enter 1/1/2020 into the Date Input field and tab out
3. Return to the Date Input field and place the insertion point to the right of the month; intention is to change 1/1/2020 to 1/2/2020
4. Use the backspace key to remove the month and then enter a 2
5. Value becomes 12/20/20

**Expected behavior**
Expected the date mask to honor the position of the date separators and expected the value to be 1/2/2020.

**Version**

- ids-enterprise: 4.28.1

**Screenshots**
datepicker-editting

**Additional context**
Landmark JIRA ticket for this: https://jira.lawson.com/browse/LMCLIENT-25021

EdwardCoyle:
This PR makes a change to the way masks are processed to allow for the sections of the date to be altered independently of each other.  Usage of the mask `pattern` array internally has been modified to include optional meta data about a mask, including caret trap information.  Also, Date masks specifically have been modified to pass new meta-data about the location and types of separators inside the date format, which is used to understand the groups.  The result is that when modifying portions of a full date field with a mask, the sections on opposite sides of formatting characters like `/` and `:` from the text caret should stay in-tact, instead of being jumbled.

Additionally tests have been added for the new use cases reported for this enhancement.

Closes #4079 

- Pull this branch, build, run the demoapp
- Open http://localhost:4000/components/mask/test-date-with-validation.html
- In the first text field, type `1/1/2020`
- Use the keyboard to navigate your cursor to after the second `1`, and press backspace.  The text field should read `1//2020` and the cursor should be between the slashes.
- Press `2`.  The text field should read `1/2/2020`.
- Press backspace and press `02`.  The field should read `1/02/2020`.
- Change the `1` in the beginning to read `01`.  The field should read `01/02/2020`.

Also test other types of masks, since some of the changes here were related to general mask processing. look at the ones on:
- http://localhost:4000/components/mask/test-number-mask-gauntlet.html
- http://localhost:4000/components/mask/example-index.html (specifically date/time mask)


jmacaluso711 (approved):

tmcconechy (changes requested):
This looks good. I see one thing worth discussing.

- on http://localhost:4000/components/mask/test-date-with-validation.html if you just type "12" then you will see the cursor is before the / i think it should be after? Also type 12/12 and again the cursor is before. It wasnt this way before https://master-enterprise.demo.design.infor.com/components/mask/test-date-with-validation.html

EdwardCoyle:
@tmcconechy That's actually part of the change needed to make the group edits work the way it was requested.  The previous version of the date mask function didn't use the caret trap feature, which (for example) is what allows the cursor to be on either side of the decimal in the number mask -- therefore allowing edits on either side of the decimal.  You can do the same thing between date groups now.  

The usability here is the same though, because if you type numbers (or any character that matches the next date group, like AM/PM work for times) the mask still allows the input and continues.  The difference now is that typing the separator also works to advance the cursor in the mask (a bit more usable than before).  The characters before/after the separators are respected (except in cases where something like AM/PM slip into a "numbers-only" group, which removes them).

tmcconechy:
ok, its something i noticed but if thats part of the fix we can go with it. Then see if anyone else notices, i fear we might have to revisit as it looks like it's in the wrong position and about to type in the wrong place

EdwardCoyle:
For the record, not sure there's a way to get this functionality without the change in cursor placement.

tmcconechy (approved):

Issue #4312, opened 2020-08-27
Lookup: Additional issues with count update when filtering - when no value is found


dhumalkishor:
Async data Lookup control count doesn't update for filtering. Its just showing total count we pass in.

**Describe the bug**
Async data Lookup control is still having issue at the time of search, record count results not get refresh.

**To Reproduce**
In EP (Possibly).

1. Go to https://master-enterprise.demo.design.infor.com/components/lookup/test-paging-server-side-keyword-search.html 
1. open the lookup
1. in the filter type "xxxxxxxx" and hit enter
1. count is not zero

Also test in NG
1. Check lookup-validation.demo example, In that “Products (Async Results)” lookup (please refer the screen ).
1. Open the lookup, It will show the result as '7'. (please see the screenshot)
1. If we search for any value, which is not available such as "xxxx" then it is not updating count

Expected behavior
The count should update for filtering when the value is not found as it does when found.

**Version**

ids-enterprise: [e.g. v4.9.0 or v4.10.0] 4.31.1

**Screenshot**

![65154841-d99e2a80-da49-11e9-857a-e44776c3e7e5](https://user-images.githubusercontent.com/42198413/91451307-ebfc2780-e89a-11ea-9ae6-90c07a83663d.png)
![L1](https://user-images.githubusercontent.com/42198413/91451393-fe766100-e89a-11ea-8271-358a8c7b6440.jpg)

dhumalkishor:
Record count not get updated when there is no records found in search.
This issue is observed in 4.31.1 library. You can reproduce this in below example of enterprise-ng.
http://localhost:4200/ids-enterprise-ng-demo/lookup

tmcconechy:
I'm going to make this issue with easier steps so it has a chance of finally being resolved. So editing the issue now.

jbrcna:
**QA Passed**
v4.33.0-dev.
record count can now be updated. Also, If the record count is 1 then the word is singular (i.e. `result`) otherwise it is displayed as `results`.
Will retest this once beta is deployed
https://master-enterprise.demo.design.infor.com/components/lookup/test-paging-server-side-keyword-search.html
![image](https://user-images.githubusercontent.com/56722184/94427640-00954d80-01c2-11eb-9edb-48e709f2271b.png)

jbrcna:
Test passed. verified in v4.33.0-beta1
https://4330-beta1-enterprise.demo.design.infor.com/components/lookup/test-paging-server-side-keyword-search.html

PR #4410, opened 2020-09-23
4312 - Fix async data lookup count doesn't update correctly

ericangeles:

Fixes the count in lookup upon searching. Replacing the `this.lastCount` to the value of `dataset` which seems more accurate. I also change the condition in `countText` to display the text correctly.

Closes https://github.com/infor-design/enterprise/issues/4312

- Go to http://localhost:4000/components/lookup/test-paging-server-side-keyword-search.html
- Open the lookup
- Search `x` then hit enter
- It should display `(0 results)`
- Search `compressor 0` then hit enter
- It should display `(1 result)`
- More than 1 data should display `results`
- Try any combination to check if the count is displaying the numbers correctly


jmacaluso711 (approved):

tmcconechy (changes requested):
Looks good except for one thing. For "0" it should say "0 Results" not "0 Result". One is singular; all else is plural. 

https://english.stackexchange.com/questions/248173/0-result-vs-0-results

tmcconechy (approved):

tmcconechy (dismissed):

jmacaluso711 (approved):

Issue #4403, opened 2020-09-22
Timepicker: Missing Audible Labels


tmcconechy:
**Describe the bug**
Recent VPAT scan indicates our timepicker has some flaws. Its missing labels (maybe a dropdown issue) and has a wrong role.

- role="combobox" is expecting children so is not correct
- there is a label on the select but it isn't read since its attached to the hidden select

```
<div class="dropdown" data-options="{'noSearch': 'true'}" role="button" aria-haspopup="listbox" tabindex="0"><span>1</span></div>
```

**To Reproduce**
1. Go to https://master-enterprise.demo.design.infor.com/components/timepicker/example-index.html
2. Click on the trigger to open
3. Open Axe/Inspect DOM

**Expected behavior**
Fix the role combobox - add a hidden label feature so it says "1 minutes", "30 hours ect"

**Version**
- ids-enterprise: [e.g. v4.32.0]

**Platform**
 - Chrome / Any

**Additional context**
https://jira.lawson.com/browse/LMCLIENT-31955

ssaniel:
Hi @tmcconechy , This is one of the identified High Priority VPAT items. thank you!

jbrcna:
QA Failed
Failing this issue for now. please see notes below. 
https://master-enterprise.demo.design.infor.com/components/timepicker/example-index.html?layout=nofrills
![image](https://user-images.githubusercontent.com/56722184/109445792-59238c00-7a7b-11eb-9807-efb334ade469.png)

https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html
- Tab / Shift+Tab between the Dropdown inputs (which should now also display visible labels for each field). When focused, each should read their label and currently-selected contents (fx: "Hours, 1" or "Minutes, 00"), followed by their role and state (fx "Button, Collapsed/Expanded")
- it doesn't mention the "Hours", "Minutes", "Seconds", "Period" before the number

**Date Time Picker with Custom Pattern**
- `Hours` is incorrect (default value: it says 1 instead of the value displayed)

**Date Time Picker in Locale (Time to Midnight)**
- `Hours` is incorrect (default value: it says 1 instead of the value displayed)

**Date Time Picker in Locale (Time to Current)**
- `Hours` is incorrect (default value: it says 1 instead of the value displayed)

**Date Time Timestamp**
- `Hours` is incorrect (default value: it says 1 instead of the value displayed)
- `Minutes` is incorrect (default value: it says 00 instead of the value displayed)
[...]

PR #4487, opened 2020-10-13
4403 - Fix accessibility issue and missing audible labels

ericangeles:

Fixes accessibility issue, correct the locale of time period and add missing audible labels.

Closes https://github.com/infor-design/enterprise/issues/4403

- Go to http://localhost:4000/components/timepicker/example-index.html
- Click the timepicker icon to trigger the popup
- Open Axe in developers tool
- Click the `Analyze` button to check if there's an issue
- There should be no issue
- Check the audible labels of minutes, hours, seconds, and time period. Should be there.


tmcconechy (changes requested):
A few improvements needed here.

a) you dont need to translate all the files. the translation team will do that, i usually just translate es-ES for testing
b) Some tests appear to need adjusting
c) the issue that it " its attached to the hidden select" is still the case. the new label you added isnt read by the screen reader since both are disabled. need to put this label in the dropdown div for example

```
<div class="dropdown" role="button" aria-haspopup="listbox" tabindex="0">
<span><span class="audible">Hours </span>1</span></div>
```

And i I think that they labels can stay hours, minutes second instead of Time Period as that is more clear. I tested this in chrome vox BTW.

ericangeles:
All good, @tmcconechy. One thing tho, I don't see the last issue that you mentioned, I see it as `role="option"`.

tmcconechy (commented):
Im just rereading this...

`The Hour, Minutes, and Meridian inputs in time picker controls do not have valid labels, but instead rely on the current value of the input and its formatting to convey its purpose.
Although a <label> element is correctly associated with the hidden <select> element associated with these controls, the label is not announced because users are not interacting with the hidden <select>, but instead the visible custom controls.
[Code Reference]
<div class="time-parts">
    <label for="timepicker-hours-timepicker-11-id" class="audible">Hours</label>
    <select id="timepicker-hours-timepicker-11-id" data-options="{'noSearch': 'true'}" class="hours dropdown">
        ...
    </select>
    <div class="dropdown-wrapper">
        <div class="dropdown" data-options="{'noSearch': 'true'}" role="button" aria-haspopup="listbox" tabindex="0">
            <span>7</span>
        </div>
        <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
            <use xlink:href="#icon-dropdown"></use>
        </svg>
    </div>
    <span class="label colons">:</span>
    ...
</div>
[Recommendation]
Consider whether it is possible to use native html <select> elements or text inputs instead of custom controls here. If this is not possible, ensure that the visible custom controls have valid labels.`

So actually the problem is more complicated i think. What im seeing with the screen reader seems incorrect.

a) its not a role="textbox" so maybe we didnt need to change that (even if it is correct then its already an input so i think we need to check the dropdown).

It acts like a button but its really a combobox/dropdown. And when i open it and arrow down i dont get read the values.

Think it should look and act like "Example 2: List Autocomplete with Automatic Selection" on 
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html when using the screen reader on the dropdowns in the picker

ericangeles:
Was also thinking maybe we wrap the input to a div element that has a combobox role? Since the combobox role requires child element. What do you think?

ericangeles:
So something like this,
```
<div role="combobox" aria-expanded="true" aria-haspopup="listbox">
    <input id="timepicker-main" class="timepicker" type="text" data-validate="time" aria-autocomplete="list" data-validation-events="{&quot;time&quot;:&quot;blur change enter&quot;}" aria-activedescendant>
</div>
```

tmcconechy:
Yes i think it should work like https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html but i think we may have had it this way. And for some reason we changed it @EdwardCoyle for details? Keep in mind there a lot of functionality around the dropdown and multiselect to test for this if we do it.

EdwardCoyle:
@tmcconechy I don't remember it being this way before, but maybe it was earlier on.  As far as the new functionality goes in the Dropdown/Multiselect, we should try to keep the current behavior as default, and maybe make automatic selection "opt-in" (boolean setting in Dropdown API).

tmcconechy:
@EdwardCoyle i think your talking about a different issue. This one is to do with only accessibility and the reason we dont have an input element for the select and use a DIV

EdwardCoyle:
I see, I misunderstood.  We did have the in-page element as an `<input>` before, but we moved to the `<div>` because:
- I think originally there was an accessibility problem the other way around - there was an input in the page that wasn't actually the one used, because the one that the user actually interacts is inside the dropdown list.  IIRC this was causing some other wonky things in the screen reader, but not 100% sure.
- When we adopted the tag list in the Multiselect, the display of these wouldn't be possible inside an `<input>` element.

tmcconechy:
ok right it was the tag list stuff. Ok @ericangeles see if you can make it like https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html with aria changes but the div should stay as a div that acts like an input.

tmcconechy:
@ericangeles i put the label in WIP as i wasnt sure if this was done. But if you think its ready to review again, please remove the label.

ericangeles:
> @ericangeles i put the label in WIP as i wasnt sure if this was done. But if you think its ready to review again, please remove the label.

Yes, this is still WIP. Am still getting error on axe. Still finding a way to fix it.

tmcconechy:
Can you reraise a PR when done? The change logs are all out of sorting so keep getting conflicts

Issue #4452, opened 2020-10-05
Listview: Readonly / Non-selectable listview should not have hover / selected styling


volante007:
**Describe the bug**
A listview with `selectable: false` still gets focus / hover styles. 

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/listview/example-index?theme=uplift&variant=light&colors=0066D4
2. Hover the items
3. Select an item

![list](https://user-images.githubusercontent.com/1101770/95077301-8f9aec00-0713-11eb-99f8-73fa1f4c3810.gif)

**Expected behavior**
I think a readonly / non-selectable list should just be a list of items, no special hover / selected styling. The items are just info.
There are of course scenarios where a non-selectable listview _should_ have some kind of hover style, e.g. list of links / clickable (but not selectable) items, but I think a "plain" list should be the default, especially since these styles are difficult to override by the consumer. Alternatively an option that makes the list "plain", to avoid having to override focus / hover / active etc for info-only lists.

**Version**
- ids-enterprise: v4.32.0

janahintal:
This issue is now resolved. Thank you. 
![image](https://user-images.githubusercontent.com/45757016/96577045-a5103880-1305-11eb-9572-15d34d95f784.png)

PR #4494, opened 2020-10-14
4452 - Fix readonly and non-selectable listview should not have hover state

ericangeles:

Readonly and non-selectable listview should not have hover state.

Closes https://github.com/infor-design/enterprise/issues/4452

- Go to http://localhost:4000/components/listview/example-index?theme=uplift&variant=light&colors=0066D4
- Hover the items
- It should not have hover styling
- Test this also on all themes and variants combinations

### DIFF
--- a/app/views/components/popupmenu/test-return-focus.html
+++ b/app/views/components/popupmenu/test-return-focus.html
@@ -1,0 +1,23 @@
+<div class="row">
+  <div class="twelve columns">
+
+    <div class="field">
+      <button id="popupmenu-trigger" class="btn-menu" data-init="false" >
+        <span>Normal Menu</span>
+        <svg role="presentation" aria-hidden="true" focusable="false" class="icon icon-dropdown">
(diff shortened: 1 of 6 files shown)


tmcconechy (changes requested):
There is a trick to this to. We still need the focus state and to be focusable (could make that an option tho).

But in the example https://design.infor.com/code/ids-enterprise/latest/demo/components/listview/example-index?theme=uplift&variant=light&colors=0066D4 its correct it gets a blue border on click and its correct that you can use keys to move up and down. The only thing NOT correct is the hover effect since that is only needed if selected.

So we still need focus on http://localhost:4000/components/listview/example-index?theme=uplift&variant=light&colors=0066D4

tmcconechy (dismissed):
Cool 👍 
This works as i am expecting now

tmcconechy (approved):

deep7102 (approved):

Issue #4498, opened 2020-10-14
Testing: Add automation ID functionality


tmcconechy:
**Is your feature request related to a problem or use case? Please describe.**

This feature is to add the automation ID functionality (attributes) setting to all components. The issue is broke into sections for each person. You can do one or several PR's against this ticket and feel free to swap the components out if the breakdown makes more sense for one the other is doing. Rough work is:

- add attributes setting js doc
- add utils.addAttributes
- in the build or place where the modal is generated add an addAttributes for the component root
- and one for each thing that is clickable that is "appened" for example in the toast for `btn-close`
- add readme.md section for testability with same text and noting what's appended for the component
- an e2e test

Can look at modal, message, pager, toast, colorpicker now as completed. Example for color picker https://github.com/infor-design/enterprise/pull/4492

**Component Assignments Checklist**

**Eric**
- [x] Busyindicator
- [x] Circlepager
- [x] Emptymessage
- [x] Error Page
- [x] Expandablearea
- [x] Notification
- [x] Rating
- [x] Slider
- [x] Spinbox
- [x] Splitter
- [x] Switch
- [x] Textarea
- [x] Timeline
- [x] Timepicker
- [x] Signin

**John**
- [x] Autocomplete
- [x] Contextmenu
- [x] Field Filter
- [x] Field Options
- [x] Fieldset
- [x] Fileupload
- [x] Fileupload Advanced
- [x] Dropdown
- [x] Multiselect
- [x] Popdown
- [x] Popover (#4551)
- [x] Popupmenu (#4544)
- [x] Searchfield
- [x] Toolbarsearchfield

**Ed**
- [x] Accordion (ids just go on the elements?) (#4514)
- [x] Applicationmenu (ids just go on the elements?) (#4514)
- [x] Breadcrumb (#4514)
- [x] Button (#4544)
- [x] Input (check cases it may not be needed?) (#4529)
- [x] Listview (#4529)
- [x] Lookup
- [x] Tabs (#4518)
- [x] Tabs Header (#4518)
- [x] Tabs Module (#4518)
- [x] Tabs Multi (#4518)
- [x] Tabs Vertical (#4518)
- [x] Toolbar Flex (#4544)
- [x] Toolbar (#4558)
- [x] Tooltip (#4551)

**Tim** 
- [x] About
- [x] Contextualactionpanel
- [x] Donut
- [x] Pie
- [x] Calendar
- [x] Calendar Toolbar
- [x] Week View
[...]

PR #4510, opened 2020-10-19
4498 - Automation IDs for autocomplete, field filter, dropdown and popupmenu

jmacaluso711:
Adds automation is attributes for autocomplete, field filter, dropdown and popupmenu

related to #4498 

- pull and run branch
- go to http://localhost:4000/components/autocomplete/example-index.html and make sure the element gets automation id from the settings
- go to http://localhost:4000/components/field-filter/example-index.html and make sure the field wrapper, label and select get an automation id from the settings
- go to http://localhost:4000/components/dropdown/example-index.html open the dropdown and make sure the dropdown list, label, input, trigger and listbox get an automation id from the settings
- go to http://localhost:4000/components/popupmenu/example-index.html open the dropdown and make sure the popupmenu and options get an automation id from the settings

### DIFF
--- a/src/components/autocomplete/autocomplete.js
+++ b/src/components/autocomplete/autocomplete.js
@@ -253,6 +253,15 @@ Autocomplete.prototype = {
       this.list = $('<ul id="autocomplete-list" aria-expanded="true"></ul>').appendTo('body');
     }
 
+    // Add test automation ids
+    utils.addAttributes(this.list, this, this.settings.attributes, 'list');
+    setTimeout(() => {
+      const options = this.list.find('li a');
+      [...options].forEach((opt, i) => {
+        utils.addAttributes($(opt), this, this.settings.attributes, `list-option${i}`);
+      });
+    });
+
     this.list[0].style.height = 'auto';
     this.list[0].style.width = `${this.element.outerWidth()}px`;
     this.list.addClass('autocomplete');
--- a/src/components/autocomplete/autocomplete.js
+++ b/src/components/autocomplete/autocomplete.js
@@ -222,8 +222,8 @@ Autocomplete.prototype = {
       role: 'combobox',
       autocomplete: 'off'
     });
-    console.log(this.element);
-    utils.addAttributes(this.element, this, this.settings.attributes, 'autocomplete');
+    // Add test automation ids
+    utils.addAttributes(this.element, this, this.settings.attributes);
   },
 
   isLoading() {
--- a/src/components/autocomplete/autocomplete.js
+++ b/src/components/autocomplete/autocomplete.js
@@ -222,6 +222,8 @@ Autocomplete.prototype = {
       role: 'combobox',
       autocomplete: 'off'
     });
+    console.log(this.element);
+    utils.addAttributes(this.element, this, this.settings.attributes, 'autocomplete');
   },
 
   isLoading() {
--- a/src/components/autocomplete/readme.md
+++ b/src/components/autocomplete/readme.md
@@ -94,8 +94,4 @@ Setting the id/automation id with a string value:
 
 If setting the id/automation id with a function, the id will be a running total of open autocomplete.
 
-For the autocomplete, `attributes` can be set either on the root settings or in each button element. For the button elements you can also use the `id` as before or the new attributes setting to set an id or automation id on the button.
-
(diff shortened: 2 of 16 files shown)


src/components/autocomplete/autocomplete.js
```diff
@@ -222,6 +222,8 @@ Autocomplete.prototype = {
       role: 'combobox',
       autocomplete: 'off'
     });
+    // Add test automation ids
+    utils.addAttributes(this.element, this, this.settings.attributes);
```
tmcconechy, line 226: We probably need to add id's to the menu items (like i did in pager). But im not sure if that would be part of dropdown/ menu or autocomplete

src/components/autocomplete/readme.md
```diff
@@ -74,4 +74,28 @@ The data returned should be an array with a specific data structure containing `
...
+
+```js
+  attributes: [{ name: 'id', value: 'my-unique-id' }, { name: 'data-automation-id', value: 'my-unique-id' }]
+```
+
+If setting the id/automation id with a function, the id will be a running total of open autocomplete.
+
+For the autocomplete, `attributes` can be set either on the root settings or in each button element. For the button elements you can also use the `id` as before or the new attributes setting to set an id or automation id on the button.
+
+If you set the attributes on the root message, you will get an ID added to the root of the message dialog. Also the message area will get an id with `-message` appended after the id given. And the h1 area will get an id with `-title` appended after the id given. And finally the close button (if used) with get `btn-close` appended.
```
tmcconechy, line 99: This comment is related to message still.

src/components/dropdown/dropdown.js
```diff
@@ -1861,6 +1861,13 @@ Dropdown.prototype = {
       .attr('aria-expanded', 'true')
       .addClass('is-open');
 
+    // Add test automation ids
+    utils.addAttributes(this.list, this, this.settings.attributes);
+    utils.addAttributes(this.list.find('label'), this, this.settings.attributes, 'label');
+    utils.addAttributes(this.list.find('input'), this, this.settings.attributes, 'input');
+    utils.addAttributes(this.list.find('.trigger'), this, this.settings.attributes, 'trigger');
+    utils.addAttributes(this.list.find('ul'), this, this.settings.attributes, 'listbox');
+
```
tmcconechy, line 1870: Probably need to do the options. I think this means appending something to the a tags (even if id's work directly there the automation-id wont)

src/components/dropdown/readme.md
```diff
@@ -77,7 +77,31 @@ The dropdown list has been coded to act similar to the [aria best practices coll
...
+
+```js
+  attributes: [{ name: 'id', value: 'my-unique-id' }, { name: 'data-automation-id', value: 'my-unique-id' }]
+```
+
+If setting the id/automation id with a function, the id will be a running total of open dropdown.
+
+For the dropdown, `attributes` can be set either on the root settings or in each button element. For the button elements you can also use the `id` as before or the new attributes setting to set an id or automation id on the button.
+
+If you set the attributes on the root message, you will get an ID added to the root of the message dialog. Also the message area will get an id with `-message` appended after the id given. And the h1 area will get an id with `-title` appended after the id given. And finally the close button (if used) with get `btn-close` appended.
```
tmcconechy, line 102: Message component

tmcconechy (commented):
A Few suggestions. Also can you check off docs/automation.md and the checks on the issue https://github.com/infor-design/enterprise/issues/4498

src/components/autocomplete/autocomplete.js
```diff
@@ -251,6 +253,15 @@ Autocomplete.prototype = {
       this.list = $('<ul id="autocomplete-list" aria-expanded="true"></ul>').appendTo('body');
     }
 
+    // Add test automation ids
+    utils.addAttributes(this.list, this, this.settings.attributes, 'list');
```
EdwardCoyle, line 257: When we implement attributes on the items in the Popupmenu, this is probably going to conflict with that.  Maybe this belongs in there instead? (could resolve the need for the timeout as well)

EdwardCoyle (changes requested):
Looks good to me but just wondering about one thing:

src/components/autocomplete/autocomplete.js
jmacaluso711: I thought that as well, but it seems like it's necessary to add the attributes in the autocomplete specifically. That said, I added the attributes to the `popupmenu` in this PR

EdwardCoyle (dismissed):

tmcconechy (dismissed):

tmcconechy (approved):

EdwardCoyle (approved):

PR #4515, opened 2020-10-20
4098 - Add reveal password feature

### LINKED ISSUE #4098: Input: A method for obscuring sensitive data by default with user reveal method
Per our internal security department, sensitive PII data should be obscured rather than openly display on web pages. Some examples of this date are social security numbers, bank account numbers, drivers license numbers. etc.

The recommendation is displaying these with a * similar to password type and allow a way for the user to view the data,,, either by hovering the mouse over the field, an icon for temp reveal, or clicking an icon for permanent reveal while the page is still loaded.

This should ideally support both inputs and display areas like spans/etc.

Implementing this field by field is possible. But, a feature built-into the framework would prevent duplication.

**Additional context**
And what application do you work on? [e.g. {Infor XYZ}]

tmcconechy:

Adds a feature that adds a toggle button to inputs to hide and show text.

Fixes #4098 

- go to http://localhost:4000/components/input/example-password.html
- click "show" and the text should show (toggle back and forth)
- enter the input and press CTRL + R and the text should also toggle back and forth between hidden and not
- check the MD addition for mistakes
- test it on ES locale on http://localhost:4000/components/input/example-password.html?locale=es-ES and test should translate
- test it on RTL locale http://localhost:4000/components/input/example-password.html?locale=ar-SA


jmacaluso711 (approved):

deep7102:
Looks good! just about the keyboard shortcut `CTRL + R ` also use to reload browser default as `F5`

tmcconechy:
I saw a lot of comments about CTrl +R being the right key https://www.google.com/search?q=keyboard+shortcut+password+reveal&rlz=1C5CHFA_enUS860US860&oq=creveal+password+shortcut&aqs=chrome.2.69i57j0i22i30i457j0i8i13i30.5312j0j7&sourceid=chrome&ie=UTF-8 

Also not seeing it reload the browser? ut even if it does this only takes place in the field, so worst case, exit the field.

We can refine later if needed or we find a better key?

deep7102 (approved):

PR #4558, opened 2020-10-28
4498/4370 - Add attributes API to regular Toolbar

### LINKED ISSUE #4370: Toolbar Menu button items: Add support for Automated testing
We have a requirement to support Automated testing frameworks in Infor products. This means any parts of a control that perform an interaction should easily be located within the document (via a unique statically named id element or surrogate attribute i.e. "data-auto-uid") the automated test framework can then easily fire a click event to interact with that component and not have to perform complex lookups/searches in order to find the element and trigger the event.

This issue applies to drop down menus on Toolbars:
When we create a toolbar that has a Menu button (with items that drop down on click) if we assign an attribute such as "data-auto-uid" to each menu item then the attribute is copied to the "live" version of the down menu, but also remains in the original.  If we look in the method "addItemLinksRecursively" we can see a copy of the original menu is created.  In the case of the id property this is moved to the new copy and removed from the old.  In order to support automated testing we need a method of setting the data-auto-uid attribute on the live copy of the menu.  At present if we set the attribute on the original copy of the menu we end up with duplicates of the data-auto-uid attribute which makes it difficult for automated test frameworks to execute a click event on the correct menu item.

We could wait until the menu/toolbar is initialised and then apply the data-auto-uid attribute to each menu item, but then the attribute would be applied to the original copy of the menu button and not the new "live" copy that is created by the framework.  We could also use the id element instead, but it would be nice to support data-auto-uid (or other attribute) so automated test frameworks can use a consistent approach to finding page elements.

EdwardCoyle:
This PR adds the `attributes` API to the standard toolbar, and labels various elements within.

- related to #4498 
- closes #4370 

Pull this branch, build, run the demoapp.  Open http://localhost:4000/components/toolbar/example-index.html and check the following:

- The main Toolbar element should have an automation id of `my-toolbar`
- The Searchfield's input element should have an automation id of `my-toolbar-searchfield`
- All the Toolbar buttons, including the hidden ones, and except for the action button, should have an automation id of `my-toolbar-button-{0}`, where `{0}` is the index of the button within the buttonset.
- The More Actions button is labeled with `my-toolbar-actionbutton` and passes this to its menu items/subitems using the Popupmenu API.

Also look over the Toolbar docs for completeness.


tmcconechy (commented):
Looks good but something is failing on the tests

tmcconechy (dismissed):
I may have fix the test

tmcconechy (approved):

deep7102 (approved):
🍌

PR #4565, opened 2020-10-30
3993 - Implement sticky header demo

### LINKED ISSUE #3993: Datagrid: Sticky header feature
Would like an option to allow for datagrid column header to freeze upon scrolling similar to Excel.

New data grid option to freeze header.

No alternatives.

**Additional context**
Team: Infor Landmark

tmcconechy:

Added an experimental sticky header feature. Calling this experimental because its limited in a few ways.

a) If there is too many columns then the vertical scroll is on the page (this is the only way that works) and it can cause some wierdness depending on contents
b) frozen columns isnt possible
c) may be other limitations that may make this in the end not viable
d) toolbar is not possible to stick or it causes issues

Fixes #3993

- go to http://localhost:4000/components/datagrid/example-sticky-headers.html
- scroll the page and notice the headers stick at the top


deep7102 (approved):
🍌

EdwardCoyle (approved):

PR #4570, opened 2020-11-02
4527 - Add lookup icon and datagrid fixes

### LINKED ISSUE #4527: Datagrid:  Request to allow a custom icon for the trigger icon of a filterable column
**Describe the feature**
Asking for a way to specify a custom icon for the trigger icon of a filterable column.

**Additional context**
In Landmark working on filtering a code block column.  Currently setting the column to use the filter type of 'lookup'. The current lookup icon will be confusing for landmark users since filtering a code block column we will not be opening a select list dialog.

Additionally when specifying column editorOptions of editable: false, the css background needs to change.

```
 Code snippet for setting filtering of a code block column:
    column.filterConditions =  [ 'contain' ];
    column.filterType = 'lookup';
    column.editorOptions = {
        editable: false,
        click: codeBlockFilterClickFunction
   };
```

**Tasks**

- [x] Add a setting to lookup for any icon
- [x] Allow readonly inputs on the filter row and style them as needed
- [x] Style the clearable x so it is positioned ok

tmcconechy:

Adds a setting to change the lookup icon. Also fix some alignment issues when the lookup is used with clearable and editable options in the datagrid filter row.

Fixes #4527

- go to http://localhost:4000/components/lookup/example-icon
- see that the icon is changed
- try RTL, Themes ect
- go to http://localhost:4000/components/datagrid/test-lookup-filter.html
- see that the lookup column here also has changed icons
- select a value and see that the clearable X is aligned
- try RTL, Themes and row heights


ericangeles (changes requested):
Noticed some UI issues on uplift theme when lookup has clearable options

- Add clearable options
- Go to http://localhost:4000/components/lookup/example-icon?theme=uplift&variant=light&colors=003876
- Does the trigger icon should bigger than the clearable icon? If that's the case, should we vertically align the clearable icon?
- This same goes to RTL

issue1

issue2


On the Datagrid,

- Go to http://localhost:4000/components/datagrid/test-lookup-filter.html
- Change to uplift
- The background color of the input field is not match to the rest of input fields
- Hover the icon, it disappear (probably because the hover icon color is white same with the bg)
- Icon also looks bigger

dg-issue1

- Open up the lookup modal
- Select any value
- Close icon and trigger icon are too close to each other
- Also when hover the close icon, it disappear
- Also on high contrast uplift, it has different bg color compare to the rest of the input fields

tmcconechy:
All fixed @ericangeles good points - i should have tested this better

ericangeles (approved):
🍌 

Also, this PR has the solution for failing test that am getting in my PR here https://github.com/infor-design/enterprise/pull/4572

Issue #4473, opened 2020-10-08
Button: Add Option to remove default "more" tooltip


guoliang:
**Is your feature request related to a problem? Please describe.**

When adding an action button there's no option to remove the default tooltip. Which has the default text "More"

**Describe the solution you'd like**
Allow a title to be set to "" or some other tooltip when using btn-actions

**Additional context**
Check gif for the issue
![default-tooltip](https://user-images.githubusercontent.com/542598/95509420-c80d2500-09b4-11eb-8a1a-52998b99109c.gif)

tmcconechy:
The problem is there is no test here https://github.com/infor-design/enterprise/blob/master/src/components/button/button.js#L387 if the title is overriden.

jbrcna:
**QA Passed**
v4.35.0-dev
https://master-enterprise.demo.design.infor.com/components/button/test-remove-more-tooltip.html
![image](https://user-images.githubusercontent.com/56722184/98802593-7d1c7c80-244e-11eb-8e3b-016153f9f854.png)

PR #4586, opened 2020-11-09
4473 - Fix tooltip to not display when there's no title in setting

ericangeles:

Fixes the tooltip in action button to be not visible when there's not title option.

Closes https://github.com/infor-design/enterprise/issues/4473

- Go to http://localhost:4000/components/button/test-remove-more-tooltip.html
- In `Default: No Tooltip (No Title)` section, hover the action button `...`
- It should not display the tooltip
- In `With Tooltip (Added Title)` section, hover the action button `...`
- This button in this card should show the tooltip says `Tooltip will show!`. (Because it has title prop)


app/views/components/button/test-remove-more-tooltip.html
```diff
@@ -0,0 +1,174 @@
+<div class="row">
+  <div class="one-half column">
+    <h2 class="section-title">Default: No Tooltip (No Title)</h2>
+    <div class="card">
+      <div class="card-header">
+        <h2 class="card-title">Tasks</h2>
+        <button class="btn-actions" type="button" id="btn-1">
```
tmcconechy, line 7: I think this is working as i expect but i'm wondering if there is no title here , shouldnt it show the more tooltip like before?

The top tooltip in this example shows it? I think maybe you can make it such that

1) if title="" means no tooltip at all
2) if no title means show more tooltip as before
3) if title="anythng" show that tooltip?

Also I have fixes for the tests on another PR

tmcconechy (commented):
See above question/comment

tmcconechy (approved):
🍌

EdwardCoyle (approved):

PR #4604, opened 2020-11-12
4550 - Fix click for keyword search on hyperlinks with datagrid

### LINKED ISSUE #4550: DataGrid: Click event is not triggered when using key search
**Describe the bug**
We have columns marked as hyperlinks that relies on the click event to perform different actions. 

A SohoDataGridColumn:
```
            {
                id: column.name,
                name: column.description,
                field: column.name,
                sortable: true,
                hidden: false,
                formatter: this.getOrderNumberFormatter(m3DialogParams.h5Url),
                click: this.orderNumberCallback(m3DialogParams),           // The callback is not triggered when using key search
            };
```
But when using key search and the field is marked as found by key search the attached callback function is never called.

**Expected behavior**
A click event should be triggered so the callback function is called when clicking on a field even when using key search.

**Version**
- ids-enterprise-ng: 7.6.0

**Screenshots**
With no key search the hyperlinks works fine and the click event is triggered when clicking on the order no:
image

But when using key search and the field is marked as found by key search the callback which is attacked to the click event is not triggered when clicking on the order no:
image

deep7102:
Fixed click event was not fire for hyperlinks keyword search results with Datagrid.

Closes #4550

- Navigate to: http://localhost:4000/components/datagrid/example-keyword-search.html
- Open developer tools
- Type `I Love` in keyword search and press enter
- It should come with one row result
- Click anywhere on highlighted part or not highlighted part to hyperlink in column `Product Name`
- See the console should log on clicked


jmacaluso711 (dismissed):

tmcconechy (changes requested):
If you watch the console the click event now fires twice for when you click the <i> also the event data is different on both.

I think the data should be the same and it should just fire once. Possibly by ensuring the click propagate to the anchor tag or some other solution?

tmcconechy (approved):

EdwardCoyle (approved):

PR #4615, opened 2020-11-17
4265 - Add support to select all on active page only for client side with datagrid

### LINKED ISSUE #4265: Datagrid: Client Side Paging doesn't have an option to configure select all (all pages vs the current page)
We want the user to only select the page where they are at when clicking on Select All. This is a client side grid.

We would like to see in the grid options an attribute that we can indicate if the select all should select all rows or just the page.

We don't have an alternative at the moment other than disable the select all feature

deep7102:
Added support to select all rows on active page only for client side paging with Datagrid.

Closes #4265

- Navigate to: http://localhost:4000/components/datagrid/test-paging-select-clientside-multiple-current-page.html
- Click on top `Select All` checkbox (first cell in header first column), to select/unselect all
- Go to any other page, see the selection should not effect to any other page


tmcconechy:
@AnaLauraMartinez Can you test this branch out? One main question i have is if it should "reset" the selections when you change page. So far its not doing that. I think it probably should?

Also @deep7102 i think the name is a bit awkward 'selectAllOnlyActivePage` what about `selectAllCurrentPage` not sure but open to other suggestions @EdwardCoyle is good at that?

AnaLauraMartinez:
> @AnaLauraMartinez Can you test this branch out? One main question I have is if it should "reset" the selections when you change page. So far its not doing that. I think it probably should?

Could we please have a way to tell the widget that we want to keep the selected rows and add to them the selected in the next page?

tmcconechy:
Do you mean?

1. Select all on page one
2. Go to page 2
2. Select all on page two

Now all are selected on noth page one and two?

AnaLauraMartinez:
> Do you mean?
> 
> 1. Select all on page one
> 2. Go to page 2
> 3. Select all on page two
> 
> Now all are selected on the page one and two?

yes, that will work.

tmcconechy:
Ok then this is doing the right thing. Just need to decide on setting name @EdwardCoyle ? what you think?

EdwardCoyle:
`selectAllCurrentPage` seems to make more sense to me

deep7102:
Hi @tmcconechy @EdwardCoyle updated now to use `selectAllCurrentPage`

EdwardCoyle (approved):

ericangeles (approved):
🍌

PR #4635, opened 2020-11-20
NG#926 - Enable use of `span` in place of `a` on the Breadcrumb Item API

EdwardCoyle:
This PR changes the Breadcrumb API to fix a regression where `<span>` tags inside of Breadcrumb Items were not being respected, and converted to anchor tags.  This was technically not respecting existing CSS-only breadcrumb components. The changes include modifying any internal methods that use the `<a>` to instead rely on the `<li>` for reference, and allowing the rendering of either span or anchor.

- Depended on by infor-design/enterprise-ng#926
- Required by infor-design/enterprise-ng#944

- Pull this branch, build, and run the demoapp.
- On http://localhost:4000/components/breadcrumb/example-add-remove.html?theme=uplift.  Try adding breadcrumbs with and without an `href` property.  The ones with an `href` property will be rendered with a hyperlink -- the ones without will be rendered with a `span` tag.  The coloration of these two things will be slightly different but the font size/alignment/padding should be roughly the same.
- Do a general smoke test of http://localhost:4000/components/breadcrumb/list (make sure to test uplift as well).  All should work as expected.
- All tests have been updated to reflect the change so all tests should pass.


tmcconechy (approved):
Just wondering about the wording of the change log. Is it a breaking change that they HAVE to add the span?

deep7102 (approved):

Issue #4617, opened 2020-11-18
Header: Searchfield is also triggered when you open the Appmenu


jbrcna:
**Describe the bug**
when the Appmenu is triggered the search field also opens up

**To Reproduce**
1. Go to https://master-enterprise.demo.design.infor.com/components/header/example-long-title-with-button.html
2. Click on the hamburger button
3. Notice that the search field expands
4. See error.

**Expected behavior**
![Large GIF (1128x526)](https://user-images.githubusercontent.com/56722184/99522881-3b04b500-29d1-11eb-88ed-bcd508ed9588.gif)

**Version**
- ids-enterprise: v4.35.0-dev

**Screenshots**
![Large GIF (1128x526)](https://user-images.githubusercontent.com/56722184/99523071-7a330600-29d1-11eb-9eda-8300b5bcf3d9.gif)

**Platform**
all major browsers

**Additional context**

tmcconechy:
This must have started in 4.33 as it works in https://4320-enterprise.demo.design.infor.com/components/header/example-long-title-with-button.html

I think we should try and fix this asap

jbrcna:
**QA Passed**
v4.35.0-dev
https://master-enterprise.demo.design.infor.com/components/header/example-long-title-with-button.html
notice that when you open that app menu the `Hamburger` icon turns to `X`. Is this working as expected?
![image](https://user-images.githubusercontent.com/56722184/100838622-bbd1af80-34ad-11eb-9640-2d1853c9a9f2.png)

tmcconechy:
@EdwardCoyle I noticed that as well but i didnt say anything. You changed the icon right to a clone? Think that works just fine but just checking...

EdwardCoyle:
@tmcconechy Yeah that was changed in a separate PR: #4646.  The (X) behavior is something that's been there but was no longer working

tmcconechy:
Sure - i think it works better anyways. But as noted it is a change.

jbrcna:
Thank you @tmcconechy and @EdwardCoyle  for clarifying.

PR #4644, opened 2020-11-25
4617 - Fix searchfield also opens when appmenu is triggered

ericangeles:

Fixes the search field also expands when clicking the app menu button.

Closes https://github.com/infor-design/enterprise/issues/4617

- Go to http://localhost:4000/components/header/example-long-title-with-button.html
- Click the app menu button
- Searchfield should not expand


tmcconechy:
@EdwardCoyle when you check this.. why does the animation opening the app menu look wierd? Like one side isnt animating

@ericangeles 
1. isnt the test page http://localhost:4000/components/header/example-long-title-with-button.html ?
1. If so its fixed but it looks like the fix relies on the fact that there is zero other buttons there - does it fix it for sure if there are buttons? Maybe have to make http://localhost:4000/components/header/example-buttons-larger-search.html not collapsed initially to verify this but the logic seems so.
1. Can you fix the text color of the placeholder too. in non focus mode its "yellow" not white like focus

Screen Shot 2020-11-25 at 9 45 30 AM

ericangeles:
> @EdwardCoyle when you check this.. why does the animation opening the app menu look wierd? Like one side isnt animating
> 
> @ericangeles
> 
> 1. isnt the test page http://localhost:4000/components/header/example-long-title-with-button.html ?
> 2. If so its fixed but it looks like the fix relies on the fact that there is zero other buttons there - does it fix it for sure if there are buttons? Maybe have to make http://localhost:4000/components/header/example-buttons-larger-search.html not collapsed initially to verify this but the logic seems so.
> 3. Can you fix the text color of the placeholder too. in non focus mode its "yellow" not white like focus
> 
> Screen Shot 2020-11-25 at 9 45 30 AM

Right. My bad, wrong link.

EdwardCoyle:
@ericangeles @tmcconechy The weird page bounce is [from this](https://github.com/infor-design/enterprise/blame/master/app/views/components/header/layout.html#L3), which I added for some of the header test layouts that expand/collapse the height of the header.  Most of the examples in this folder don't have a hamburger button so I didn't see this before, but the issue is this [`animate-containers` class in the demoapp](https://github.com/infor-design/enterprise/commit/b1dc36ca0366e2203b3632263cf9a2aa6c40a62c#diff-6b700b8ba747b672794607561895ccc8d6bbeeb013773aed723536944bc56172R80) that animates the height.  Seems that it's blowing away the normal horizontal animation of the containers.

Some of the other tests look funny during animation if you drop this class, so for this page specifically, we might want to change the rules overall to [include the normal animation too](https://github.com/infor-design/enterprise/blob/master/src/components/applicationmenu/_applicationmenu.scss#L12-L13):
```
.page-container {
  @include transition(
    height 200ms cubic-bezier(0.17, 0.04, 0.03, 0.94)
    left 300ms cubic-bezier(0.17, 0.04, 0.03, 0.94),
    box-shadow 300ms cubic-bezier(0.17, 0.04, 0.03, 0.94)
  );
}
```

EdwardCoyle (dismissed):

tmcconechy (dismissed):

EdwardCoyle (approved):

Issue #4640, opened 2020-11-24
Homepage (Editable): Widget border color issue in uplift dark theme


volante007:
**Describe the bug**
The border color for widgets on an editable homepage behaves different in uplift dark theme compared to other themes / colors. After hovering, the border does not change back correctly. Also, if items have been hovered and then edit mode is exited, the border is not reverted.

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/homepage/example-editable.html?theme=uplift&variant=dark&colors=0066D4
2. Notice the correct border around the widgets
2. Hover a widget, see the blue border (correct, or should perhaps be white?)
3. Hover out of the widget, see the border become white (incorrect)
4. Click "Toggle Edit Mode" button to make the Homepage non-editable, see the white border remains (incorrect)

**Expected behavior**
The hover border color is removed when hovering out of the widget / exiting edit mode. 

**Version**
- ids-enterprise: v4.34.2

**Additional context**
Other variants in uplift + all variants in subtle theme seem to look fine. This isn't major, just looks a bit glitchy in dark theme

jbrcna:
The issue is now resolved. Thank you

PR #4660, opened 2020-12-03
4640 - Fix border behavior when hovering the card in editable mode

ericangeles:

Fixes the behavior of border when hovering the card/widget in editable mode.

Noticed that the color of draggable/resizable were not using our color palette. Made some adjustments on scss file for us to easily adjust the color/s on different themes and variants (If we think we need to change it).

Expectation in this fix,

- Basically, the border color of widget/card (in editable) when hovered, will be blue (by default) on all themes and variants. Maybe we can discuss the color of it (I see white looks good on dark mode).
- When hovering out, it should retain the default border color of the widget/card. There should be no white border color.

Open for discussions.

Closes https://github.com/infor-design/enterprise/issues/4640

- Go to http://localhost:4000/components/homepage/example-editable.html
- Hover the widget. (It should have blue border color)
- Hover out. It should go back to its original border color. No white border color will see.
- Test this to different combinations of themes and variants


tmcconechy (changes requested):
- on subtle - dark mode - suggest we keep the borders as before 
- on subtle - high contrast mode- the bottom one has an extra border around
- on uplift - high contrast mode- the bottom one has an extra border around
- on uplift - dark contrast mode - the top two has an extra border around (should be like the bottom one)

basically i thin the hover should just be the blue around them all and thats the only state?

ericangeles:
Hi @tmcconechy, I don't see these issues

on subtle - high contrast mode- the bottom one has an extra border around
on uplift - high contrast mode- the bottom one has an extra border around
on uplift - dark contrast mode - the top two has an extra border around (should be like the bottom one)

Do you have a screenshot to identify these?

Thanks

EdwardCoyle (approved):
:banana: Looked good to me, @tmcconechy seems like those issues are fixed now

tmcconechy:
What i mean is ... In dark theme(s) we changed the border and i think it looked better before

http://localhost:4000/components/homepage/example-editable.html?theme=soho&variant=dark&colors=2578a9

Before
Screen Shot 2020-12-09 at 9 44 29 AM

Now
Screen Shot 2020-12-09 at 9 44 10 AM

I think thats all i see now.

ericangeles:
Thanks, @tmcconechy @EdwardCoyle. All set!

EdwardCoyle (approved):

tmcconechy (approved):
Thanks Eric looks good now 👍🏻  🍌

PR #4681, opened 2020-12-11
4532 - Add optional "More Actions" button to Tabs

### LINKED ISSUE #4532: Tabs: Enhancement to tab panel header to include a menu button
I have another enhancement request for the Supply Chain Management product.  We have a requirement to include a menu button within the tab panel container as follows:

tabmenubutton1

Basically, this is a design that was proposed some time ago by the hook and loop team as a replacement for our existing design that includes a floating expandable toolbar (see bottom).  The toolbar is not exactly SohoXI compliant, so we’d like to replace things with the design proposed by H&L.  I think it would be more than enough to simply include a drop down menu (with the …) rather than attempting to include other visible buttons as well.  I think then the size calculation is also simplified as you’d only calculate the width of the drop down menu with the 3 dots.

I think we can already do something similar in module tabs, but we’d also like this in the standard tab panel.  Ideally, we’d like a quick turnaround on this as we aim to start to rewrite our existing layouts in the next few weeks. (appreciate you’re probably busy with other things though)  A summary of what we’re trying to do can be found in the following Jira: https://jira.infor.com/browse/IBP-23529.  In a nutshell we’re rewriting some old legacy layouts to use SohoXI home page type layouts.

Old menus look like this (they float across content in order to maximise real estate on the screen).  Tab panels in this view aren’t SohoXI, so they also need to be replaced.

 tabmenu_old_collapsed

Menu expanded.
 
tabmenu_old_expanded

EdwardCoyle:
This PR adds in styling support for a "More Actions" button that sits in-line with a horizontal tab set.

Closes #4532

- Pull this branch, build, run the app.
- Open http://localhost:4000/components/tabs/test-more-actions-button?theme=uplift.  Check the more actions button style against the tabset.  Shrink the page and make sure it works with the "More Tabs" dropdown.
- Also test this with an add button:  http://localhost:4000/components/tabs/test-more-actions-button-add-button.html?theme=uplift
- On both tests above, use the keyboard and make sure the button is accessible.  Also try shrinking the page (which displays the "More Tabs" dropdown button) and test keyboard handling there as well.
- Test both on RTL, also check Subtle styles.


tmcconechy (changes requested):
Screen Shot 2020-12-14 at 9 50 35 AM

1. Seems like the icon sits a little low when focused the bottom border is touching 
2. How does the keyboard work when i have it at this smaller size the overflow keyboard conflicts. I.E. try tabbing vs arrowing to the new button. Should probably be arrow but only tab works?

EdwardCoyle:
@tmcconechy reworked this and added keyboard support for the new button.

tmcconechy (approved):

ericangeles (approved):

PR #4698, opened 2020-12-18
4693 - Add new clear callback to lookup

### LINKED ISSUE #4693: Datagrid: Need to be notified when clearable filter icon is clicked
Describe the bug
Not notified when the clearable filter icon is clicked.  Issue exists when the grid options has filterWhenTyping set to false.

1. Using 'test-lookup-filter.html'
2. Add the property of filterWhenTyping: false to the grid options
3. Go to 'test-lookup-filter.html'
4. Perform a lookup on the filter row for Product Id column
5. Select a value
6. Click on the clear icon on the filter row for Product Id column 

Landmark will need to notified to react to the filter value being cleared.

Version
Landmark is on v4.35.2

tmcconechy:

Fixes #4693

- go to http://localhost:4000/components/datagrid/test-lookup-filter-click-clear
- click on the lookup -> it will populate the lookup field and show the X (no filtering will happen as this wasnt the point)
- click on the clear X 
- check the console and you can see that the clear callback was called

### DIFF
--- a/src/components/lookup/lookup.js
+++ b/src/components/lookup/lookup.js
@@ -58,7 +58,7 @@ function addSuffixToAttributes(parentAttrs = [], childAttrs = [], suffix) {
  * @param {function} [settings.clear] Provide a special function to run when the clear x is pressed
  * @param {string} [settings.clearArguments={}] If a clear method is defined, this flexible object can be passed
  * into the clear method, and augmented with parameters specific to the implementation.
-  * @param {string} [settings.field='id'] Field name to return from the dataset or can be a function which returns a string on logic
+* @param {string} [settings.field='id'] Field name to return from the dataset or can be a function which returns a string on logic
  * @param {string} [settings.title] Dialog title to show, or befault shows  field label + "Lookup"
  * @param {string} [settings.icon] Swap out the lookup id for any other icon in the icon set by name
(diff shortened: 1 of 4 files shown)


src/components/lookup/lookup.js
```diff
@@ -55,7 +55,10 @@ function addSuffixToAttributes(parentAttrs = [], childAttrs = [], suffix) {
  * @param {function} [settings.click] Provide a special function to run when the dialog opens to customize the interaction entirely.
  * @param {string} [settings.clickArguments={}] If a click method is defined, this flexible object can be passed
  * into the click method, and augmented with parameters specific to the implementation.
- * @param {string} [settings.field='id'] Field name to return from the dataset or can be a function which returns a string on logic
+ * @param {function} [settings.clear] Provide a special function to run when the clear x is pressed
+ * @param {string} [settings.clearArguments={}] If a clear method is defined, this flexible object can be passed
+ * into the clear method, and augmented with parameters specific to the implementation.
+  * @param {string} [settings.field='id'] Field name to return from the dataset or can be a function which returns a string on logic
```
tmcconechy, line 61:
```suggestion
* @param {string} [settings.field='id'] Field name to return from the dataset or can be a function which returns a string on logic
```

deep7102 (approved):

ericangeles (approved):

Issue #4702, opened 2020-12-21
[Editor] Hyperlinks not clickable when readonly


anhallbe:
**Describe the bug**
https://github.com/infor-design/enterprise/pull/4215 seem to have broken the behavior and functionality of Hyperlinks in a read-only editor. The following css (light-theme.css) makes them non-clickable, and the links are indistinguishable
from other text nodes (black color) if it is the first element in the editor. See demo below

```css
.editor-container.is-readonly .editor a {
  pointer-events: none;
}

.editor-container.is-readonly .editor > * {
  color: #1a1a1a !important;
}
```

**Version**
- ids-enterprise: 4.35.0 (but the bug was introduced when we upgraded to 4.33.0)

**Screenshots**

https://user-images.githubusercontent.com/4991954/102797027-62271b80-43af-11eb-98f8-51d4fe0ee6dd.mov

https://user-images.githubusercontent.com/4991954/102797040-64897580-43af-11eb-9d4e-a5cde60188e7.mov

**Platform**
 - Infor Application/Team Name: Homepages
 - OS Version: Windows 10, MacOS Big Sur
 - Browser Name: Chrome, Safari

**Additional context**
https://jira.infor.com/browse/LIME-6575

jbrcna:
**QA Passed**
v4.36.0-dev
https://master-enterprise.demo.design.infor.com/components/editor/test-readonly.html
![image](https://user-images.githubusercontent.com/56722184/103198306-f6026580-4922-11eb-85f9-c00116ec0f9c.png)

anhallbe:
@ericangeles @tmcconechy The hyperlinks are now clickable in read-only mode (thanks for the fix!). However they still get the wrong format if they are added at the beginning due to this one:
```
.editor-container.is-readonly .editor > * {
  color: #1a1a1a !important;
}
```


https://user-images.githubusercontent.com/4991954/103878576-e8916c00-50d6-11eb-861f-5b3764e2e619.mp4

ericangeles:
@anhallbe I see. Yeah, I think this should be fix and it can fix quickly. @tmcconechy should we create another issue?

tmcconechy:
@ericangeles no this is fine, can just do a PR and reference this same issue

jbrcna:
**QA Passed**
v4.37.0-dev
https://master-enterprise.demo.design.infor.com/components/editor/test-hyperlink-readonly.html

PR #4706, opened 2020-12-22
4702 - Fix hyperlinks not clickable in readonly state

ericangeles:

Fixes hyperlinks were not clickable when the editor is in readonly state.

Closes https://github.com/infor-design/enterprise/issues/4702

- Go to http://localhost:4000/components/editor/test-readonly.html
- Hyperlink should be clickable


tmcconechy (changes requested):
Sorry this isnt related to the fix but i think we can easily fix it. See the top border when readonly here:

Screen Shot 2020-12-22 at 9 31 02 AM

tmcconechy:
Also @anhallbe can you test this? Seems right but in your videos you have a different background readonly editor? So seems like it could be slightly different

tmcconechy (approved):

EdwardCoyle (approved):

PR #4785, opened 2021-01-27
4730 date validation on datagrid

### LINKED ISSUE #4730: Datagrid/Validation: Date Validation on the Datagrid not working the same as fields
**Describe the bug**
The validate cell function within the datagrid passes the grid information that contains information about the column's setting, eg disabled dates. This is not used by the built in validation rules of date, or availableDate, and the format and other settings are not used in the rules. 

The date rule is unable to parse the date using the sourceFormat, or dateFormat used on the grid and will fallback to the short format for the locale, and this may or may not work.

The availableDate rule does not work as the datePicker options are not extracted from the passed gridInfo.column.editorOptions.

**To Reproduce**
1. Go to 'http://localhost:4000/components/datagrid/test-datepicker-dynamic-disabled-dates.html'
2. Enter by typing '11/11/1111'
3. Open the date picker to see that the date is disables but no error was shown

**Expected behavior**
The standard validations work in the same way if they are either a field or within a datagrid.

**Version**
- ids-enterprise: master

**Platform**
 - Infor Application/Team Name: Pathway
 - OS Version: Windows 10
 - Browser Name: chrome
 
**Additional context**

marknolaninfor:
Use the gridInfo parameter that is passed to the validation via the datagrid. The standard date validations, (date, avaliableDate) did not use these so the datepicker settings could not be used within the validation 

Closes #4730 

Go to 'http://localhost:4000/components/datagrid/test-datepicker-dynamic-disabled-dates.html'
Enter by typing '11/11/1111'
Open the date picker to see that the date is disables but no error was shown


tmcconechy (approved):
LGTM 👍🏻

ericangeles (approved):

PR #4800, opened 2021-02-02
4798 - Fix wrong metadata for resize, reorder and remove card with homepage (4.37.x)

### LINKED ISSUE #4798: Homepage: Resize / reorder events send faulty information for Editable Homepage (Regression)
**Describe the bug**
When resizing/reordering widgets in an editable Homepage, the events that fire do not send correct metadata. The events are sending the size / position the widget had _before_ the event, not the resulting size / position.

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/homepage/example-editable.html
2. Open Dev console
3. Move Widget A (top left) to position of Widget B (top right)
4. Inspect the event metadata being sent in the reorder event

**Expected behavior**
Widget A should be at index 1 in metadata blocks array. Its x value should be 2.

**Actual behavior**
Widget A is at index 0 in metadata blocks array. Its x value is 0.

See similar fault in the metadata blocks when resizing a widget.

**Version**
- ids-enterprise: latest / 4.36.1

**Screenshots**
image

**Additional context**
Works fine on 4.35.0, https://design.infor.com/code/ids-enterprise/4.35.0/demo/components/homepage/example-editable.html
image

deep7102:
Fixed the wrong metadata was sending for resize, reorder and remove card events with Homepage.

Closes #4798

- Navigate to: http://localhost:4000/components/homepage/example-editable.html
- Open developer tools
- Move Widget A (top left) to position of Widget B (top right)
- See the logs in console, should have updated data
- Resize or Remove card
- See the logs in console, should have updated data

### DIFF
--- a/app/views/components/homepage/example-editable.html
+++ b/app/views/components/homepage/example-editable.html
@@ -64,6 +64,5 @@
     .on('removecard', function(event, card, data){ console.log("Homepage: removecard event: ", [card, data])});
   $('.widget')
     .on('resizecard', function(event, card, data){ console.log("Widget: resizecard event: ", [card, data])})
-    .on('reordercard', function(event, card, data){ console.log("Widget: reordercard event: ", [card, data])})
-    .on('removecard', function(event, card, data){ console.log("Widget: removecard event: ", [card, data])});
+    .on('reordercard', function(event, card, data){ console.log("Widget: reordercard event: ", [card, data])});
 </script>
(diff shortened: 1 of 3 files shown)


tmcconechy (approved):
Looks good @deep7102 but one question: What does it mean that there is two events?

Widget: reordercard event
vs
Homepage: reordercard event:
x

deep7102:
Those are from here [example page](https://github.com/infor-design/enterprise/blob/808d319ef32d2e9220914073bb4df1e52a21a818/app/views/components/homepage/example-editable.html#L61-L67). Those events fired on both homepage element and current card/widget.

tmcconechy:
yeah but im wondering if its clear if they should listen on the widget or the homepage element? like which one "should" be used.

EdwardCoyle (approved):

volante007:
Hi,

I noticed a regression with this change, not sure whether to create a new ticket or not.

As you've discussed already, the events fire on both the widget and homepage elements, which I assume need to be kept for backwards compatibility? However, with these changes, the `removecard` event no longer fires on the card, only on the homepage.

The other events fire on both still.

tmcconechy:
@deep7102 Can you take a look pease? What do you think?

deep7102:
Pushed PR https://github.com/infor-design/enterprise/pull/5208

Issue #4763, opened 2021-01-21
Tabs Module: Cannot close tabs on iPad


aaronpikkarainen:
**Describe the bug**
The X button to close module tabs does not work when using an iOS device (iPad/iPhone). There seems to be a problem with the tap/click getting registered on the X. When I test with an Android device, things work ok. The issue only seems to exist on iOS devices.

The X's on the tab overflow menu work ok. It's the normal X's on the tabs that don't work. If you zoom in very close, then sometimes a tap on the X will close the tab (but we can't expect users to do that).

**Note:** I tried simply increasing the size of the X's with a CSS hack, but that didn't help.

**To Reproduce**
1. Launch Safari or Chrome on an iPad or iPhone
2. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/tabs-module/example-index
3. Click the X on a tab to close it
4. Observe that it does not close (occasionally it may close if you are lucky)

**Expected behavior**
The tabs should be easy to close the first time you tap the X button. Without being able to close module tabs, it makes our app very difficult to use on iPads.

**Version**
4.36.1 (and older)

**Platform**
 - Infor Application/Team Name: CSD
 - Device: iPad and iPhone 6s (both using iOS 14.3)
 - OS Version: iOS 14.3
 - Browser Name: Safari and Chrome
 - Browser Version: ?

janahintal:
QA Passed. Issue is now resolved. 
Verified in https://master-enterprise.demo.design.infor.com/components/tabs/example-dismissible-tabs.html?theme=uplift&variant=light&colors=0066D4 & https://master-enterprise.demo.design.infor.com/components/tabs/example-dismissible-tabs.html?colors=0066D4

PR #4852, opened 2021-02-17
4763 - Re-enable dismissing Tabs with taps on iOS

EdwardCoyle:
This PR fixes an issue that was causing iOS's "Link Preview" feature to conflict with Dismissible Tabs, causing them to become impossible to be dismissed.  We were able to fix it by reworking the CSS on tabs to prevent direct clicks on the anchor, causing events to be handled at the tab level as intended.

#4763

- Pull, build, run
- Open an iOS simulator (iPad and iPhone and a recent iOS version were requested)
- Open http://localhost:4000/components/tabs-module/example-index.html
- Try to dismiss either/both of the dismissible tabs.  It should now be possible.
- Also try standard in-page Tabs to smoke test (or any other tabs sample): http://localhost:4000/components/tabs/example-dismissible-tabs.html


EdwardCoyle:
@aaronpikkarainen Please let us know if this fixes your issue

tmcconechy (dismissed):

tmcconechy (approved):
LGTM unless you want to use https://github.com/infor-design/enterprise/pull/4856

EdwardCoyle:
I'm fine with keeping this one to Safari, seems to not affect the desktop site at all

aaronpikkarainen:
@EdwardCoyle, yes this seems to fix our issue well. At first I thought the ".is-safari" would mean Chrome on iOS would still be broken, but I tested Chrome on an iPhone and it works. Looks like Chrome on iOS is just re-skinned Safari.

Thank you for the fix!

ericangeles (approved):

deep7102 (approved):

PR #4856, opened 2021-02-18
4855 - Fix feature detection on ipad 13 and up

### LINKED ISSUE #4855: Environment: Fix feature detection for Ipad
**Describe the bug**
in ipad ios 13 the feature detection changed:

> I opened Settings -> Safari -> Request Desktop Website -> All websites. That option is enabled by default. I disabled it and the proper User Agent is now displayed.
> 
> 
> Mozilla/5.0 (iPad; CPU OS 13_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0 Mobile/15E148 Safari/604.1
> 
> IMO it should be disabled by default and if a user enables it a modal window should be displaying indicating not all web sites will be supported properly. Or maybe you may consider to add the iPad word to the first user agent to allow us to detect if it is an ipad emulating safari desktop or enable mouse event emulation, because I think when it is released several websites will fail.

**To Reproduce**

1. Go to http://localhost:4000/components/environment/example-browser-specs.html
2. Compare this page on ipad 11, 12, 13, 14 -> 13 and 14 are wrong

**Expected behavior**
Correct detection on ios 13 and 14 

**Platform**
 - Device: ipad
 - OS Version: 13/14
 - Browser Name: safari

tmcconechy:

We found out that ipad has new treatment for user agent strings and this caused faulty detection and missing browser classes.

Fixes #4855 

- open the following page http://localhost:4000/components/environment/example-browser-specs.html
- do it on IOS 11, 12, 13, 14 - on an ipad (Can do phone too for fun)
- things should all make sense and the browser classes should show ios 
- on ios 13 and 14 toggle the setting on the device in the settings app under Safari > Request Desktop Website and try it with it on and off


EdwardCoyle (dismissed):
It still shows "Mac OS X" for the OS when `Request Desktop Site` is enabled, but everything else seems correct.  Have a feeling that part is still coming from the user agent.  I'm ok accepting as-is because you probably can't set it to anything else reliably:

**Enabled:**
Screen Shot 2021-02-18 at 12 44 25 PM

**Disabled**
Screen Shot 2021-02-18 at 12 45 56 PM

:

EdwardCoyle:
Just wanted to post the link to the Apple Developer Forums issue I read that helped us figure this out a bit:  https://developer.apple.com/forums/thread/119186

tmcconechy:
@EdwardCoyle fixed the IOS label.... it was no problem

EdwardCoyle (approved):

ericangeles (approved):

deep7102 (approved):

PR #4883, opened 2021-02-25
4403 - Improve Timepicker accessibility

### LINKED ISSUE #4403: Timepicker: Missing Audible Labels

EdwardCoyle:
This PR aims to improve the accessibility of the Timepicker.  It makes the following changes:
- Allow the `combobox` role to work properly on the input field by adding other ARIA roles in the popover/trigger icon.
- Makes the Dropdown fields' labels' visible.
- Enables screen readers to have better descriptions of the elements inside the Timepicker.

Additionally, this PR moves a few things around in some DemoApp layout templates that were being flagged in Axe as accessibility problems (headers in the wrong places, etc).

Closes #4403 

- Install a screen reader such as Chrome Vox
- Pull, run, build
- Open http://localhost:4000/components/timepicker/example-index.html?layout=nofrills
- Activate the screen reader and focus the Timepicker field.  Vox should read out something like "Timepicker. Use Arrow Keys to Select. Combobox, Collapsed."
- Press the down arrow to activate the Popup. When it opens, Vox should read out something like "Entering Dialog.  Timepicker Popup Dialog. Hours, 1, Menu Button, Collapsed."
- Tab / Shift+Tab between the Dropdown inputs (which should now also display visible labels for each field). When focused, each should read their label and currently-selected contents (fx: "Hours, 1" or "Minutes, 00"), followed by their role and state (fx "Button, Collapsed/Expanded").
- While the Timepicker is open, run an Axe Scan with dev tools.  The only issues that will be present are color contrast issues (there should be none regarding missing/incorrect ARIA-roles)

_Additional Pages to test:_
- http://localhost:4000/components/dropdown/example-index.html?layout=nofrills - When the Dropdown is open run an Axe Scan with dev tools.  The only issues that will be present are color contrast issues (there should be none regarding missing/incorrect ARIA-roles)
- http://localhost:4000/components/datepicker/example-timeformat.html - Check the open Date/Time picker against Vox
- http://localhost:4000/components/timepicker/test-seconds-picker.html - Check the "seconds" field against Vox


tmcconechy:
looks good but does this need updating https://github.com/infor-design/enterprise/pull/4883/checks?check_run_id=1982120500#step:9:1265 ?

PR #4891, opened 2021-02-26
4838 - Fix accordion events

### LINKED ISSUE #4838: Accordion: On a collapse the open state of the header does not change
**Describe the bug**
When you collapse an accordion option, the open state remains unchanged at "true".

**To Reproduce**
1. Add the below code to the bottom of 'example-accordion-panels.html'
2. Click on any of the accordion options to expand
3. Look at the console and note the open states of the accordion
4. Click on that same accordion option to collapse
5. Look at the console and note that the open state of that option is unchanged

**Code Segment to Add**
```
<script>
  const $accordions = $('div.accordion ');
  function listOpenHeaders() {
    const states = $(this).children('.accordion-header').map((i, header) => {
      return {
        text: $(header).text().trim(),
        expanded: $(this).data('accordion').isExpanded(header)
      }
    }).toArray();
    console.log(states);
  }
  $accordions.on({
    afterexpand: listOpenHeaders,
    aftercollapse: listOpenHeaders
  });
</script>
```

**Expected behavior**
It is expected that on a collapse for an accordion header, the open state would be false.

**Version**
- ids-enterprise: v4.38.0-dev

**Platform**
 - Device: (if applicable) N/A
 - OS Version: Windows10
 - Browser Name: chrome
 - Browser Version: 88.0.4324.150 (Official Build) (64-bit)

tmcconechy:

The the afterexpand and aftercollapse events fired before the states are set. Have a look @EdwardCoyle not sure why they might be like that?

Fixes #4838

- go to http://localhost:4000/components/accordion/example-accordion-panels.html
- paste the folling in the console
```
  const $accordions = $('div.accordion ');
  function listOpenHeaders() {
    const states = $(this).children('.accordion-header').map((i, header) => {
      return {
        text: $(header).text().trim(),
        expanded: $(this).data('accordion').isExpanded(header)
      }
    }).toArray();
    console.log(states);
  }
  $accordions.on({
    afterexpand: listOpenHeaders,
    aftercollapse: listOpenHeaders
  });
```
- expand and collapse the first pane and check the result has the correct value for expanded in the array..

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### v4.38.0 Fixes
 
 - `[Application Menu]` Fixed visibility of expander icon on classic theme. ([#4874](https://github.com/infor-design/enterprise/issues/4874))
+- `[Accordion]` Fixed an issue where the afterexpand and aftercollapse events fired before the states are set.  ([#4838](https://github.com/infor-design/enterprise/issues/4838))
 - `[Breadcrumb]` Fixed unnecessary scrollbar in safari on a flex toolbar. ([#4839](https://github.com/infor-design/enterprise/issues/4839))
 - `[Datagrid]` Added the ability to use shift click to select in mixed selection mode. ([#4748](https://github.com/infor-design/enterprise/issues/4748))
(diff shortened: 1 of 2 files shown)


EdwardCoyle (approved):
🤷‍♂️  Not sure why it was originally this way, but this seems fine to me

PR #4928, opened 2021-03-09
4925 - Unsetting height:auto when pie is no longer small

### LINKED ISSUE #4925: Pie/Donut charts don't fill entire space when browser resizes
**Describe the bug**
When resizing the browser window small enough so that the donut or pie goes into small view, then making the browser window bigger again, the donut or pie does not resize itself bigger again.

**To Reproduce**
This is a little tricky to reproduce, but:

1. go to https://design.infor.com/code/ids-enterprise/latest/donut 
2. make the window smaller _very slowly_. The donut goes into small view. 
3. Then, make it wider again.

**Expected behavior**
The donut/pie should get bigger again.

**Version**
4.37.1

**Platform**
 - Infor Application/Team Name: Infor Nexus
 - OS Version: MacBook Pro
 - Browser Name: Chrome
 - Browser Version: 88

**Additional context**
I think this can be fixed in pie.js's build by tacking an else to the if statement where the height:auto is added, to reset it to the normal size:
`else if (this.element[0].style.height === 'auto') { // unset height from auto if previously set so chart can fill space again
      this.element.css({ height: '' });
    }`
Happy to add a pull request if that's helpful.

tfavorite:
Explain the details for making this change. What existing problem does the pull request solve?
Removes height: auto when pie no longer needs to be rendered in small dimensions.

Related github/jira issue (required):
Fixes #4925

Steps necessary to review your pull request (required):
On a page with pie or donut charts, resize browser window slowly. Then make it bigger


tmcconechy (changes requested):
Requesting changes as im not sure if its actually fixed the issue in some cases. But if this is good enough for you im willing to let it go in

tmcconechy (approved):
Since this is a minor improvement will go with this fix for now

Issue #4950, opened 2021-03-16
Datepicker: The Apply and cancel buttons cannot be reached in small screens.


SofiK:
**Describe the bug**
When we open a datepicker in a small viewport such as on a phone, it is not possible to access lower part of the datepicker, including the Apply and Clear buttons.

**To Reproduce**

1. Go to https://master-enterprise.demo.design.infor.com/components/datepicker/example-timeformat.html 
2. Make the browser smaller so that the 4 datepickers are just showing
3. Click the top datepicker
4. Try to click the Apply button

**Expected behavior**
This is perhaps an edge case, but perhaps (if possible) there could be a scroll in the datepicker that could be used in these cases?

**Version**

- ids-enterprise: 4.38.0

**Screenshots**
![image](https://user-images.githubusercontent.com/8664975/111295959-3f25a400-864c-11eb-8665-45c1622789f0.png)

**Platform**
 - Infor Application/Team Name: Homepages
 - OS Version: Windows 10 (but could be reproduced in others as well)
 - Browser Name: Chrome
 - Browser Version: Version 89.0.4389.82 (Official Build) (64-bit)

**Additional context**
-

tmcconechy:
Yeah its a bit of a problem because there is so much crammed into that dialog. I guess we could try to make it scrollable but not 100% sure how easy that would be. Should be doable (perhaps for 5.0)

Workaround -> resize the page?

janahintal:
Apply & Cancel buttons can be reached already Portrait screens in mobile, but the outline seems cut off. 
![image](https://user-images.githubusercontent.com/45757016/113144359-ba4ba480-925f-11eb-8ae7-3db31493b021.png)

Issue still exist in Landscape mobile view. Kindly see image below. 
![image](https://user-images.githubusercontent.com/45757016/113144410-c7689380-925f-11eb-99aa-d9075860f6a3.png)

tmcconechy:
Lets make a new issue for these as - as i think we need complete reproduction/context for it and dont have anymore bandwidth for this issue right now. Thanks!

PR #4957, opened 2021-03-17
4950 - Fix datepicker buttonset visibility on small screens

ericangeles:

Fixes the datepicker apply and cancel buttons where unable to see/reach on a small screens such as small phones (iPhones mainly) and when resizing the browser to a smaller height.

Added a vertical scrollable when viewing it and adjust some elements of datepicker to accommodate more to the smaller screens. It also supports both portrait and landscape orientation view of the small screen phones.

Closes https://github.com/infor-design/enterprise/issues/4950

- Go to http://localhost:4000/components/datepicker/example-timeformat.html
- Open the dev tools and toggle device toolbar
- Select iPhone 5/SE `(Since 320px is the minimum width that we support)`
- Test first on portrait orientation: Open the datepicker per input field
- Datepicker should fit to the screen
- Then test in landscape: Open the datepicker per input field
- Datepicker should be vertically scrollable and able to see the buttonset
- In browser, resize the height to at least `500px`
- Then open the datepicker per input field to test
- It should be vertically scrollable and able to see the buttonset


EdwardCoyle (changes requested):
The scrolling in general works pretty well, but on the last step you mentioned I'm seeing this in Mac OSX Chrome:

Screen Shot 2021-03-17 at 10 01 44 AM

ericangeles:
> The scrolling in general works pretty well, but on the last step you mentioned I'm seeing this in Mac OSX Chrome:
> 
> Screen Shot 2021-03-17 at 10 01 44 AM

@EdwardCoyle its because I only support up to 500px max height. Let me add more px to accommodate this. Thanks!

tmcconechy:
Also @ericangeles i see that the text is cut off... with the .... can we fix this as well? Myabe move the text over a bit if we have to. I saw this before on some smaller devices but didnt get a chance to look at it
Screen Shot 2021-03-18 at 10 25 11 AM

tmcconechy:
You fixed all this right @ericangeles ? Im not seeing my issue anymore and it looks pretty good to me except for one thing:

Screen Shot 2021-03-22 at 9 54 45 AM

See how the container isnt extending to the edges anymore?

ericangeles:
@tmcconechy yes I made some fixes in text being cut off. I didn't catch what your saying. So you want the popover datepicker to be adjusted at the edge of the screen?

tmcconechy:
Yes i havnt seen it look like this before? Maybe someone used margin instead of padding? in the non mobile view it looks ok.

ericangeles:
@tmcconechy, this is the look from the latest version. It looks not proportion to me. Should we still make the datepicker popover adjusted to the edge of the screen? I also noticed that the `monthview-container` and the `monthview-popup` don't have the same width, and it looks odd.

datepicker

tmcconechy:
Yes, i think we should probably try as it looks quite weird.

1. It should be centered in the frame
2. Should not be a gap where the lines around the buttons touch the edge

tmcconechy:
@ericangeles im not seeing any change? But i do see the problem. Its this rule here highlighted  in the screen shot. I dont think this rule is correct...
Screen Shot 2021-03-24 at 9 38 54 AM

ericangeles:
So the goal would be to look like this?

mb

tmcconechy:
I think that it should look "roughly" like this:

Screen Shot 2021-03-25 at 8 56 19 AM

- for smaller devices to make it easier can reduce the space between the edges and side of the device
- should be centered
- "lines" should extend to the bottom and edges

tmcconechy (approved):
Looks good now 👍🏻

EdwardCoyle (approved):

Issue #3144, opened 2019-11-19
Charts: Changing the chart type doesn't remove old class name


kvaldillez:
You can reproduce locally by initializing a chart with a pie chart type and then calling the `.chart(settings)` with a different chart type and a different dataset (optional). You can also reproduce on the IDS demo by:

1) Going to: https://design.infor.com/code/ids-enterprise/latest/demo/components/pie/example-
2) Open console and enter:
`var areaData = [{ data: [{ name: 'Component D', value: 10.1, id: 'A', tooltip: 'Component D <b>{{percent}}</b>' }, { name: 'Component E', value: 12.2, id: 'B', tooltip: 'Component E <b>{{percent}}</b>' }] }];`
`$('#pie-chart-example').chart({type: 'area', dataset: areaData});`
3) Observe that the `chart-container` class still has `chart-pie` even though it is no longer a pie chart.

Expected: `chart-pie` class to be removed and replaced by new chart type class (`line-chart` in this case).

tmcconechy:
Seems like the destroy of chart is not doing removing the class.

kvaldillez:
Yes, I tried doing a manual `.destroy()` on it and then running `.chart(settings)` again which ended in the same result of the previous class still being there.

tmcconechy:
yeah i think the chart "reseting" is also calling destroy. So that's where the fix would lie.

kvaldillez:
FYI - As a workaround, I'm removing all classes from the container and then adding back in `chart-container` prior to calling `.chart(settings)` which then adds the correct chart class. Not ideal but it works.

tmcconechy:
Yeah that works. I think our code would individual destroy each component (chart) and each should strip its class.

tmcconechy:
Descoping some older issues we didnt get a chance to complete.

tfavorite:
@tmcconechy we'd still like this addressed at some point. What's the correct way to keep this in your backlog?

tmcconechy:
Ok, For more details:

We can fix it but i was thinking that going forward it would be individual charts so you won't be able to switch the type depending on chart. For example pie and donut is probably a setting but pie to bar is a seperate chart. But bar to bar grouped is the same chart ect...

So you will in the future have to just invoke one or the other and hide/show. But the fix for this in the meantime is fairly simple so can do it but going forward it wont be possible anymore. (Also the classes wont matter so much anyways)

janahintal:
This issue is now resolved. 
[...]

PR #4965, opened 2021-03-18
439 changing the chart type doesnt remove old class name issue 3144

danielortiz1982:
I added a destroy method inside of the charts objects that will check the type of chart being used and adjust the class name based on it. This method will run right before the **this.init()** that contains the chart build logic.

Fixed https://github.com/infor-design/enterprise/issues/3144

Include:
- Navigate to the any of the chart components, for this sample I have used the pie chart.
(http://localhost:4000/components/pie/example-index.html)
- Right click and select the inspect, then navigate to the console.  
- Paste in the follow code snippet inside the console to change the chart as outlined by **kvaldillez** in issue #3144
**Code Snippet** 
`var areaData = [{ data: [{ name: 'Component D', value: 10.1, id: 'A', tooltip: 'Component D <b>{{percent}}</b>' }, { name: 'Component E', value: 12.2, id: 'B', tooltip: 'Component E <b>{{percent}}</b>' }] }];
$('#pie-chart-example').chart({type: 'area', dataset: areaData});`
- Press enter, the chart should update along with the DOM classes. 
**Screenshot**
example-index
preview
Screen Shot 2021-03-18 at 11 19 11 AM
Screen Shot 2021-03-18 at 11 19 27 AM
Screen Shot 2021-03-18 at 11 19 40 AM
Screen Shot 2021-03-18 at 11 19 51 AM

### DIFF
--- a/src/components/charts/charts.js
+++ b/src/components/charts/charts.js
@@ -8,21 +8,29 @@ const charts = {};
 
 charts.destroy = function destroy(el, s) {
   const e = el[0] === undefined ? window.document.getElementsByClassName('chart-container')[0] : el[0];
-  e.className = 'chart-container';
 
-  switch (s.type) {
-    case 'pie':
-      e.classList.add('chart-pie');
-      break;
-    case 'bar':
-      e.classList.add('bar-chart');
-      break;
-    case 'line':
-      e.classList.add('line-chart');
-      break;
-    default:
-      e.className = 'chart-container';
-  }
+  const chartClasses = [
+    {
+      chart: {
+        type: "pie",
+        containerClass: ['chart-pie', 'has-right-legend']
+      }
+    },
+    {
+      chart: {
+        type: "line",
+        containerClass: ['line-chart']
+      }
+    },
+    {
+      chart: {
+        type: "bar",
+        containerClass: ['bar-chart', 'bar-chart-stacked']
+      }
+    }
+  ];
+
+  chartClasses.forEach( arrItem => e.classList.remove(...arrItem.chart.containerClass));
 };
 
 // Reference to the tooltip
--- a/src/components/charts/charts.js
+++ b/src/components/charts/charts.js
@@ -6,7 +6,7 @@ import { theme } from '../theme/theme';
 
 const charts = {};
 
-charts.destroy = function destroy(el, s) {
+charts.destroy = function destroy(el) {
   const e = el[0] === undefined ? window.document.getElementsByClassName('chart-container')[0] : el[0];
 
   const chartClasses = [   
--- a/src/components/charts/charts.js
+++ b/src/components/charts/charts.js
@@ -7,9 +7,11 @@ import { theme } from '../theme/theme';
 const charts = {};
 
 charts.destroy = function destroy(el, s) {
-  const e = el[0];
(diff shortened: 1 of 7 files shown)


tmcconechy (approved):
Looks good 🍌

src/components/charts/charts.js
```diff
@@ -6,6 +6,24 @@ import { theme } from '../theme/theme';
 
 const charts = {};
 
+charts.destroy = function destroy(el, s) {
+  const e = el[0];
+  e.className = 'chart-container';
```
deep7102, line 11: This looks good approach to handle this issue, only concern about using `className` here will remove all end-user classes if there is any of them in use for some other purpose. So I purpose as below, map the classes we use and remove only them. Modify the `chartClasses` object as you need for which charts to be added or removed.
```
charts.destroy = function destroy(el, s) {
  const e = el[0];
  const chartClasses = {
    'bar': ['bar-chart'],
    'bar-grouped': ['bar-chart', 'bar-chart-grouped'],
    'bar-stacked': ['bar-chart', 'bar-chart-stacked']
  };

  if (e && chartClasses[s.type]) {
    const newClassList = ['chart-container', ...chartClasses[s.type]];
    const chartClassesAll = [].concat(...Object.values(chartClasses));
    e?.classList.remove(...chartClassesAll);
    e?.classList.add(...newClassList);
  }
};
```

danielortiz1982: Hi @deep7102 
Thanks for the feedback... I have tested every scenario for each of the different charts and everything seems to be working as expected. Let's get together sometime tomorrow so I can better understand the concerns you've outlined above along with your proposed solution. I popped in your sample code and the chart-pie class still persist in the DOM. I will ping you tomorrow so we can connect.  
Screen Shot 2021-03-18 at 5 41 29 PM

deep7102: Hi @danielortiz1982 sure you can ping me if you need, the issue I see if user having any class/es using with this element then it will wipe all of them for example `<div id="bar-example" class="chart-container some-user-class"></div>` here it will wipe off `some-user-class`.

That proposed solution code is just sample in order to use you need to add/remove in `chartClasses` object. So in case of `Pie` need to add as below: 
```
const chartClasses = {
    'bar': ['bar-chart'],
    'bar-grouped': ['bar-chart', 'bar-chart-grouped'],
    'bar-stacked': ['bar-chart', 'bar-chart-stacked'],
    'pie': ['chart-pie']
  };
```

tmcconechy: I think deepak is correct. It's probably safer. So instead of removing the className all together just remove each of the N possible class names from there and then add the new one.

There could be other classes added to the root (unrelated to the chart classes in question)

deep7102 (changes requested):
Hi, I just thought of that way you may use your own way if you have another solution, it was in my mind so save some time for you I just put it in there.

Just make it remove all possible chart class names then add the new one - but don't remove all classNames

src/components/charts/charts.js
danielortiz1982: @deep7102 I see your point, I did not take into count that end users implementation of the chart on their end.  
Thanks for the feedback @deep7102  and @tmcconechy

deep7102 (approved):

Issue #4751, opened 2021-01-20
Editor: customize buttons missing separator


jbrcna:
**Describe the bug**

1. Go to https://master-enterprise.demo.design.infor.com/components/editor/example-customize-buttons.html
2. See issue
3. Notice that there's no separator after the dropdown

**Expected behavior**
![image](https://user-images.githubusercontent.com/56722184/105168342-f6260780-5b54-11eb-8ccb-2513120ff20a.png)

**Version**
- ids-enterprise: v4.37.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/56722184/105168474-24a3e280-5b55-11eb-8323-df18939649f8.png)

**Platform**
All major browsers

**Additional context**

tmcconechy:
For this issue i think either there really is a bug in editor.js or you just need to modify this setting https://github.com/infor-design/enterprise/blob/master/app/views/components/editor/example-customize-buttons.html#L21 in the exampe.

I would debug the editor code on the page that works and adds a seperator http://localhost:4000/components/editor/example-index.html and if the fix is changing editor/example-customize-buttons.html thats perfect

chrisakinrinade:
Ok thanks! I'll take a look.

PR #4986, opened 2021-03-23
4751 - Fix Editor separator

chrisakinrinade:
Fixed an issue where separators that appear immediately after the header drop-downs don't appear.

#4751 

Go to http://localhost:4000/components/editor/example-customize-buttons.html and see that there is now a separator after the initial header dropdown.

Before:
image

After:
image

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -69,7 +69,7 @@
 
       &.is-disabled {
         span {
-          opacity: 0.5;
+          color: $breadcrumb-disabled-color;
         }
 
         .hyperlink {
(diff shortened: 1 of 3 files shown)


src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -69,7 +69,7 @@
 
       &.is-disabled {
         span {
-          color: $breadcrumb-disabled-color;
```
tmcconechy, line 72: Looks like this has an old change in it - perhaps you branched off from a preview branch rather than master. I'd check this out and remove this change / revert it back.

chrisakinrinade: Yup, you're right. Just fixed.

tmcconechy (approved):

EdwardCoyle (approved):
Excellent :banana:

Issue #4982, opened 2021-03-23
Field Filter: short is off in Uplift


jbrcna:
**Describe the bug**

**To Reproduce**

1. Go to https://master-enterprise.demo.design.infor.com/components/field-filter/example-short.html
2. See error
4. Notice that UI  is a bit off

**Expected behavior**
![image](https://user-images.githubusercontent.com/56722184/112152831-aaa7dc80-8c1d-11eb-9ed2-74d83ea26e0b.png)

**Version**
- ids-enterprise: v4.29.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/56722184/112152778-9d8aed80-8c1d-11eb-96a6-b265827e0aa6.png)

**Platform**
All

**Additional context**

tmcconechy:
Yes looks like the uplift theme is a little bit off.
1. Missing a border
2. And the focus state is only partly working covering the center
3. When you open the list it is too long for the field.

jbrcna:
issue is now resolved. v4.39.0-dev
https://master-enterprise.demo.design.infor.com/components/field-filter/example-short.html
![image](https://user-images.githubusercontent.com/56722184/113104355-02080700-9233-11eb-8137-881d9061bac5.png)

PR #4988, opened 2021-03-24
4982 - Fix missing border and dropdown list misalignment in uplift

ericangeles:

This PR fixes missing border in the input field and dropdown list does not properly align when it opened. Removing the border right color which is transparent also fixes the alignment of the dropdown list.

Closes https://github.com/infor-design/enterprise/issues/4982

- Go to http://localhost:4000/components/field-filter/example-short.html
- The input field should have border now
- Open the dropdown filter
- List should be align with the input field


tmcconechy (changes requested):
I see two problems...

1. go to http://localhost:4000/components/field-filter/example-short.html
1. press on the button to open the list -> notice that the button gets bigger / arrow placement moves
1. in all themes when you focus the field the border only goes 3/4 around the field

This component isnt that critical so willing to let that go if its hard but if we really want to fix those up i would take a quick look

EdwardCoyle:
@ericangeles I was still seeing some issues with the border/box shadow after the last round of changes.

When the menu button is focused, there's a double-border:
Screen Shot 2021-03-25 at 10 07 42 AM

When the input is focused, the box shadow seems like it's "cut off" by the button:
Screen Shot 2021-03-25 at 10 07 48 AM

ericangeles:
@tmcconechy @EdwardCoyle, something weird was going on in the `New` / `Uplift` version. When the edges of the dropdown and input overlap or on top of one another, it is not visible for some reason. But in the `Classic` version, it behaves as expected.

Although this component is not critical or major, so I fix it in a tricky way.

EdwardCoyle (approved):
@ericangeles looks good, I was worried about hardcoding the z-index here but it doesn't seem to conflict with anything :banana:

tmcconechy:
Possibly make it under 20... Like does 2-3 work? We made this "system" up over time and it could conflict with contextual toolbar

```
20 - Contextual Toolbar
500 - Side Menu, Status Bar
1000 - Overlay
1010 - Busy Indicator
1015 - Popdown
1020 - Dialogs (Modals)
3000 - Popovers
4000 - Lookup
4500 - Drop Down
6000 - Tooltips
7000 - Draggables
8000 - Slide in Message Dialog/Toast
9000 - Sticky Nav / Context Menus
10000 - Full Page Sticky Nav, Search Results
10020 - Skip link
```

tmcconechy (approved):

tmcconechy (approved):

PR #5016, opened 2021-03-30
4209 - Add setting to allow child expand on match only with datagrid

### LINKED ISSUE #4209: Datagrid: Additional allowChildExpandOnMatch type search behavior
**Problem we faced**
When we set allowChildExpandOnMatch=true and find some element its showing all siblings of filtered element
also siblings of parent of filter elements.

**Solution we would like**
we would like solution like when we search element it searches like '**allowChildExpandOnMatch =false**'  with child expansion feature of  '**allowChildExpandOnMatch =true**'

**Explanation**
we expecting search result same as shown is  screenshot with '**allowChildExpandOnMatch =false**' with addition we expecting to able to expand searched node. In this example "Follow up action with Acme Canada"

We tried '**allowChildExpandOnMatch =true**' but it showing so many unnecessary nodes please find screenshot below
Screenshot (11)

Screenshot (10)

deep7102:
Added api setting `allowChildExpandOnMatchOnly` with Datagrid. It will show/hide children match only or all of them this setting only will effect if use with `allowChildExpandOnMatch:true` for Datagrid

Closes #4209

- Navigate to: http://localhost:4000/components/datagrid/test-tree-filter-child-on-match.html
- Filter 1st column (Task) for `Follow up action with Acme Canada`
- Use top switch and checkbox to change the settings
- See the result rows, and should match only or show all the child nodes


src/components/datagrid/datagrid.js
```diff
@@ -11974,7 +11994,7 @@ Datagrid.prototype = {
   },
 
   /**
-  * Unwrap the grid back to a simple div, and destroy all events and pointers.
+  * Unwrap the grid back to a simple div, and destory all events and pointers.
```
tmcconechy, line 11997:
```suggestion
  * Unwrap the grid back to a simple div, and destroy all events and pointers.
```

tmcconechy (dismissed):
This looks right to me. Do you guys want to pull this and try it @dhumalkishor  @hgodse
If not we'll just merge when approved by us and you can try later

rob2d:
looked good. Committed some minor iteration only on demo phrasing to push in faster and letting CI run.

rob2d (approved):

EdwardCoyle (approved):

Issue #4341, opened 2020-09-03
Switch: Disabled Switch On Status Not Easily Discernible


hemantg05:
Disabled ON Switch is not easily discernible.

**Describe the bug**
User can not easily discern whether the switch is ON or OFF.

**Expected behavior**
Disabled ON Switch is should be easily discernible.

**Screenshots**
<img width="338" alt="Disabled-ON-Switch" src="https://user-images.githubusercontent.com/1594815/92126452-7d6e1b00-ee00-11ea-9107-98e82dce2218.png">

**Additional context**
cc @karinthulin, PIM

tmcconechy:
I think this is a valid concern. I looked at a few other examples and most of them include the color in the disabled state for example

<img width="91" alt="Screen Shot 2020-09-03 at 10 07 20 AM" src="https://user-images.githubusercontent.com/814283/92126784-99f15f80-edce-11ea-820f-3d244e2152ac.png">
<img width="85" alt="Screen Shot 2020-09-03 at 10 07 27 AM" src="https://user-images.githubusercontent.com/814283/92126786-99f15f80-edce-11ea-923e-d6c4785c2355.png">

In addition because of accessibility we should really incorporate on/off text in the state as well.

Can design take a look @inforandy

karinthulin:
@tmcconechy this is an issue in PIM application. Can this be addressed?

tmcconechy:
Sure, taking a look this sprint

tmcconechy:
See previous comment. We are just changing the disabled state to have a color so you can easier tell on / off.
Also please fix this in both this repo and https://github.com/infor-design/enterprise-wc as its in both. So two PR's

rob2d:
@tmcconechy tried to make the most minimal change possible, seems would be easier to see with the bg bar than with the circle outline. 

Would this work?
![image](https://user-images.githubusercontent.com/1799905/113204404-44cddd00-923b-11eb-8405-0e9906b55a46.png)

rob2d:
should have re-read the issue before working on that -- adjusting the disabled on state now.

rob2d:
@tmcconechy @hemantg05 

For main theme, does this seem it may be differentiated enough? 
(kept a graphite-colored outline as having a perfectly azure circle even if lighter was a bit confusing)

![image](https://user-images.githubusercontent.com/1799905/113318949-f6721a00-92de-11eb-9d2c-97bd3fedae53.png)

tmcconechy:
[...]

PR #5029, opened 2021-03-31
4341 - Make Switch Disabled State Contrasted from Unchecked

rob2d:

Switch disabled visual state was not contrasted vs checked + disabled.

Partially addresses #4341.
(Need to address on `enterprise-wc` as well to close out)

- pull `4341-disabled-vs-unchecked-switch-contrast`
- run demo server
- visit http://localhost:4000/components/switch/example-index.html?theme=new&mode=light&colors=7928E1
- you should see via the top-right menu that the switch that is checked+disabled has an off-blue tint with grey outline.
- be sure to switch between all 6 variants of: 
-- new/classic + 
-- high contrast/light/dark

image
image
image
image

- [X] A note to the change log.


tmcconechy (changes requested):
Minor suggestions:
- Lets try azure20 instead of 30 (if you tried and think its no good thats fine - its up for debate)
- also in the two dark modes i noticed the bar part is slate but in all other cases its azure

rob2d:
@tmcconechy for the first I actually had tried `azure20` at first. It seemed to blend into the white background quite a bit/low contrast, so that's why I didn't use that in the PR. 

For the second, I think you're referencing dark + disabled + checked?  did this to differentiate from disabled + unchecked.

Actually did try both and realize what you're saying -- the inconsistency felt a little bit dirty there. 

On the second case (dark + disabled + checked), I wasn't sure how to stay in the palette but keep things looking disabled e.g. if the bar was floating, which is similar to the original issue but for dark mode. 

I have found that some people especially WFH environments may have low contrast settings on their monitors or low brightness on non OLED, so the saturation loss for the bar in disabled can help that a tiny bit.

^ thats super subjective though, just a heads up that I did try it/consider it and it's your call with that said to change it or not.

tmcconechy:
OK lets forget about the first part if you tried it and it doesnt seem as good. But this is still an issue:
In the two dark modes i noticed the bar part is slate but in all other cases its azure..

Screen Shot 2021-04-02 at 9 36 34 AM

In all but the dark themes the ---- part of the switch is azure

rob2d:
@tmcconechy that was intentional to differentiate from the non-disabled/active on the dark mode because of the high contrast vs the bg -- but its subjective so will address that.

rob2d:
adjusted:
image

tmcconechy (approved):

EdwardCoyle (approved):
Looks great :banana:

PR #5030, opened 2021-03-31
4820 - Lookup/Timepicker/Datepicker accessibility improvements

EdwardCoyle:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
This PR makes some changes to some trigger field components (Lookup, Datepicker, Timepicker) to convert the "trigger" from just an icon to a button, along with adding various missing ARIA roles, descriptive text, and other minor improvements to improve detection of these components by screen readers.

To test this issue, please pre-install Chrome Vox or another screen reader application.

**Related github/jira issue (required)**:
Closes #4820 

**Steps necessary to review your pull request (required)**:

_Test Lookup_:
- Open http://localhost:4000/components/lookup/example-index?layout=nofrills
- Enable Chrome Vox (or other screen reader)
- Focus the first input field.  The screen reader should speak "Products Lookup.  Press Down Arrow to Select. Edit Text, Required"
- Press the tab key.  The Lookup icon is now focusable and should appear focused.  The screen reader should speak "Lookup Trigger Button"
- Test RTL and check for icon/button alignment: http://localhost:4000/components/lookup/example-index?layout=nofrills&locale=ar-SA
- Test classic theme:  http://localhost:4000/components/lookup/example-index?theme=classic&mode=light&layout=nofrills
- Test non-tabbable example - the new second tabstop should be disabled, and the keyboard behavior should work like before this change: http://localhost:4000/components/lookup/example-not-tabbable?theme=new&mode=light&layout=nofrills

_Test Datepicker_:
- Open http://localhost:4000/components/datepicker/example-index?theme=new&mode=light&layout=nofrills
- Enable Chrome Vox (or other screen reader)
- Focus the first input field:.  The screen reader should speak "Date Field. Press Down Arrow to Select. Combo Box Collapsed".
- Press the Tab key.  The trigger icon is now focusable and should appear focused.  The screen reader should speak "Date Picker trigger button"
- Test RTL and check for icon/button alignment: http://localhost:4000/components/datepicker/example-index?theme=new&mode=light&layout=nofrills&locale=ar-SA
- Test classic theme: http://localhost:4000/components/datepicker/example-index?theme=classic&mode=light&layout=nofrills
- Test non-tabbable example - the new second tabstop should be disabled, and the keyboard behavior should work like before this change: http://localhost:4000/components/datepicker/example-not-tabbable?theme=new&mode=light&layout=nofrills
- Test Calendar Toolbar: http://localhost:4000/components/calendar-toolbar/example-index.html
- 
[...]

PR #5046, opened 2021-04-02
Misc - Add new translations

tmcconechy:

Got this months round of string translations from the translation team and adding them in.

NA

- Nothing to Review


EdwardCoyle:
@tmcconechy do the strings from #5030 need to be part of this too?

tmcconechy:
@EdwardCoyle its too late - i sent these over to the translators a couple weeks ago with what was there at the time and they just came back with these. So those will just be in next the cycle

EdwardCoyle (approved):

Issue #5015, opened 2021-03-30
Toast: draggable toast message at bottom right causes drag issue


pwpatton:
**Describe the bug**
if I have a draggable toast message bottom right and attempt to move it it moves up, but if I let it go and try and move it back down it won't go below it's current position

**To Reproduce**

1. Go to 'example-draggable.html'
2. Set `position: 'bottom right',` on toast message options
3. run  'example-draggable.html' and click show toast message button
4. Drag toast message up then let go
5. grab again and attempt to drag down fails

**Expected behavior**
toast message should be able to be dragged back down to the bottom

**Version**

- ids-enterprise: 4.37+

**Screenshots**
![toast-drag-issue](https://user-images.githubusercontent.com/4737815/113019529-53d36300-9147-11eb-9d31-b07de56c079b.gif)

**Platform**
 - Infor Application/Team Name: Landmark
 - OS Version: windows
 - Browser Name: chrome
 - Browser Version: 89.0.4389.90
 - 
**Additional context**

jbrcna:
The issue is now resolved. 
https://main-enterprise.demo.design.infor.com/components/toast/test-bottom-right-draggable-position.html

PR #5081, opened 2021-04-14
5015 - Fix draggable toast message at bottom right position

ericangeles:

This PR fixes the issue of  where it's unable to drag down the toast message to its current position if the `position` setting is set to `bottom right`.

Closes https://github.com/infor-design/enterprise/issues/5015

- Go to http://localhost:4000/components/toast/test-bottom-right-draggable-position.html
- Click the button to trigger the toast message
- Drag toast message up then let go
- Grab again and try to drag down
- It should be able to move/drag it down

### DIFF
--- a/app/views/components/toast/test-bottom-right-draggable-position.html
+++ b/app/views/components/toast/test-bottom-right-draggable-position.html
@@ -9,6 +9,7 @@
     $('body').toast({
       title: 'Application Offline',
       message: msg,
+      savePosition: true,
       draggable: true,
       position: 'bottom right',
       uniqueid: uniqueid,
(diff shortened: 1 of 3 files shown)


app/views/components/toast/test-bottom-right-draggable-position.html
```diff
@@ -0,0 +1,21 @@
...
+    <button class="btn-secondary" type="button" id="show-toast-message1">Show Toast Message</button>
+  </div>
+</div>
+
+<script>
+  function showToastMsg(uniqueid, msg) {
+    $('body').toast({
+      title: 'Application Offline',
+      message: msg,
+      draggable: true,
```
tmcconechy, line 12:
```suggestion
      draggable: true,
      savePosition: true,
```

Lets add save position or this example looks like its not working properly when you reload the page or click a toast again as it will go back to the old position.

tmcconechy (changes requested):
I Noticed a strange behavior with the saving Try these steps
1. open the toast
2. move it anywhere by dragging
3. reload the page
4. open the toast (should be in the same position)

But i think it can be fixed by adding the option savePosition to the example so we should probably do that

tmcconechy (approved):

deep7102 (approved):

PR #5100, opened 2021-04-16
5074 - Toolbar Resize Menu Stagger & Close Button fix

### LINKED ISSUE #5076: General: Add Private Class Fields

Look at adding https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields to all private properties in the base and components. For justification: we can prevent developers from using stuff they cant or that we might change.

Does this work for methods as well or what will?

- `['@babel/plugin-proposal-private-methods']` add this (check it doesn't add to much to each component in the dist
- refactor all places that need it making decisions 
- test safari/chrome and all r-1 browser

Not using this

**Additional context**
https://github.com/infor-design/enterprise-wc/pull/58

rob2d:

Searchfield in certain cases was staggering when you resize the window. In addition, the X button was briefly hovering slightly higher when this event was triggered.

Closes #5076 
(the functionality fix does not have specific issue -- was just related)

- pull  `5074-toolbar-resize-menu-stagger-and-close-button-fix`  locally
- visit http://localhost:4000/components/toolbar/example-no-action-button.html

(1) trigger staggering issue
- make sure dev tools are closed
- resize window to width slightly above the minimum breakpoint in Chrome
- type any text into search field
- drag browser edge to resize the window horizontally. This can be repeated.
- You should *not* see this staggering happening on each resize in phone size.
searchfield_issue

(2) since the resize stagger was fixed, the second issue of seeing the searchfield X off is harder to spot but can be simulated to check the X is not moving:
- type any text into search field
- open dev tools and go into selection mode
image
- click on the searchfield in the main area where text is, then go to "Elements" on right and remove `is-open` by doubleclicking in the `span` that has `searchfield-wrapper` class.
image
The close button "X" shouldn't move upwards or downwards now.
image

Note: there are not changes in the change-log since there's several searchfield issues that to the end user seem the same. This one had a functional scope as well and could also be isolated really specifically for the style, but doesn't capture all the cases yet.

### DIFF
--- a/src/components/searchfield/_searchfield-new.scss
+++ b/src/components/searchfield/_searchfield-new.scss
@@ -66,12 +66,13 @@
 
   .btn-icon.close {
     overflow: visible;
+
     &:not(.is-empty) {
       top: 6px;
     }
 
     & > svg.close.icon {
-      top: calc( 50% - 6px );
+      top: calc(50% - 6px);
       transform: translateY(-50%);
     }
   }
--- a/src/components/searchfield/_searchfield.scss
+++ b/src/components/searchfield/_searchfield.scss
@@ -542,6 +542,15 @@ html[dir='rtl'] {
   }
 }
 
-.buttonset > .toolbar-searchfield-wrapper:nth-last-child(2) > .icon.close {
-  transform: translateY(0%);
+/*
+  target phone-size case when the menu first hides on resize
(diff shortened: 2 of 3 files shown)


tmcconechy (approved):
LGTM 👍🏻

EdwardCoyle (approved):
Good stuff! 🍌

Issue #5090, opened 2021-04-15
Tabs Module: Tabbing inside clearable searchfield makes the "X" icon disappear


s-werking:
**Describe the bug**
When the focus is set to the "X" icon for a clearable searchfield, only the focus border appears around where the "X" icon should be.

**To Reproduce**
1. Go to http://localhost:4000/components/tabs-module/example-category-searchfield-go-button-home.html
2. Begin typing in the searchfield
3. Tab to access the close button
4. Note that the close button is no longer visible

**Expected behavior**
It is expected that the "X" icon is visible while focused.

**Version**
- ids-enterprise: v.4.51.0-dev

**Screenshots**
Issue seen in the sample:
![Searchfield Close Icon Lost](https://user-images.githubusercontent.com/76002134/114896840-49b39480-9dd6-11eb-94e7-325bb8114a37.PNG)

**Platform**
 - Infor Application/Team Name: lawson
 - Device: (if applicable) N/A
 - OS Version: Windows10
 - Browser Name: chrome
 - Browser Version: 89.0.4389.114 (Official Build) (64-bit)

tmcconechy:
Similar to https://github.com/infor-design/enterprise/issues/5096

janahintal:
QA Passed. Verified in https://4510-beta0-enterprise.demo.design.infor.com/components/tabs-module/example-category-searchfield-go-button-home.html
![image](https://user-images.githubusercontent.com/45757016/116987037-2313ba00-ad01-11eb-8abc-63d160b54a19.png)

Issue #5074, opened 2021-04-13
Toolbar: Misaligned close button in searchfield


janahintal:
**Describe the bug**
Toolbar: Misaligned close button in search field

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/toolbar/example-no-action-button.html
2. Open the search field
3. Minimize the screen to responsive view

**Expected behavior**
The 'x' or close button should be aligned

**Version**
IDS 4.50

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/114550518-dfdfa180-9c94-11eb-9ba5-ecd873115859.png)

**Platform**
All major browser & devices

**Additional context**
N/a

EdwardCoyle:
@rob2d wondering if you could revisit your changes to the Searchfields?  It seems this issue's cropping up in a handful of them.  The searchfield inside the Lookup fields here has the same issue: https://master-enterprise.demo.design.infor.com/components/lookup/example-index.html

I think it's related to using `transform: translateY()` instead of pixel values:
- [Here's one spot](https://github.com/infor-design/enterprise/pull/5022/files#diff-8ebeb31acceb63755764fee625921ac6cf5a0d3a33b0e439471a4d4a9c2e17a1R74-R75)
- [Here's the other](https://github.com/infor-design/enterprise/pull/5022/files#diff-8ebeb31acceb63755764fee625921ac6cf5a0d3a33b0e439471a4d4a9c2e17a1R14-R15)

I didn't remember this when we talked about the original ticket but we'd specifically avoided doing this type of alignment with transforms (I can't remember the context for why) -- bottom line is we should probably revert this and go with pixel values for alignment.

rob2d:
@EdwardCoyle the y-offset was needed for about 5-6 cases with the last ticket for going too far down into fields which I couldn't get via absolute positioning because of the structure of how other cases were wrapped in the field (so that spawned the other issues), and there were nuanced cases where it didn't but missed it. 

I'll try to look at this and isolate it next time I check it out -- maybe can inject a class with more context in these specific cases at component level because it seems there's some conflicting rules. Tried to avoid it since it was my first ticket before and wasn't familiar (though... not the most in that one but hopefully enough).

tmcconechy:
[...]

PR #5102, opened 2021-04-16
Searchfield: Listview-specific fixes for search close Button

rob2d:

Fixes listview-specific alignment issues

Related to #5090 #5074 but does not close these. Pushing as separate PR to isolate testing cases/code as there are many permutations of close button stylings.

- pull `5090-tabs-and-listview-searchfield-close-fix`
- visit http://localhost:4000/components/listview/example-search.html and verify that close button is aligned
image

-visit http://localhost:4000/components/swaplist/example-search.html and verify close button is aligned
image

- also do the same for the same URLs above with ?theme=classic


tmcconechy (approved):
Looks good - im not sure how you want to fix all these similar issues. But it could be a case where you want to do them all in one PR. Because one could break another ect...But do what makes sense thats just a thought. 

I was looking at the "double border" issue and half expecting that to be fixed here but if its later thats fine. If you tackle that it seems like you can change this rule

```
.listview-search .searchfield-wrapper {
  height: 38px;
}
```

EdwardCoyle (dismissed):

rob2d:
@tmcconechy found that we can target the input to just be 100% between both themes so just tweaked the solution a tiny bit (and also good catch, thanks).

rob2d:
Iterated/rewrote one of the more-flakey e2e datagrid tests that was failing on this branch... and that was at least passing locally consistently. Hopefully it does not trigger other cases 🤞

tmcconechy (approved):

PR #5125, opened 2021-04-22
5124 - Fix hierarchy data passing issue

### LINKED ISSUE #5124: Hierarchy: Changes to getActionMenuItems() is breaking JSON.stringify of the data object
Describe the bug
Changes to the method getActionMenuItems is breaking the JSON.stringify of the data object.

1. Go to 'example-context-menu-with-details.html'
2. Open the browser's debugger
3. For Kaylee Edwards record, click on '...' to open the menu items
4. Click on 'Profile' menu item
5. In the console the displayed value for eventInfo.actionReference is just "{"

Expected behavior
The entire value for the actionReference is displayed

Version
- ids-enterprise: v4.39.0-dev

Screenshots

tmcconechy:

The automation id fix caused a regression in the way the hierarchy passes data around. cc @deep7102 

Caused by https://github.com/infor-design/enterprise/commit/e4d8d4c2a15c8ade940e9f617cfb162c38e6e423#diff-00df80da42564b16572a3e535dc8f02bcefdf882b59d0749d6fbab15ba4a5b23R424

I just reverted the change, but if you have a better fix let me know.

Fixes #5124

- http://localhost:4000/components/hierarchy/example-context-menu-with-details.html
- open the ... menu
- select "profile"
- look in the console and this should be a JSON string


EdwardCoyle (approved):

src/components/hierarchy/hierarchy.js
```diff
@@ -421,7 +421,7 @@ Hierarchy.prototype = {
       }
       return `
         <li data-disabled="${a.disabled}" class="${a.menu ? 'submenu' : ''}">
-          <a href="${a.url}"${opAttrStr(idx)} data-action-reference="${JSON.stringify(a.data)}">
+          <a href="${a.url}"${opAttrStr(idx)} data-action-reference='` + JSON.stringify(a.data) + `'>
```
deep7102, line 424: Looks good 🍌 

wonder why it was breaking, did not think it can make the difference both were looking same as below.
```
const val = `<a data-action-reference="${JSON.stringify(a.data)}">`;
```
vs
```
const val = `<a data-action-reference='` + JSON.stringify(a.data) + `'>`;
```

deep7102 (approved):

src/components/hierarchy/hierarchy.js
tmcconechy: I think its weird the way it was done as @krishollenbeck decided to use a single quote. I think its related to that as the JSON in side it has double quotes...

For now lets just get it working...even though stylistically its a bit wierd

deep7102: oh you right, forgot about JSON return could break the double quotes. -thanks

deep7102 (approved):

ericangeles (approved):

Issue #5048, opened 2021-04-02
ListView: Changing selectable from 'mixed' to 'single' does not remove checkboxes 


claudenbach:
**Describe the bug**
Changing the selectable input on a listview from 'mixed' to 'single' is not removing the checkboxes.  It works from when changing the selectable input from 'single' to 'mixed'. 

**To Reproduce**
- Render a listview with selectable set to 'mixed'
- Change the selectable to 'single' (using the updated(newSettings) method)
- Notice that the checkboxes are still rendered

**Expected behavior**
When setting the selectable to 'single', checkboxes should not be rendered

**Version**
- ids-enterprise: v4.37.3

claudenbach:
![Git-5048](https://user-images.githubusercontent.com/39804038/113427534-d0aa4b00-939a-11eb-9b3c-bca800c53bee.gif)

CindyMercadoReyes:
This issue is now resolved.

PR #5131, opened 2021-04-23
5048 - Fix selectable setting from mixed to single to remove checkboxes

ericangeles:

This PR fixes the listview selectable setting from mixed to single to remove the checkboxes.

Closes https://github.com/infor-design/enterprise/issues/5048

- Go to http://localhost:4000/components/listview/test-changing-selectable-setting.html
- Click the button actions to change from `single` to `mixed` or `multiple`
- There should be a checkboxes appear when changing to `mixed` or `multiple`
- It should reset everything once change the selectable setting


tmcconechy (changes requested):
@ericangeles i think there is two things we could improve here...

- turn on multi select
- select 3 rows
- turn on single select -> problem is three rows are selected but only one can be...

@claudenbach is it ok if the selections are reset when you change selection mode?

Also in mixed mode when you click the checkbox it is also activating the row. It ideally should not do this. Only the checkbox should check in that case.

claudenbach:
Yes, it's okay to reset the selections when the selection mode changes to match the mode.

ericangeles:
@tmcconechy looks like it doesn't activating the row when clicking the checkbox, but I noticed it has border.

tmcconechy:
@ericangeles not sure what you mean? I think we just need to:

- reset all selections if changing this option (in updated)
- in mixed mode make sure the click of the checkbox doesnt activate the row (if possible?)

ericangeles:
I meant in mixed mode, checking the checkbox doesn't affect the row to be activated.

tmcconechy:
Ok i see - it works better here https://master-enterprise.demo.design.infor.com/components/listview/example-mixed-selection.html 
Maybe we need to set or have a showHoverState (like datagrid)

ericangeles:
@tmcconechy I added a more better test page to test this. I also updated the details of the PR above.

Test here: http://localhost:4000/components/listview/test-changing-selectable-setting.html

tmcconechy (approved):
Ok approving. I see we have a hoverable setting in list view. I think maybe in the example they could turn that on but its optional. SO this is working as expected 👍🏻

deep7102 (approved):

PR #5136, opened 2021-04-26
4520 - Add sortable drag and drop to tabs

### LINKED ISSUE #4520: Tabs: Reorder Tabs via drag and drop
Our team is currently migrating the classic program to Angular. One of the feature that is being asked now for our User certification is the tab reordering via drag and drop, but currently soho-tabs doesn't support it. Please see below for sample behavior:

expected_tab_reorder

Enable users to reorder tabs via drag and drop
Targeting Module tabs only and visual tabs

**Additional context**
And what application do you work on? M3 User Productivity

This has been reported before -> https://github.com/infor-design/enterprise/issues/1366 
We don't have any alternatives for the tab reordering and this is a must feature for the users.

deep7102:
Added support to sortable drag and drop tabs. Non touch devices it's good with almost every type of tabs `Module`, `Vertical`, `Header`, `Scrollable` and `Regular`. For touch devices only support with `Module` and `Vertical` Tabs.

Closes #4520

Pull this branch and build/run the demo app
- Navigate to: http://localhost:4000/components/tabs-module/example-sortable.html
- Rearrange tabs use drag and drop
- It should work fine

### DIFF
--- a/app/views/components/tabs-module/example-sortable.html
+++ b/app/views/components/tabs-module/example-sortable.html
@@ -1,0 +1,104 @@
+<body class="no-scroll">
+
+  <a href="#maincontent" class="skip-link" data-translate="text">SkipToMain</a>
+  {{> includes/svg-inline-refs}}
+  {{> includes/applicationmenu}}
+
+  <div class="page-container no-scroll">
+
+    <!-- Module Tabs -->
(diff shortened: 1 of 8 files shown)


tmcconechy (commented):
This looks really really good. I just have one comment and its hard to explain as i cant do a screen shot...

- drag first tab to second tab
- notice how when the second tab moves over its "dark blue" -> i think that should be the drop area. so basically all the tabs should retain their default color and the empty space should always be the dark blue

That should make it a bit easier to use... Ping me if unclear.

ericangeles (approved):

Issue #5141, opened 2021-04-27
Soho Editor Anchor link problem


junxyan:
**Describe the bug**
in the angular form, there is a soho editor along with other input controls. and save button at the end of the form.
When added a link to an existing word in the soho editor, anchor link popup dialog opened, and click insert button in that dialog , and that word in the editor was changed to a hyperlink automatically. however, onChange event of soho editor component was not triggered in this case, if save button was clicked without editing more texts in rich text editor, link change wasn't saved the database at all.

**Expected behavior**
soho editor should reflect changes link did, and trigger onchange event or other events so when saving the correct value will be saved.

**Version**
- ids-enterprise-ng: [v9.2.0]

**Platform**
 - Browser Name [chrome]

jbrcna:
QA Passed
v4.51.0-dev
https://main-enterprise.demo.design.infor.com/components/editor/example-with-events.html
![image](https://user-images.githubusercontent.com/56722184/116836141-03449f00-abf8-11eb-822f-dfe2cd4938d7.png)

PR #5148, opened 2021-04-28
5141 - Fix anchor link on firing the change event in editor

ericangeles:

This PR fixes the anchor link in editor to fire once there's any changes that the link did.

Closes https://github.com/infor-design/enterprise/issues/5141

- Go to http://localhost:4000/components/editor/example-with-events.html
- Select some text
- Click the anchor link buton in toolbar section
- A modal pops up, just click `insert`
- Notice below, the `Change Event Detected` should be color red
- Use also the hotkey for triggering the anchor link
- Select some text again
- Then hit `cmd+h` for the hotkey
- A modal pops up, just click `insert`
- Output should be the same


tmcconechy (changes requested):
@ericangeles something isnt 100% here. Try this example

http://localhost:4000/components/editor/example-dirty-tracking.html
- select text
- use the link modal and link some text
- in this case the dirty indicator is not shown on that change

So while the change event works the dirty functionality isnt yet

deep7102:
I see same as @tmcconechy comment above. Also see change event fire soon the anchor model popup opens. It should only trigger if anything change.

tmcconechy:
@ericangeles FYI i fixed a test that will likely fail for you on the other open PR

tmcconechy:
This is working now but something was introduced in these two editor tests https://github.com/infor-design/enterprise/pull/5148/checks?check_run_id=2468052118#step:11:3926

The searchfield test is the one i fixed in the other PR

EdwardCoyle (approved):
This looks good to me now 👍

tmcconechy (approved):

PR #5151, opened 2021-04-28
5149 - Add fix for non-editable lookup

### LINKED ISSUE #5149: Forms: read only Look up field issue
**Describe the bug**
you can open the lookup even if it's a read-only element

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/form/example-readonly-inputs.html
2. Click on the Lookup icon to open Lookup
5. Try to  select and change the value
6. See error
7.  A read-only element should not be editable

**Expected behavior**
should not be able to open lookup
image

**Version**

- ids-enterprise: v4.51.0-dev

**Screenshots**
image

**Platform**
all

**Additional context**

tmcconechy:

Reverted some changes in https://github.com/infor-design/enterprise/commit/b90bb8e8406baac8c842932af039991cfcbb8339#diff-3f1934dc1d84a5236f9e5386034c6900619e2e2b910caef8c0d8153dc8c8ec26 as it caused readonly lookups to appear not readonly. It was perhaps not noticed that there is a non-editable lookup and a readonly lookup

Fixes #5149

- go to http://localhost:4000/components/lookup/test-states.html
- notice there is a readonly field with right styling again
- and now a non editable lookup example (you cant type)
- recheck http://localhost:4000/components/form/example-readonly-inputs.html -> should also be readonly state again

### DIFF
--- a/app/views/components/lookup/test-states.html
+++ b/app/views/components/lookup/test-states.html
@@ -14,6 +14,11 @@
     </div>
 
     <div class="field">
+      <label for="non-editable" class="label">Not Editable</label>
+      <input id="non-editable" class="lookup" data-options="{editable: false}" name="readonly" type="text">
+    </div>
+
+    <div class="field">
       <label for="disabled" class="label">Disabled</label>
       <input id="disabled" class="lookup" name="disabled" disabled type="text">
     </div>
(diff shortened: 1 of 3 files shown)


deep7102 (approved):

ericangeles (approved):
Looks good! 🍌

PR #5169, opened 2021-05-03
4857 - Prevent single buttons from remaining in Toolbar overflow

### LINKED ISSUE #4857: Tabs Module: Resizing browser moves home button to "More" menu
**Describe the bug**
This occurs when the toolbarsearchfield setting 'showCategoryText' is set to false. When you contract and expand the browser, the home button is sent into the 'More' menu.

**To Reproduce**
1. Go to http://localhost:4000/components/tabs-module/example-category-searchfield-go-button-home.html
2. Set the 'showCategoryText' setting to false
3. Contract/Expand the browser
4. See that the home button remains in the 'More' menu

**Expected behavior**
It is expected that the home button will appear on its own when resizing.

**Version**
- ids-enterprise: v4.38.0-dev

**Screenshots**
image

**Platform**
 - Device: (if applicable) N/A
 - OS Version: Windows10
 - Browser Name: chrome
 - Browser Version: 88.0.4324.182 (Official Build) (64-bit)

EdwardCoyle:
This PR fixes a bug in the original Toolbar code (when coupled with Searchfields) that in some cases would force a single toolbar button to sit inside of the "More Actions" menu, when the button could instead just stand by itself.  The order of the overflow calculation has been adjusted to prevent this issue.

Closes #4857 

- Pull/build/run
- Open http://localhost:4000/components/tabs-module/example-category-searchfield-go-button-home.html?layout=nofrills
- Open developer tools and toggle the device toolbar.  Set the mode to "responsive"
- Shrink the size of the window to be incredibly small, width-wise.
- Resize the width back to a "normal" size.
- See that the icon in the top-right section of the Toolbar should now be a "home" button instead of a "More Actions" button containing the Home button.

Screen Shot 2021-05-03 at 12 14 18 PM

### DIFF
--- a/app/views/components/tabs-module/example-category-searchfield-go-button-home.html
+++ b/app/views/components/tabs-module/example-category-searchfield-go-button-home.html
@@ -71,8 +71,7 @@
       'goButtonAction': goButtonAction,
       'showGoButton': true,
       'categories': [ 'Books', 'Movies', 'Music', 'Video Games' ],
-      'showCategoryText': true,
+      'showCategoryText': false,
       'clearable': true
     });
-
   </script>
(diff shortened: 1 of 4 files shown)


tmcconechy:
How small? Asking because at a certain point i see it again overflowing. But maybe this is enough?

Screen Shot 2021-05-03 at 1 00 54 PM

EdwardCoyle:
@tmcconechy when you get down really small (to the point where text is getting cut off, etc), it might still collapse.  The real problem from #4857 was that it would never "go back" and always be stuck in overflow, even if there was room to grow out.  This at least fixes that problem.

Overall I'd recommend to @s-werking @awbuboltz -- wherever the application is using a standard Toolbar, it should be upgraded over time to use our [Flex Toolbar](https://design.infor.com/code/ids-enterprise/latest/toolbar-flex), which is CSS-driven and solves all of these kinds of collapse/resize issues.  Our experience is that the Javascript-based resizing method doesn't work all that well, considering all the different cases the Toolbar has to support, which is why we built the new one.

tmcconechy (approved):
OK as long as we think it solves the issue

deep7102 (approved):

Issue #4503, opened 2020-10-19
Lookup / Datagrid: Unselect All issue across pages (server side paging)


volante007:
**Describe the bug**
If unselecting all items on a particular page in a lookup / datagrid, it will unselect all items on other pages as well.

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/lookup/example-multiselect-paging-serverside?theme=uplift&variant=light
2. Click on browse button
3. Tick top checkbox to Select all on first page
4. Go to page 2
5. Tick top checkbox to Select all on second page
6. Tick top checkbox  to Unselect all on second page
7. Go Back to page 1
Result: All items unselected on page 1

Can also be reproduced for a server side paged datagrid
https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/example-paging?theme=uplift&variant=light

**Expected behavior**
Items on page 1 should still be selected. Select all only applies for the active page, so Unselect all should also only apply for the active page.

**Version**
- ids-enterprise: 4.33.0

**Screenshots**
![lookup](https://user-images.githubusercontent.com/1101770/96451912-90a24200-1218-11eb-9f23-c064b8cde0a2.gif)

tmcconechy:
Yes seems wrong since select all is applying to one page. This may be related to https://github.com/infor-design/enterprise/issues/4265

tmcconechy:
How long would you estimate for this @ericangeles ? Trying to determine if it makes the release and was hoping to do that Monday FYI

ericangeles:
@tmcconechy trying to fix it EOD or by Monday. But I'll let you know later offline.

tmcconechy:
Moving this to next sprint as it didnt quite make it. But please continue with the fix when you can.

jbrcna:
QA Passed
v4.52.0-dev
https://main-enterprise.demo.design.infor.com/components/lookup/example-multiselect-paging-serverside
![image](https://user-images.githubusercontent.com/56722184/117614980-af0f6100-b19b-11eb-83fc-652e50109533.png)

PR #5174, opened 2021-05-04
4503 - Fix unselected items on active page and display counts

ericangeles:
 
This PR fixes a bug where unselecting all items in an active page affects other selected items on other pages as well.

There's a line of code that resets the `this._selectedRows` after mapping it.

I also noticed that the selected counts don't reflect immediately. Fixed it too.

Closes https://github.com/infor-design/enterprise/issues/4503

- Go to http://localhost:4000/components/lookup/example-multiselect-paging-serverside
- Click the lookup button
- Tick top checkbox to Select all on first page
- Move to Page 2
- Tick top checkbox to Select all on second page
- Tick top checkbox to Unselect all on second page
- It should back to `5 Selected` and not remove the toolbar
- Go Back to page 1
- It should not reset or unselect all the items on Page 1

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7374,6 +7374,10 @@ Datagrid.prototype = {
     // Sync the Ui and call the events
     this.dontSyncUi = false;
 
+    // It should not clear the selectedRows in lookup
+    const isLookup = $('.lookup-modal.is-visible');
+    this._selectedRows = !isLookup.length ? [] : this._selectedRows;
+
     // Update the display counts when unselecting all rows
     this.displayCounts();
 
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7374,10 +7374,6 @@ Datagrid.prototype = {
     // Sync the Ui and call the events
     this.dontSyncUi = false;
 
-    // It should not clear the selectedRows in lookup
-    const isLookup = $('.lookup');
-    this._selectedRows = !isLookup.length ? [] : this._selectedRows;
-
     // Update the display counts when unselecting all rows
     this.displayCounts();
 
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -7373,7 +7373,9 @@ Datagrid.prototype = {
     this.settings.dataset.map((row) => { delete row._selected; }); //eslint-disable-line
     // Sync the Ui and call the events
     this.dontSyncUi = false;
-    this._selectedRows = [];
+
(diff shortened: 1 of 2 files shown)


tmcconechy (dismissed):
Great fix 👍🏻

src/components/datagrid/datagrid.js
```diff
@@ -7373,7 +7373,13 @@ Datagrid.prototype = {
     this.settings.dataset.map((row) => { delete row._selected; }); //eslint-disable-line
     // Sync the Ui and call the events
     this.dontSyncUi = false;
-    this._selectedRows = [];
+
+    // It should not clear the selectedRows in lookup
+    const isLookup = $('.lookup');
```
deep7102, line 7378: The fix looked nice but it might mess up if multiple instances on page, not sure what we trying to target why it needs to do that 100%, can see this to find if the grid is in a lookup or any other way to find

https://github.com/infor-design/enterprise/blob/main/src/components/datagrid/datagrid.js#L6681-L6687

deep7102 (changes requested):

ericangeles:
@tmcconechy @deep7102, I reverted back my initial fix, and tested again the failing e2e and it seems to work now on my local machine. Will wait for travis if it's the same case. But if you have suggestion about this fix, feel free to add.

ericangeles:
@tmcconechy @deep7102, I reverted back my initial fix, and tested again the failing e2e and it seems to work now on my local machine. Will wait for travis if it's the same case. But if you have suggestions to fix this, feel free to add.

src/components/datagrid/datagrid.js
ericangeles: Thanks, @deep7102 ! I specified the lookup modal when open instead of if there's a lookup instance only. :)

src/components/datagrid/datagrid.js
```diff
@@ -7373,7 +7373,13 @@ Datagrid.prototype = {
     this.settings.dataset.map((row) => { delete row._selected; }); //eslint-disable-line
     // Sync the Ui and call the events
     this.dontSyncUi = false;
-    this._selectedRows = [];
+
+    // It should not clear the selectedRows in lookup
+    const isLookup = $('.lookup-modal.is-visible');
```
tmcconechy, line 7378: Another case this could go wrong maybe is if a modal is left in the page unopened or there is a nested modal.
Maybe you could try this instead...

```
const isLookup = (this.element.closest('.lookup-modal').length === 1);
````

tmcconechy (approved):

deep7102 (approved):
Looks good 🍌 
Thanks! @ericangeles

Issue #4825, opened 2021-02-08
Datagrid: Cells with a leading space have a dirty indicator on second blur without changes


s-werking:
**Describe the bug**
When a cell has a purposeful leading space, the space is replaced with '&nbsp', triggering a dirty indicator after the cell is clicked twice without changing the value.

**To Reproduce**
1. Go to http://localhost:4000/components/datagrid/example-editable.html
2. Add the "showDirty: true" setting to the datagrid and set the quantity of the first row to "  1".
3. Click on that cell and blur it, and repeat this step.
4. Notice that the dirty indicator appears without ever changing the cell value.

**Expected behavior**
Since our use of the datagrid allows the use of leading spaces, it is expected that these spaces are preserved and that without changing the value of the cell the dirty indicator should not appear.

**Version**

- ids-enterprise: v4.37.0-dev

**Platform**
 - Infor Application/Team Name: lawson
 - Device: (if applicable) N/A
 - OS Version: Windows10
 - Browser Name: chrome
 - Browser Version: 88.0.4324.146 (Official Build) (64-bit)

janahintal:
This is now working as expected.
Verified in https://main-enterprise.demo.design.infor.com/components/datagrid/test-cell-with-leading-space-dirty-indicator.html

PR #5184, opened 2021-05-06
4825 - Fix cell value to not trigger dirty indicator with a leading space

ericangeles:

This PR fixes where the cell/s with a leading space triggers the dirty indicator even without changing the cell value on second blur/selection.

For conditional purpose, I use `trim` method to remove the white space. Trim method doesn't mutate the original state, safe to say that the display will look as it is.

Closes https://github.com/infor-design/enterprise/issues/4825

- Go to http://localhost:4000/components/datagrid/test-cell-with-leading-space-dirty-indicator.html
- Click the first row quantity ` 1` to trigger editing
- Click outside the current cell
- Then do it twice or repetitively 
- Dirty indicator should not show

### DIFF
--- a/app/views/components/datagrid/test-cell-with-leading-space-dirty-indicator.html
+++ b/app/views/components/datagrid/test-cell-with-leading-space-dirty-indicator.html
@@ -1,0 +1,191 @@
+
+<div class="row">
+    <div class="twelve columns">
+      <div role="toolbar" class="toolbar">
+  
+        <div class="title">
+          Data Grid Header Title
+          <span class="datagrid-result-count">(N Results)</span>
+        </div>
+  
+        <div class="buttonset">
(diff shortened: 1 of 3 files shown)


tmcconechy (commented):
The fix itself looks good but the tests need a fix https://github.com/infor-design/enterprise/pull/5184/checks?check_run_id=2518664738#step:10:1247 as seems like an error in there...

tmcconechy (changes requested):
Just fix the test then looks good 👍🏻

tmcconechy:
Still seems to have some new errors https://github.com/infor-design/enterprise/pull/5184/checks?check_run_id=2526296445#step:10:1247

tmcconechy (approved):

Issue #5239, opened 2021-05-27
Datagrid: Clear filter icon for a lookup is rendered to high when using the new theme


claudenbach:
Describe the bug:
The clear filter icon for a lookup is rendered to high in the filter row when using the new theme.  Placement of the icon is correct when using the classic theme.

1. Go to 'https://master-enterprise.demo.design.infor.com/components/datagrid/test-lookup-filter'
2. Ensure the theme is set to New
3. Enter a value in the Product Id filter row
4. Notice that the clear icon is rendered to high

Expected behavior:
Icons in the filter row for a lookup are aligned.

Version
- ids-enterprise: v4.51.0-dev

claudenbach:
![image](https://user-images.githubusercontent.com/39804038/119888128-9ec8e700-befa-11eb-8024-58852970d7e3.png)

tmcconechy:
@claudenbach i dont think this a bug anymore we renamed master to main so the latest link is now https://main-enterprise.demo.design.infor.com/components/datagrid/test-lookup-filter

Master is now from some older point in time

claudenbach:
Okay.  So Landmark is seeing the icon rendered to high.  Is there something that we need to change then?  We recently moved to 4.51.


![image](https://user-images.githubusercontent.com/39804038/119888730-5b22ad00-befb-11eb-8d8a-5a77831318f4.png)

claudenbach:
On the main-enterprise.demo.design.info, I'm seeing the following when using classic theme...

![image](https://user-images.githubusercontent.com/39804038/119889159-dbe1a900-befb-11eb-979b-7afc5b000fdb.png)

tmcconechy:
I see the issue on classic theme but i cant say what the issue is on your side? But i can address that.

claudenbach:
If I change the row height to extra-small, I'm seeing the issue on main with the new theme.

![image](https://user-images.githubusercontent.com/39804038/119889650-76da8300-befc-11eb-813f-3262943682ce.png)

claudenbach:
See the issue when the row height is small or extra-small and use the new theme.

tmcconechy:
Adding this one to the current sprint as well. Can you have a look @ericangeles ?

ericangeles:
@tmcconechy will do.

janahintal:
Failing in https://4520-beta0-enterprise.demo.design.infor.com/components/datagrid/test-lookup-filter

[...]

PR #5241, opened 2021-05-28
5239 - Fix the position of clear filter icon when changing the row height

ericangeles:

This PR fixes the position of clear filter icon when chaning the row height and also when doing filtering for both themes.

Closes https://github.com/infor-design/enterprise/issues/5239

- Go to http://localhost:4000/components/datagrid/test-lookup-filter
- Then open the lookup on `Product Id`
- Select any Product to trigger the clear filter icon
- Then change the row height to `Extra Small`
- The icon should properly centered in lookup field
- Test this to different row heights
- Then try to filter/sort the `Product Id` column
- It should stay to its position
- Test on both themes then repeat the steps


tmcconechy:
Seeing one wierd thing about this @ericangeles 
- when you change between the small/med/normal row heights
- the filter value is lost in the field and the x goes away

Is that a new issue?

ericangeles:
Hmm, that's one of the fix in this issue. Could you tell me how you reproduce it, @tmcconechy?

tmcconechy:
- go to http://localhost:4000/components/datagrid/test-lookup-filter
- notice it has an X and text in the lookup column
- in the ... menu pick extra small (or any other row)
- notice it has missing X and text now and the grid is also filtered

Seems like it did this before looking on a branch but maybe we can check why the value disappears as it is out sync

ericangeles:
Will take a look at it.

deep7102:
I see same as @tmcconechy all other looks fine!

ericangeles:
I see the issue now, @tmcconechy. That is another issue, but we can try to fix that too.

tmcconechy:
Maybe not important right now then if it was there

Issue #5202, opened 2021-05-18
Lookup: attributes are not added to Save and Cancel button


SofiK:
**Describe the bug**
when I add attributes to a lookup it is added to the input, trigger, modal title and searchfield. But not to the Cancel and Save buttons. They still only have id="modal-button-1". It would be useful if they to got the added attributes!

**To Reproduce**

1. Go to  https://main-enterprise.demo.design.infor.com/components/lookup/example-index.html
2. Click on the first lookup so the dialog is opened
3. Open dev tools
4. Inspect the cancel button

**Expected behavior**
The cancel button has the attribute data-automation-id with value "my-lookup-modal-cancel"
and the save button has the attribute data-automation-id with value "my-lookup-modal-save"

**Version**

- ids-enterprise: 4.51

**Screenshots**
![image](https://user-images.githubusercontent.com/8664975/118662123-0e025500-b7f0-11eb-87e8-8a16c401dae0.png)

**Platform**
 - Infor Application/Team Name: Homepages
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 90.0.4430.212 (Official Build) (64-bit)

**Additional context**
-

janahintal:
QA Passed. Thank you. 
![image](https://user-images.githubusercontent.com/45757016/122028944-1190f800-cdff-11eb-86e7-1cfb3698e7ca.png)

PR #5264, opened 2021-06-03
5202 - Lookup attributes are not added in the cancel and apply/save button

tjamesallen15:

This PR fixes where lookup attributes are not added in the cancel and apply/save button.

Closes https://github.com/infor-design/enterprise/issues/5202

- Go to http://localhost:4000/components/lookup/example-index.html
- Click on the first lookup so the dialog is opened
- Open dev tools
- Inspect the cancel button


tmcconechy (dismissed):
Fix looks good. I also am trying to fix the test suite so i put that here...

deep7102 (approved):

src/components/modal/modal.js
```diff
@@ -916,6 +916,17 @@ Modal.prototype = {
...
 
+      const modalButtonSet = this.element.find('.modal-buttonset');
+      const modalButtonList = modalButtonSet.find('.btn-modal');
+      const self = this;
+      modalButtonList.each((index, item) => {
+        const btnItem = $(item);
+        if (btnItem.hasClass('btn-modal-primary')) {
+          utils.addAttributes(btnItem, self, self.settings.attributes, 'save', true);
+        } else {
+          utils.addAttributes(btnItem, self, self.settings.attributes, 'cancel', true);
```
EdwardCoyle, line 927: @tjamesallen15 This has an unexpected effect on Modals that have non-primary buttons that aren't necessarily labelled "close" or "cancel".  We didn't really have a good test for this setup anywhere, so I pushed one to this branch: http://localhost:4000/components/modal/test-inline-buttons.html.  If you open the modal on that page and inspect the first button (non-primary) you can see it labelled as a cancel button.

I was trying to think of a more generic way we can handle this, and realized that the actual ID for all of these Modal buttons is generated here: https://github.com/infor-design/enterprise/blob/main/src/components/modal/modal.js#L533 -- there is a section beneath that attempts to add a close handler for cancel buttons.  It might be better to remove the fix from here and rework it into that logic (you'd be able to accurately identify cancel buttons in this spot).

EdwardCoyle (changes requested):

src/components/modal/modal.js
tjamesallen15: Hello @EdwardCoyle , will check this one. Thanks for this one!

tjamesallen15: @EdwardCoyle I saw the code there. We can definitely put the code there. Although the example-index in the lookup don't have a class "btn-cancel" for the Cancel button.

src/components/modal/modal.js
```diff
@@ -569,7 +569,19 @@ Modal.prototype = {
           btn.element[0].classList.add('no-validation');
         }
 
-        utils.addAttributes(btn.element, this, settingsJSON.attributes, '', true);
+        if (settingsJSON.attributes) {
+          // If Buttons have its own attribute settings
+          utils.addAttributes(btn.element, this, settingsJSON.attributes, '', this);
+        } else if (btn.element.hasClass('btn-modal-primary')) {
+          // Buttons tag as btn-primary will be tag as save
+          utils.addAttributes(btn.element, this, self.settings.attributes, 'save', true);
```
tmcconechy, line 577: I still think we cant assume these are all save buttons. So for an ID name i would suggest something more generic like `primary` for primary. I think we can assume `btn-cancel` is ok as cancel.

src/components/modal/modal.js
```diff
@@ -572,10 +572,11 @@ Modal.prototype = {
         if (settingsJSON.attributes) {
           // If Buttons have its own attribute settings
           utils.addAttributes(btn.element, this, settingsJSON.attributes, '', this);
-        } else if (btn.element.hasClass('btn-modal-primary')) {
+        } else if (btn.element.is('[class*=primary]')) {
+          console.log('hotdog');
```
tmcconechy, line 576: We can remove this...

src/components/modal/modal.js
```diff
@@ -572,10 +572,11 @@ Modal.prototype = {
...
           // Buttons tag as btn-primary will be tag as save
           utils.addAttributes(btn.element, this, self.settings.attributes, 'save', true);
```
tmcconechy, line 578: Suggest:
```
utils.addAttributes(btn.element, this, self.settings.attributes, 'primary', true);
```

tmcconechy (approved):
Works for me for now. Thanks @tjamesallen15

tjamesallen15:
> Works for me for now. Thanks @tjamesallen15

Welcome @tmcconechy , anytime :)

tmcconechy (approved):

ericangeles (approved):

EdwardCoyle (approved):

PR #5293, opened 2021-06-09
5219 - Fixed Bug Issue for Highlighted Text in New Light Default Theme Header

### LINKED ISSUE #5219: Header: default color makes make it impossible to highlight the selected a text in the header
It looks like the default color of the light-new theme is identical to the browser text selection color (I tried Edge and Chrome).
It makes it look like the text is non-selectable and very confusing; please see the gif
header_selection_default_color
I think it is the only header that is affected

Expected: for this particular default color, apply custom text selection color for _.header_

RiyaTayal20:

This PR fixes the contrast of the highlighted text within the header when using the new light default theme.

This PR also fixes the below bug that had the top and bottom sections of the header in two different colors
Screen Shot 2021-06-18 at 11 34 27 AM

Closes #5219 

- Pull this branch down, then build and run the app
- Go to http://localhost:4000/components/personalize and select an example form
- Check if the header looks like the following when highlighting using the new light default theme:
Untitled
- Check if the top and bottom sections of the header are the same color when using the new light theme and any color like the following image in azure: 
  Screen Shot 2021-06-18 at 11 38 40 AM


tmcconechy (dismissed):
Looks great to me

src/components/input/_input.scss
```diff
@@ -496,13 +496,11 @@ span.required::after {
 
 // Text Selection
 ::selection {
-  background: $ids-color-brand-primary-base;
-  color: $inverse-color;
+  background: $ids-color-palette-azure-90;
```
EdwardCoyle, line 499: This color change is affecting the whole document, including some opposite-color text in the page body.  You may want to try keeping the original color overall, and scoping selections of this color strictly to `.header` components:

Screen Shot 2021-06-11 at 9 36 52 AM

EdwardCoyle (changes requested):
@RiyaTayal20 found a small thing below:

src/components/input/_input.scss
tmcconechy: good catch -> i think it should be constrained under is-personalize class so the change should be here https://github.com/infor-design/enterprise/blob/main/src/components/personalize/personalize.styles.js#L10

Glad you caught that because i missed it @EdwardCoyle

tmcconechy:
Also this may be why the visual tests failed as well. So see if you can fix @EdwardCoyle 's comment and then maybe you dont need to do the images

src/components/personalize/personalize.styles.js
```diff
@@ -7,7 +7,11 @@ function personalizeStyles(colors) {
   return `
 
 .is-personalizable ::selection {
-  background: ${colors.selection} !important
+  background: ${colors.selection} !important;
```
tmcconechy, line 10:
```suggestion
  background: ${colors.darker} !important;
```

Not sure if this will work better if you just change it for all elements inside a is-personalizable area?

But what is wierd is i dont see it working at all on this page?
http://localhost:4000/components/personalize/example-form.html

EdwardCoyle: @tmcconechy @RiyaTayal20 I think we're still missing a CSS rule for `::selection` to be applied to headers without personalization rules on this PR

EdwardCoyle: Probably need one here somewhere: https://github.com/infor-design/enterprise/blob/main/src/components/header/_header.scss#L15

RiyaTayal20: Oh this is what I pushed on Friday so you would be able to see what was going on @EdwardCoyle this issue hasn't been finalized yet.

tmcconechy:
ok @RiyaTayal20 we have a label for WIP we usually put on and then people won't reveiw your PR then we take it off when its ready. Addit for you

tmcconechy:
@RiyaTayal20 is this ready to review again or not yet?

EdwardCoyle (approved):
👍  Nice fix, great job @RiyaTayal20!

tmcconechy (approved):

Issue #5250, opened 2021-06-01
Cards: Add content variants and Improve Hitboxs


tmcconechy:
**Is your feature request related to a problem or use case? Please describe.**

We want to add some examples to cards for a WFM demo and also improve the hit boxes.

**Describe the solution you'd like**

 - Add a status card with an actions menu (to card examples - can start with https://main-enterprise.demo.design.infor.com/components/cards/example-auto-size.html)
 - Add an example with busy status
 - Add an example with hyperlink
 - Add an example with photo card
 - Improve the hit box size for tapping
 
**Design**
![Screen Shot 2021-06-01 at 11 38 26 AM](https://user-images.githubusercontent.com/814283/120351535-f89c2900-c2cd-11eb-98c4-49f6044d2f98.png)

**Additional context**
See these sections on the figma sketch if needed https://www.figma.com/file/ri2Knf3KchdfdzRAeds0Ab/IDS-Mobility-v4.6?node-id=1%3A5740

And also this section:
![Screen Shot 2021-06-01 at 12 37 22 PM](https://user-images.githubusercontent.com/814283/120362932-9b0dd980-c2d9-11eb-98e8-e379765d0dba.png)

janahintal:
All examples were working as expected. However, no single action section found in the example page. 
![image](https://user-images.githubusercontent.com/45757016/122025719-2c15a200-cdfc-11eb-9b2d-0267ebd9b466.png)

tjamesallen15:
> All examples were working as expected. However, no single action section found in the example page.
> ![image](https://user-images.githubusercontent.com/45757016/122025719-2c15a200-cdfc-11eb-9b2d-0267ebd9b466.png)

Hello. Will work on this one tonight.

CindyMercadoReyes:
BDD added to this new enhancement. 
[story_github5250.docx](https://github.com/infor-design/enterprise/files/6656192/story_github5250.docx)

janahintal:
Hi @tmcconechy,  I've notice a difference in tab focus of '...' button when in Classic Theme. The focus in New version looks centered but not in classic version. 

New Version
![image](https://user-images.githubusercontent.com/45757016/122553729-a2b7d700-d06a-11eb-879b-35f63f2700d3.png)

Classic Version
![image](https://user-images.githubusercontent.com/45757016/122553774-b2cfb680-d06a-11eb-90d6-59f44ef82e87.png)

tmcconechy:
[...]

PR #5316, opened 2021-06-15
5250 - Update on Card Variations (Single Action)

tjamesallen15:

This PR is a follow update on Card Variations and Hitboxes

Closes https://github.com/infor-design/enterprise/issues/5250

- Go to http://localhost:4000/components/cards/example-variations-hitboxes.html
- Check the cards


tmcconechy (changes requested):
Just noticed something i did not see before.

- try tabbing the page
- notice the ... buttons do not have tab stops? they should
- also the "hit box" on the call icon button is backed right up to the edge. should be a bit of space to the end so that looks better

Screen Shot 2021-06-15 at 9 15 28 AM

ericangeles:
To adds @tmcconechy mentioned, also fix,

- the alignment of the button when focused (tabbed) e.g. screenshot above's Tim comment
- The spacing of link when focused (too much i think, and it's too close to the text above)

focus

tmcconechy:
regarding the focus on the links that was the current state https://main-enterprise.demo.design.infor.com/components/hyperlinks/example-index.html but we could reduce the top padding a little if that helps?

tmcconechy (approved):

tmcconechy:
The only thing im wondering @EdwardCoyle ? Is if this should be a change to all buttons. But i guess that would be too problematic so ok with this for now

ericangeles (approved):

PR #5341, opened 2021-06-18
NG764/5197 - Separate Popover/Tooltip APIs

### LINKED ISSUE #5197: Popup/Tooltip: Using Popover with a tooltip on the same object (button) causes conflicts
**Describe the bug**
When replacing the deprecated **soho-popdown** with a **soho-tooltip [popover]="true"**, the opening element's "**title**" attribute becomes the header of the **popover** and it no longer shows on hovering that element as a _tool-tip_.

This means that an element cannot have a tool-tip independent from its popover; which could be done before.

soho_popdown_vs_popover

**To Reproduce**
This can be reproduced in the demo by simply trying to use a **soho-tooltip** as a popover, for example:
```html
<button soho-button="icon" title="This is the button title..." soho-tooltip [popover]="true" ...>
```

**Expected behavior**
For popovers, you may want to have a different "tool-tip" for the opening button explaining what it does ("hint") and display a different popover header or **no header at all**.

We can reproduce the desired results if we do something like this:
```html
<button soho-button="icon" [attr.title]="'This is the button title...'" title="xxx" soho-tooltip [popover]="true" ...>
```
soho_tooltip_popover

However, this might be _hacky_, and it doesn't allow you to have a soho formatted tooltip (in case you want it).

**Version**
- ids-enterprise-ng: 7.0.0-dev.20200219

**Platform**
 - OS Version: Windows 
 - Browser Name: Chrome
 - Browser Version: Latest

EdwardCoyle:
Currently in `main` it's impossible for a Tooltip and a Popover to share the same trigger element.  The API for both components is the same, and was previously using the same key when stored against `$.data()`.  This PR makes an attempt to "separate" these two APIs in a way that they can share the same code, but be stored separately so they can co-exist on one element.

This PR enables the following use cases:
- A popover and a tooltip that share the same triggering element (#5197)
- A popover containing a nested tooltip (infor-design/enterprise-ng#764)

- Closes #5197 
- Related to infor-design/enterprise-ng#764

Pull, build, run!  Then try the following cases:

_Popover/Tooltip don't close each other:_
- Open http://localhost:4000/components/popover/test-popover-tooltip.html
- Open the popover
- Hover the "Buttons with Tooltips" button.  When the tooltip appears, the Popover shouldn't close.

_Popovers with Nested Tooltips:_
- Open http://localhost:4000/components/popover/test-popover-with-nested-tooltip
- Open the popover
- Hover the "Apply" button. When the tooltip appears, the Popover shouldn't close.

_Popovers and Tooltips sharing the same trigger element:_
- Open http://localhost:4000/components/popover/test-popover-tooltip-same-button.html
- open the Popover
- Hover the "Click to Show Popover" button.  Both the Tooltip and the Popover should be displayed at the same time.

Also smoke test these components, which use Popover internally and ensure their functionality is the same:
- http://localhost:4000/components/calendar
- http://localhost:4000/components/datepicker
- http://localhost:4000/components/timepicker
- http://localhost:4000/components/modal/test-close-tooltip.html (test that the modal closes when a tooltip is showing)
- http://localhost:4000/components/modal/test-popover.html (test that the popover can open/close)


tmcconechy (approved):
Looks good. I fixed an issue where the browser tooltip will also show.

deep7102 (approved):
Nice! 🍌

PR #5356, opened 2021-06-23
NG#1046 - Fix unwanted navigation/link after using destroy method

ericangeles:

This PR fixes unwanted navigation/link after using the destroy method on popupmenu. Click event should call preventDefault method to stop the browser from setting the hash.

Closes https://github.com/infor-design/enterprise-ng/issues/1046

- Go to http://localhost:4000/components/popupmenu/test-unwanted-navigation-when-menu-destroyed.html
- Click `Normal Menu` button
- Click the first menu item `Menu Option #1`
- Click again the Popupmenu button
- It should destroy it and should not open
- Hash should not be appended in URL


tmcconechy (changes requested):
Can this go at the click event on line 1021? That way it will get detached in the detach event.
 If not can you add a namespace and turn it off in the detatch event?

tmcconechy (commented):
To be safe, I think just need to add `click.btn-menu.${this.id}` to:

 https://github.com/infor-design/enterprise/blob/049220606043b978fe2eb694fd3ae1ee4b80367b/src/components/popupmenu/popupmenu.js#L2334

tmcconechy (approved):

EdwardCoyle (approved):

Issue #5179, opened 2021-05-05
Home and End Key behavior when in edit mode in editable cell


marclouisgenedeguzman:
https://master-enterprise.demo.design.infor.com/components/datagrid/example-editable.html

When a user focuses on the editable cell, and press Home key or End Key this is what happens:
Press Home key - editable field cell focus is removed and focus will be placed on the first column of the row
Press End key - editable field cell focus is removed and focus will be placed on the last column of the row

Add a behavior in editable cell under edit mode:
Press Home key - the text cursor will only be placed in front of the text value
Press End key - place the text cursor at the end of the text value

tmcconechy:
To elaborate. The Current home/end key functionality is correct. But if in edit mode it should be ignored so it works in the field to use home/end.

CindyMercadoReyes:
This issue is now resolved.

PR #5362, opened 2021-06-24
5179 - Fix home and end key behavior in editable setting

ericangeles:

This PR fixes the home and end keys behavior that should behave as is (by default) when the `quickEditable` set to true. Otherwise, the home key should shift to the first row, and the end key should shift to the end of the row.

Closes https://github.com/infor-design/enterprise/issues/5179

- Go to http://localhost:4000/components/datagrid/example-editable.html
- Click on any cell that is editable
- Then hit `Home` key
- The cursor should be shifted/placed in front  of the cell value
- Then click again to another cell that is editable
- Hit `End` key
- The cursor should be shifted/placed at the end of the cell value


tmcconechy (changes requested):
@ericangeles i think we need a different fix because this seems to take away the home/end functionality entirely.

What was wanted was:
- home / end key (fn+right/left arrow on mac) - works all the time except when in edit mode
- in edit mode it works in the field

So would expect the fix to have some logic about editing maybe something like the cell class can be used to tell `cellParent.hasClass('is-editing'`

tmcconechy (approved):

deep7102 (approved):
🍌

Issue #5350, opened 2021-06-23
Monthview: Unnecessary scroll bar in Monthview calendar


janahintal:
**Describe the bug**
Monthview: Unnecessary scroll bar in Monthview calendar

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/monthview/example-index.html
2. Notice the monthview has unnecessary scroll bar

**Expected behavior**
No unnecessary scroll bar

**Version**
ids - 4.53 - dev

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/123055915-f3e41400-d438-11eb-820a-61023a718dbb.png)

**Platform**
All major browser, OS and devices

**Additional context**
N/a

janahintal:
QA Passed. Thank you!
![image](https://user-images.githubusercontent.com/45757016/126606979-e94f95b0-dad3-4719-827e-60842410e55a.png)
![image](https://user-images.githubusercontent.com/45757016/126606953-cf712b2f-17dc-45ae-8ceb-2ccc6bdf4f5e.png)

PR #5393, opened 2021-07-06
5350 - Fix Unnecessary Vertical Scroll

tjamesallen15:

This PR fixes the vertical scroll showing when it is really unnecessary.

Closes https://github.com/infor-design/enterprise/issues/5350

- Go to http://localhost:4000/components/monthview/example-index.html
- Check the vertical scroll if it now hidden


tmcconechy (changes requested):
@tjamesallen15 there is a trick to this. it does need to scroll in some cases for example try http://localhost:4000/components/calendar/example-index.html and this should be possible:

Screen Shot 2021-07-06 at 9 25 18 AM

So i think the fix needs to be something where it fixes it just when its at full size and still showing (why does it show for 1/2 extra pixels)

tjamesallen15:
@tmcconechy Not much sure if I done it correctly. I will wait for you guys to review this one tonight on your time. Open for suggestions.

tmcconechy:
@tjamesallen15 I think i found a better /  safe fix. Can you try to add `overflow: hidden` to this rule and see if that is a better fix?

Screen Shot 2021-07-08 at 10 10 15 AM

tjamesallen15:
Hello @tmcconechy , I tried your suggestion. It looks much better than my fix.

tmcconechy (dismissed):

tmcconechy (approved):

ericangeles (approved):

PR #5404, opened 2021-07-08
986 - Added new events to lookup

tmcconechy:

Exposed the `afterpaging` and `selected` event to solve for a use case where the lookup is showing non unique information in the field.

TODO: Make these available in NG - and update the stackblitz

Fixes 

- go to http://localhost:4000/components/lookup/test-multiselect-description-only.html
- open the lookup and go to page 2
- select any 2 rows and hit apply
- reopen the lookup and without going to the 2second page hit apply
- same values should be shown 
- try more scenarios with selecting / reopening and paging


tmcconechy:
@deep7102 i think i fixed the issue you mentioned.

deep7102 (approved):
Looks good 🍌 
Thanks!!!

ericangeles (approved):

PR #5406, opened 2021-07-09
5353 - Datepicker: Fix on -/+ key detection

### LINKED ISSUE #5353: Datepicker: keyboard +/- doesn't change the date
**Describe the bug**
Previously, pressing the + or - key on the keyboard would increment or decrement the date. This was missed in the move to 4.0.

**To Reproduce**
1. Go to /components/datepicker/example-index.html
2. In the first date field, press t to enter todays date
3. While in the date field, press the + key
4. The date should be incremented
5. While in the date field, press the - key
6. The date should be decremented

**Expected behavior**
The date should be incremented when + is pressed, or decremented when - is pressed

**Version**

- ids-enterprise: v4.51.0-dev

**Platform**
 - Infor Application/Team Name: Infor Landmark
 - All

yohannahbautista:

- Added key values of "-" and "+" in numpad.
- Removed shift key check so that there's no need to do shift and "+" to advance the date
- Datepicker updates date if input field is not empty and the value is valid. Added this because date did not update in e2e test at line 89 when assigning a value in the input field.

Closes #5353 

- Go to: http://localhost:4000/components/datepicker/example-index.html
- In the first date field, press "t" to enter todays date
- While in the date field, press the "+" key
- The date should be incremented
- While in the date field, press the "-" key
- The date should be decremented

- [ ] An e2e test for the bug.


tmcconechy (approved):
Nice fix 👍🏻

ericangeles (approved):
🍌

Issue #5253, opened 2021-06-01
Cards: Add selection state


tmcconechy:
**Is your feature request related to a problem or use case? Please describe.**
Working on features for the WFM demo. This one adds selection to cards similar to https://main-enterprise.demo.design.infor.com/components/blockgrid/example-multiselect.html

**Describe the solution you'd like**
- [x] Add selectable option to cards like listview or datagrid or block grid has
- [x] Add an example
- [ ] Add NG example and types 
- [ ] Add to WC (eventually)

**Design**
![Screen Shot 2021-06-01 at 1 32 17 PM](https://user-images.githubusercontent.com/814283/120366670-f0e48080-c2dd-11eb-8f44-f518f6b33bd6.png)

**Additional context**
See Single Select / Multi-select section in:
https://www.figma.com/file/ri2Knf3KchdfdzRAeds0Ab/IDS-Mobility-v4.6?node-id=1%3A5740

janahintal:
This is working as expected in https://main-enterprise.demo.design.infor.com/components/cards/example-multi-select.html
& https://main-enterprise.demo.design.infor.com/components/cards/example-single-select.html

![image](https://user-images.githubusercontent.com/45757016/127640157-9f6e9b4b-14a3-47b9-a8d5-ce3952236e65.png)
![image](https://user-images.githubusercontent.com/45757016/127640173-fb7f607b-9602-488a-a39f-28bdc5f7581c.png)

PR #5434, opened 2021-07-15
5253 - Add Cards' single and multiple selection state

ericangeles:

This PR implements single and multi-selection of cards.

Closes https://github.com/infor-design/enterprise/issues/5253

1.) Single Selection
- Go to http://localhost:4000/components/cards/example-single-select.html
- Click any of the cards
- It should be selectable and highlighted
- It should only select single card
- Test the UI across all themes and modes

2.) Multi-Selection
- Go to http://localhost:4000/components/cards/example-multi-select.html
- Compare to a single selection, multi-select should have checkboxes
- Click the cards, it should be selectable and the checkbox should be checked
- Clicking either the checkbox or the card itself should work the same
- Unselect the currently selected card/s
- It should unchecked and unhighlighted the card/s
- Test the UI across all themes and modes


tmcconechy (changes requested):
Looks great @ericangeles 👍🏻  Only one small thing:

- In both single and multiselect when you select the border comes on and expands the size. I think to solve this just set the default border to `border: 2px solid transparent;` so it does "jump"

ericangeles:
@tmcconechy you mean the default card border? does it change the look of the card?

ericangeles:
@tmcconechy, I pushed a solution on moving card/s when selected. Let me know if that works for you.

tmcconechy (dismissed):
Ok i see @ericangeles . Yes that works well now 👍🏻 
Thanks all looks good here.

We may need to add the NG types.

ericangeles:
Expecting this to fail because of the visual image. I will update once it failed.

ericangeles:
> Ok i see @ericangeles . Yes that works well now 👍🏻
> Thanks all looks good here.
> 
> We may need to add the NG types.

Okay, the same ticket as well?

tmcconechy:
yeah might as well just pr the new types quick in there so we dont forget (on the same ticket)

tmcconechy (approved):

EdwardCoyle (approved):

Issue #5303, opened 2021-06-10
Personalize: form3 issue in mobile/responsive view


jbrcna:
**Describe the bug**
there's some gradient color in tabs when in mobile/responsive view

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/personalize/example-form3-short.html
    or https://main-enterprise.demo.design.infor.com/components/personalize/example-form3.html
2. Resize to moible view or open in Browserstack
3. See error
4. Play around with tabs

**Expected behavior**
![image](https://user-images.githubusercontent.com/56722184/121481800-13257f00-c9ff-11eb-8fa3-c38fb1d718a5.png)

**Version**

- ids-enterprise: v4.53.-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/56722184/121481767-0e60cb00-c9ff-11eb-8070-91994a406f64.png)

**Platform**
mobile/responsive view

**Additional context**

jbrcna:
QA Failed
When I switch to `Classic` theme & `Dark` mode the header becomes black so I refresh the page to see if it's the default color on `classic dark` but after refresh the header becomes azure and a gradient color appears in tabs.
v4.54.0-dev
verified in https://main-enterprise.demo.design.infor.com/components/personalize/example-form3-short.html
![Large GIF (416x434)](https://user-images.githubusercontent.com/56722184/126729775-6877634d-f707-476f-814c-2b31d38fd295.gif)
_iPhone 12 Pro Max browserstack_

![image](https://user-images.githubusercontent.com/56722184/126730350-9c126a7e-8e97-42b8-aab2-60ff1fc9d35f.png)
_chrome devtools_

jbrcna:
the issue is now resolved. moving this ticket Done. thank you.
v4.54.0-dev
verified in https://main-enterprise.demo.design.infor.com/components/personalize/example-form3.html
![image](https://user-images.githubusercontent.com/56722184/127620469-8c7f9ca4-4ec1-4b82-8cbe-cbff53e3b2ca.png)

tmcconechy:
Should be closed

PR #5446, opened 2021-07-21
5303 styling issue with tab overflow mobile

Jah25:

added style to the personalizer that sets the tab-container background image to none when the tab container is horizontal and has overflow tabs

Closes ([#5303](https://github.com/infor-design/enterprise/issues/5303))

-Build and run the application.
- navigate to this example: https://main-enterprise.demo.design.infor.com/components/personalize/example-form3-short.html
- Open devtools in chrome and display in responsive mode, choose a tablet or smaller screen size
- there should be no gradient on the overflow tab see image comparison below

Screen Shot 2021-07-22 at 10 12 51 AM


tmcconechy (changes requested):
@Jah25 rather than remove the effect i think we want to change color of the radial gradient as you can see it is working on the left side

Screen Shot 2021-07-21 at 10 04 59 AM

```
.tab-container.header-tabs::before {
  background-image: linear-gradient(to right, #0072ed, rgba(0, 114, 237, 0));
  left: 20px;
}
```

So maybe we can dynamically update that rule for the other side based on the current color.

Jah25:
@tmcconechy sure I can make that change.

tmcconechy:
See comment again added some more context for that was looking for the hextToRgba... But that might work what i just added / should be close

Jah25:
@tmcconechy So the rgba($body-color-primary-background) looks like it works when it's in the personalize component, as opposed to in the tabs-horizontal, I'm not exactly sure where it's inheriting the different body color value so I wanted to run it by you.

tmcconechy (approved):
Looks good to me now. Go ahead and update the images

deep7102 (approved):
🍌

PR #5470, opened 2021-07-28
5387 fix for tabbing issue toolbarflex

### LINKED ISSUE #5387: Issue tabbing through toolbar flex with allowTabs: true
**Describe the bug**
When a flex toolbar has no overflowed items, only one button is focusable when the allowTabs setting is turned on.

**To Reproduce**
https://js-447mrh.stackblitz.io/
Code here: https://stackblitz.com/edit/js-447mrh?file=index.html

Attempt to tab through two buttons (Text button and Text button 2), only Text button 2 gets focused

**Expected behavior**
Both buttons should be focusable.

**Version**

- ids-enterprise: 4.52.1

**Platform**
 - Infor Application/Team Name: Infor Nexus
 - OS Version: Mac OS X Catalina
 - Browser Name: Chrome
 - Browser Version: 91

Jah25:

added logic that sets tab-index to zero when allowTabs is set to true for a toolbar-flex item. this should allow the user to tab through toolbar options and prevent the situation where the user can neither tab or use arrow key navigation to reach valid options

([Closes #5387](https://github.com/infor-design/enterprise/issues/5387))

-build and run the application
-navigate to the following example http://localhost:4000/components/toolbar-flex/example-allowed-tabs-test.html
-using tab navigation you should be able to reach every button on the page.

### DIFF
--- a/app/views/components/toolbar-flex/example-allowed-tabs-test.html
+++ b/app/views/components/toolbar-flex/example-allowed-tabs-test.html
@@ -1,0 +1,17 @@
+<div class="flex-toolbar" data-options="{ 'attributes': [ { 'name': 'data-automation-id', 'value': 'my-toolbar'} ] }" role="toolbar" data-automation-id="my-toolbar">
+  <div class="toolbar-section buttonset">
+    <button class="btn" tabindex="0" data-automation-id="my-toolbar-button-0">
+      <span>Text Button</span>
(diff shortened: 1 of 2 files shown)


EdwardCoyle (dismissed):
@Jah25 Looks good

ericangeles (approved):

Issue #5286, opened 2021-06-09
Datagrid: Unable to reopen filter options after doing pagination and clicking other filter options


ktfields:
**Describe the bug**
Unable to open filter option again after doing pagination and clicking other filter options. Issue occurs on both server-side and client-side paging.

**To Reproduce**

Case 1:
1. Go to 
[client-side] https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/test-filter-paging-client-side.html 
[server-side] https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/test-filter-paging-server-side.html
2. Click on any filter options dropdown (keep it open)
3. Click on "Next Page" or "Last Page" button
4. Close the filter options dropdown
5. Try to open it again

Case 2:
1. Go to 
[client-side] https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/test-filter-paging-client-side.html 
[server-side] https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/test-filter-paging-server-side.html
2. Click on any filter options dropdown (Open at least 3)
3. Reopen the previous filter option

See attached gif files for reference.

**Expected behavior**
Filter options should be opened

**Version**
ids-enterprise: 4.52

**Screenshots**
Case 1:
![fileterOption_pagination_issue](https://user-images.githubusercontent.com/39935379/121300148-08001f80-c929-11eb-8034-350a6bb868b7.gif)

Case 2:
![fileterOption_randomClicks_issue](https://user-images.githubusercontent.com/39935379/121301795-4dbde780-c92b-11eb-950e-7ff71698da21.gif)

**Platform**
 - Browser Name: Google Chrome, Microsoft Edge
 - Browser Version: Google Chrome version 91.0.4472.77, Microsoft Edge version 91.0.864.41

**Additional context**
N/A

tmcconechy:
When fixing the first one i think it should close the filter when going to another page

nganotice:
This has been QA tested and passed on v4.55.0-dev.

https://main-enterprise.demo.design.infor.com/components/datagrid/test-filter-paging-client-side.html & https://main-enterprise.demo.design.infor.com/components/datagrid/test-filter-paging-server-side.html

Case 1:
**Client side**
![Client1](https://user-images.githubusercontent.com/84306838/129310491-60de500d-6eab-4729-b96e-65b4521e9b90.GIF)

**Server side**
[...]

PR #5484, opened 2021-08-03
5286 - Unable to reopen filter options after paging

ericangeles:

This PR fixes the reopening of filter options after doing pagination. In Popupmenu component, doesn't meet the condition for wrapping and unwrapping of the menu. (In `teardown()` method)

Closes https://github.com/infor-design/enterprise/issues/5286

1.) Scenario 1
- Go to http://localhost:4000/components/datagrid/test-filter-paging-client-side.html and http://localhost:4000/components/datagrid/test-filter-paging-server-side.html
- Click on any filter options dropdown (keep it open)
- Click on "Next Page" or "Last Page" button
- Close the filter options dropdown
- Try to open it again
- It should open the filter options

2.) Scenario 2
- Click on any filter options dropdown (Open at least 3)
- Reopen the previous filter option
- It should open the filter options


tmcconechy (changes requested):
Sorry @ericangeles I think this should work another way. (which might make it simpler actually).

So when you click any of the ager buttons it should close then (and then reopen) as when you click out of the popup it should usually close.

ericangeles:
@tmcconechy does this not behave as expected?

ericangeles:
@tmcconechy, hmm weird coz it gets close when moving to the other page on my end.

ericangeles:
The thing here is that the close method (from Popupmenu) doesn't really work and it doesn't meet the condition under `teardown()`. Can find here `if (!otherTriggers.length) {`.  For some reason, the id is just being replicated/moved when you click the other filter options.

ericangeles:
filter-reopen

tmcconechy:
This is what im seeing:
https://drive.google.com/file/d/1STVdW-S_Aq3qgI1lYA3uMvJlm2rbF0Zg/view

1. http://localhost:4000/components/datagrid/test-filter-paging-client-side.html
2. open filter dropdown
3. hit next page

tmcconechy:
Maybe you can add this https://github.com/infor-design/enterprise/blob/main/src/components/datagrid/datagrid.js#L1687 to the pager button click or page events

ericangeles:
@tmcconechy it seems that the `this.currentHeader` is not the right target that's why the `close()` method doesn't work at all.

tmcconechy (approved):
Ok, not sure whats wrong before but i do see it working now. Maybe my branch was out of sync. Sorry 👍🏻

EdwardCoyle (approved):

PR #5494, opened 2021-08-06
5482 Datagrid changing selectable

### LINKED ISSUE #5482: Datagrid: Issue changing cellNavigation when switching the selectable
Describe the bug
We are changing the selection mode between 'single' and 'mixed' on a datagrid.  
When in 'single' mode, set the option cellNavigation to false to enable moving through the datagrid rows with the hand pointer. When in 'mixed' mode, set the option cellNavigation to true to enable moving through the datagrid cells with the pointer cursor.

Issue is that the cursor and navigation are not changing when updated.

1. Using the attached 'test-change-selectable-and-cellnavigation.html'
2. Notice that datagrid navigation is row navigation and the cursor is a hand
3. Click on the 'Switch to Mixed Mode' button
4. Notice that datagrid navigation is still row and the cursor is still a hand

Expected behavior
Changing selection and cellNavigation take affect when switching.

Version
- ids-enterprise: v4.51.4

andriivarhanov:
In the datagrid selectable and cellNavigation settings apply on initialization only. I used hover/focus events to check the settings during the event to add/remove is-clickable/is-active-row css class rather than on initialization

Closes #5482 

- Go to: http://localhost:4000/components/datagrid/test-change-selectable-and-cellnavigation.html
- Hover/click a row, see cursor pointer, no cell navigation
- Click on the 'Switch to Mixed Mode' button
- See cell navigation is on and cursor is default
- Switch back, see cursor pointer and no cell navigation


tmcconechy (commented):
Thanks this seems to work just fine. The only improvement is maybe we can fix the double border when the first column is hidden.

We have a rule that hides it for first-child maybe we just need to add one so it also takes it off for when .is-hidden is the first-child.

andriivarhanov:
@tmcconechy I've fixed the double border https://github.com/infor-design/enterprise/pull/5494/commits/b4ceef6cb474612896350a179b37c5bb6fbf7e3e 

The other issue I've noticed datagrid header columns disappear when switching mode, video below. Should I fix it as well?

https://user-images.githubusercontent.com/5204821/128685133-40b146f8-4c71-4988-9d13-74bff96515a3.mov

tmcconechy:
@andriivarhanov yes please, I think that’s going to be a problem otherwise

andriivarhanov:
@tmcconechy fixed, the issue was in `hideColumn` method where `selectionCheckbox` removed in any case and added back in `showColumn` only if frozenColumns are in settings

tmcconechy (approved):

EdwardCoyle (approved):

Issue #5193, opened 2020-10-27
AppMenu: Ability to for end user to resize


marclouisgenedeguzman:
**Is your feature request related to a problem or use case? Please describe.**
Ability for users to resize Application Menu. In both 4.x and 5.x

**Describe the solution you'd like**
Please see attached for sample behavior
![application_menu_resizable](https://user-images.githubusercontent.com/50192038/97297549-2713da80-188d-11eb-9869-3f94a4c03e77.gif)

**Describe alternatives you've considered**
No alternatives found

**Additional context**
Sample on -> https://master-enterprise.demo.design.infor.com/components/applicationmenu/example-index.html
And what application do you work on? M3 User Productivity

tmcconechy:
@marclouisgenedeguzman 

From https://jira.infor.com/browse/UXD-2301

The product team should probably consider:

- Better naming for page names (shorter, more effective communication to the users)
- Are there significant user complaints about not being able to see the full page name when they have tooltips that show the full name on hover?
- IDS Responsive could have a menu that navigates (drills in) > to avoid lots of truncation, although this could get users lost very quickly 
- Product teams should limit the use of nested app menu navigation and focus on secondary navigation throughout the app via tabs, etc where possible. 
- Maybe IDS can add a splitter, although this might enable bad menu practices
- Wrapping text for menu items in cases where navigation is deep and names are long.

So passing that along but we will add this even though it might encourage bad practice.

katieinfor:
Peter's comment on the Jira ticket: https://jira.infor.com/browse/UXD-2301 
"Following-up on this from a UXW and content strat recommendations for this issue:

- Clear character limitations when the label is created/input.

- In-line help and hint text and /or cautionary messaging if labels are too long or unwieldy, delivered via modal, toaster or hover text.

- Use of beginning or middle truncation.

- Full text on hover."

tmcconechy:
Ok i guess what we can do with that info is add it to the app menu docs as guidellines? But not sure what the beginning or middle truncation means.

We do have tooltips now on hover https://main-enterprise.demo.design.infor.com/components/applicationmenu/test-tooltips.html

I think M3 still wants the resize no matter what guidance give though.

ericangeles:
@tmcconechy should we wait for the design to finalize?

tmcconechy:
Yes, please work on it @ericangeles they keep requesting it this way so even if we give the recommendations for design they still want it

nganotice:
[...]

PR #5501, opened 2021-08-06
5485 - Add puppetteer

tmcconechy:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

This PR adds jest-pupetteer to the EP project for use. We can start to convert some of the deprecated protractor tests. To start i converted about.e2e already.

Added some docs on how to use to TESTING.md and a section on some conversion notes on what i found during the process.

Also adds some late locales that came back late.

**Related github/jira issue (required)**:
Closes #5485

**Steps necessary to review your pull request (required)**:
- Note the checks are running puppeteer
- checks pass
- review the md files
- review the tests

**Included in this Pull Request**:
- [x] An e2e or functional test for the bug or feature.

PR #5505, opened 2021-08-09
5193 - Add ability to resize the app menu

ericangeles:

This PR adds the ability to resize the application menu. You can activate it via settings, `resizable: true`.

With this implementation, the app menu will,
- always opened
- it has a minimum width of `300px` and cannot be resized below `300px`
- it should save the last position of app menu after reloading (via `this.settings.savePosition`)
- the behavior on smaller viewport should still be the same
- show ellipsis if texts were too long / also on a deep level menu
- tooltip should also work

Not sure about the style of the resizer. (Open for suggestions and changes)
Might add puppeteer test (once the https://github.com/infor-design/enterprise/pull/5501 merge)

Closes https://github.com/infor-design/enterprise/issues/5193

- Go to http://localhost:4000/components/applicationmenu/example-resizable-menu
- Drag the edge of the app menu
- Play it around by resizing (scale up, scale down)
- Combine with opening and closing the menu then do resize
- On smaller view when scaling down, the app menu should behave the same way (which will disappear on a certain viewport)
- Open all the level menu to show how it will look
- Ellipsis should work when the text is too long, and tooltip should display


tmcconechy (commented):
Looks really good so far.

1) can you add some longer menu items and deeper sections to the example so we can test how that works?
2) can you push the scrollbar over a few pixels so the black section isnt visible on the right of it?

Screen Shot 2021-08-10 at 9 55 33 AM

3) Can it be saved in local storage so when you reload it stays in the same position?

deep7102:
@ericangeles this looks good,
only thing add api setting to toggle on/off the localstorage save

tmcconechy:
Yeah agreed we should add the save or not as a setting

tmcconechy (dismissed):
Looks great 👍🏻 
I guess that the scrollbar cant be all the way over covering the black or we run into the issue of it being hard to grab?

ericangeles:
Yes, we can maybe add a bit of px to be able to grab it easily?

tmcconechy:
Yeah or if the scrollbar can be up next with no dark on the rightside, maybe we can add some "invisible" part that lets you grab and drag. But the behavior is looking nicer to me now so willing to except the dark bit if its not fixable

tmcconechy:
ok good for now - lets see what other reviewers think

deep7102 (dismissed):

PR #5508, opened 2021-08-11
5443: Message max-width is hardcoded in modal

### LINKED ISSUE #5443: Message: Increasing the maximum width of the text message is not effective in the Messages
**Describe the bug**
The text message of ids card component does not expand with respect to the title length for 'Message'.

**To Reproduce**
1. Go to 'Message' component
2. Increase the 'title' length
3. Verify the text message is not adjusted w.r.t title length

**Expected behavior**
When the length of title is increased, the text message in the modal should be responsive w.r.t the modal at a minimum threshold.

**Version**
- ids-enterprise-ng: [v4.53.2]

**Screenshots**
image

**Platform**
 - Device : iPad Air 4
 - OS Version : v14.1
 - Browser : chrome, safari, stock browser]
 - Browser Version : Version 91.0.4472.77 (Official Build) (64-bit)

**Additional context**
unable to increase the width of text message.
https://github.com/infor-design/enterprise/blob/main/app/views/components/message/example-index.html

davidcarlsonberg:

Message max-width is hard-coded in modal so when title size is long the message width does not increase to the size of it's container.

Fixes #5443 

- Pull branch, run app, visit http://localhost:4000/components/message/test-long-title.html
- Click the  button
- Notice the message text extends to the full width of it's container.
- Also notice that the modal now resizes toward it's max-width if the message text is long (this may be an undesired side-effect of simply removing the max-width from the message and perhaps we should consider placing a max-width setting option on modal and determine a default there)

### DIFF
--- a/app/views/components/message/example-index.html
+++ b/app/views/components/message/example-index.html
@@ -9,7 +9,6 @@
     <button class="btn-secondary" type="button" id="allow-tags">Allow Tags</button><br/><br/>
     <button class="btn-secondary" type="button" id="disallow-tags">Disallow Tags</button><br/><br/>
     <button class="btn-secondary" type="button" id="huge-message">Huge Message</button><br/><br/>
-    <button class="btn-secondary" type="button" id="huge-title">Huge Title</button><br/><br/>
   </div>
 </div>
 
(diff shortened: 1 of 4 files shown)


tmcconechy (changes requested):
Yeah i dont think that behavior is going to be very good as the message looks "broken" at those lower break points.

I think what seems to work is to set `.modal-content .message` to `max-width: initial`
So maybe we can add a maxWidth setting instead where you can set the maxWidth to initial or any other value and it will override the default 368... then the example would just have maxWidth: 'initial'

Also could do this on a new page in the message folder so you dont need to update the visual regression image

tmcconechy (approved):

deep7102 (approved):

tmcconechy (approved):

Issue #5420, opened 2021-07-13
TimePicker: Issue with AM/PM in Chinese language (zh-CN)


dhumalkishor:
**Describe the bug**
Time picker control is not working properly with Chinese language.

**To Reproduce**
1. Set Soho Locale to zh-CN.
2. Open timepicker and select hour and minute value.
3. Click on 'Set Time' button (It would be translated in Chinese)
4. It will show error '无效时间' (Invalid time I guess).

Also, after re-opening the timepicker the Minute value will be get transferred to Hour value and Minute value becomes zero.

Or we can reproduce the same issue with below link.

https://main-enterprise.demo.design.infor.com/components/timepicker/example-index.html?locale=zh-CN
Steps to follow to reproduce the issue.

1. Open timepicker and select hour and minute value.
2. Click on 'Set Time' button (It would be translated in Chinese)
3. It will show error '无效时间' (Invalid time I guess).

**Expected behavior**
Time picker should show correct values on revisit. 

**Version**
- ids-enterprise-ng: V4.53.0

**Screenshots**
![MicrosoftTeams-image (7)](https://user-images.githubusercontent.com/42198413/125456319-cab98874-efe7-4af5-8f85-23132392e27c.png)

**Platform**
 - Device (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name [e.g. chrome, safari, stock browser]
 - Browser Version [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**

cpinatel7:
When can we expect a resolution for this issue? It's a showstopper for this new GRC language. Thanks

tmcconechy:
This is schedule for next months release (July - Aug)

cpinatel7:
Will your fix be limited to the "time" area? I need to assess what can be tested or not in Chinese S. as we'll have to go ahead with our first round of testing without your fix. Thanks

tmcconechy:
Yeah the fix would be limited to what is written above. Although we probably should test other locales with translated AM/PM. I expect this would impact just the timepicker but possibly need a change to the locale time/converstion

cpinatel7:
Thanks Tim

tmcconechy:
cc @deep7102 as he has fixed similar issues before just get in touch with him if you need @ericangeles

nganotice:
[...]

PR #5518, opened 2021-08-16
5420 - Fix rendering of time and period in chinese language

ericangeles:

This PR fixes the rendering of timepicker object `hours, minutes, and period` even if the time format is in standard time. Added a condition to identify if the period is before the time (position perse).

Basically, this fixed applies to all the standard time format `ah:mm`.

Closes https://github.com/infor-design/enterprise/issues/5420

1. Test this fix on the following locales
- http://localhost:4000/components/timepicker/example-index.html?locale=zh-CN
- http://localhost:4000/components/timepicker/example-index.html?locale=zh-Hans
- http://localhost:4000/components/timepicker/example-index.html?locale=zh-Hant
- http://localhost:4000/components/timepicker/example-index.html?locale=zh-TW 

```text
Follow these steps

- Click the trigger button
- Set the hours, minutes, and period
- Hit the set time button
- You should not get an error message
- Open the timepicker again
- The hours, minutes and period should be the same
```

2. Optional: Run the puppeteer test to make sure it will pass
```text
npm run puppeteer -- timepicker
Should pass all the tests
```


tmcconechy (changes requested):
Nothing wrong with the fix but can we make the dropdown a bit wider so you can see all the characters?

Screen Shot 2021-08-16 at 10 59 53 AM

Also wondering if we dont need the protractor test?

tmcconechy (approved):

deep7102 (approved):

Issue #5325, opened 2021-06-17
Typography: Create "paragraph" text style 


katieinfor:
**Is your feature request related to a problem or use case? Please describe.**
Jira Ticket: https://jira.infor.com/browse/UXD-2503 
Our Figma file currently uses text styles with line heights that adhere to the 8 pixel grid. This works for the single-line/single-word type that is most commonly used within our designs, but these standardized line heights are often too large for instances of text with more than one line. 

**Describe the solution you'd like**
We are proposing this solution: [Line Height Figma File Page 3](https://www.figma.com/file/h6j7e2OYg671IF8vQoHksG/IDS-Line-Height-%5BUXD-2503%5D?node-id=106%3A293)

Instead of a default line height associated with our type styles, we suggest a 140% line height.

For example,

`font-size: 16px; line-height: 20px;`
would become
font-size: 16px; line-height: 140%`

**Additional context**
This is for Design: Jira Ticket: https://jira.infor.com/browse/UXD-2503  
Thanks! 
![image-2021-04-30-12-25-55-322](https://user-images.githubusercontent.com/81309537/122403910-bec55680-cf4c-11eb-934a-843121b0eeee.png)
"Add users and manage permissions" is one sentence, but the large line height separates it into two independent ideas. The font size/line height in this example is 14/24.

![image-2021-04-30-12-22-33-912](https://user-images.githubusercontent.com/81309537/122403911-bec55680-cf4c-11eb-80f3-542cb76d0c6f.png)
The sentence now clearly flows from one line to the next. The font size/line height in this example is 14/18.

jbrcna:
this issue is now resolved. moving this ticket to done
 v4.55.0-dev
verified in 
https://main-enterprise.demo.design.infor.com/components/typography/example-index.html
![image](https://user-images.githubusercontent.com/56722184/130780386-708b7d86-f388-4cce-a198-d1ecebdb30d4.png)
https://main-enterprise.demo.design.infor.com/components/emptymessage/example-widgets.html
![image](https://user-images.githubusercontent.com/56722184/130780565-39f5aff2-6b30-43fc-8a69-04655596a20e.png)

PR #5523, opened 2021-08-17
5325 - Typography paragraph text style

tjamesallen15:

This PR adds the typography paragraph text style

Closes https://github.com/infor-design/enterprise/issues/5325

- Go to http://localhost:4000/components/typography/example-index.html
- See the examples under `Paragraph Text` to check the line height on 14px and 16px
- Go to http://localhost:4000/components/emptymessage/example-widgets.html
- See the examples under `Paragraph Text (Typography - Line Height)` to check the line height on paragraphs used on cards.


tmcconechy (commented):
Are the styles on https://main-enterprise.demo.design.infor.com/components/emptymessage/example-widgets.html ok? If so that might be the only example we need.

WIth this example sometimes what happens at lower breakpoints is it gets squished.

Screen Shot 2021-08-17 at 10 22 19 AM

src/components/cards/_cards.scss
```diff
@@ -217,6 +217,54 @@
     }
   }
 
+  &.typography {
```
tmcconechy, line 220: This is still not 100% what i expect. Because basically we add a new class `typography` that is never going to be used and an example in http://localhost:4000/components/typography/example-line-height-text.html that has extra scrollbars ect and more for QA to flag.

I'd suggest to try a different approach

1) remove the example and code
2) update http://localhost:4000/components/emptymessage/example-widgets.html so the line height is increased as per the designs
3) add an example section for "paragraph text" at the bottom on  http://localhost:4000/components/typography/example-index.html that uses this line height it will use this rule which you can change https://github.com/infor-design/enterprise/blob/main/src/components/typography/_typography.scss#L99

tmcconechy (changes requested):
Sorry this design is a little confusing i think i figured out the best thing to do here in this comment https://github.com/infor-design/enterprise/pull/5523/files#r691236299

tjamesallen15:
Hopefully, this one I got it right.

tjamesallen15:
I'll fix the e2e test issues once this fix is okay.

tmcconechy (dismissed):
Looks great now 👍🏻  thanks, go ahead and fix the e2e i think

tmcconechy (approved):

ericangeles (approved):

PR #5524, opened 2021-08-17
Ng1075 - fix overlapping of checkbox with label when editorOptions.multiple is true

ericangeles:

This PR fixes the overlapping of the checkbox with the label when `editorOptions.multiple` is set to true.

Closes https://github.com/infor-design/enterprise-ng/issues/1075

- Go to http://localhost:4000/components/datagrid/test-multi-select-dropdown-selection.html
- Go to `multiple selection` column
- Open the dropdown list
- Checkboxes and labels should not be overlapping each other

### DIFF
--- a/app/views/components/datagrid/test-multi-select-dropdown-selection.html
+++ b/app/views/components/datagrid/test-multi-select-dropdown-selection.html
@@ -1,0 +1,188 @@
+
+<div class="row">
+    <div class="twelve columns">
+      <div role="toolbar" class="toolbar">
+  
+        <div class="title">
+          Data Grid Header Title
+          <span class="datagrid-result-count">(N Results)</span>
+        </div>
+  
+        <div class="buttonset">
+            <button type="button" id="toggle-row-status" class="btn">
+              <span>Add Row Status</span>
+            </button>
+            <button type="button" id="validate" class="btn-menu">
+              <span>Validate</span>
+            </button>
+            <ul class="popupmenu">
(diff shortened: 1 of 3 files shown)


tmcconechy (dismissed):
As a note when you select the items it doesnt save them when it repopens? I guess that would mean using the multiselect editor? Maybe if that solves the issue we can make this work? If not im ok with it if it solves the original bug

tmcconechy (approved):

tmcconechy:
@ericangeles actually thinking about this, maybe the fix in NG is to use

```
formatter: Soho.Formatters.Multiselect, editor: Soho.Editors.Multiselect
```

That makes more sense and they would get the functionality? Does that work?

tmcconechy (changes requested):
Pausing for a question as i  think i realize a better (non fix)

ericangeles:
@tmcconechy let me try to test that one in NG.

ericangeles:
@tmcconechy I changed Dropdown to Multiselect but doesn't seem to work. I also added the Multiselect to the typing definition but no luck.

ericangeles:
@tmcconechy I am looking at this test page http://localhost:4000/components/dropdown/test-configured-as-multiselect.html and this should be the behavior of the dropdown multi select?

tmcconechy:
Hmm seems like we don't have a formatter and editor for Multiselect. I thought we did but we only have a filter type for it. Perhaps we should add one?

tmcconechy (approved):
awesome 👍🏻  glad we have these new editors now

PR #5528, opened 2021-08-17
5428 - Datagrid: Title XSS Vulnerability

### LINKED ISSUE #5428: Datagrid title vulnerability to xss
**Describe the bug**
When column definition of datagrid contains code, datagrid will execute such code.

**To Reproduce**

1. Change column definition in any datagrid to have name such as `test`
2. Open editted datagrid in the browser
2. See error

**Expected behavior**
Any malicious code should be sanitized and not executed

**Version**

- ids-enterprise: 9.5.x

davidcarlsonberg:

This change strips/sanitizes HTML added inside the name property of the columns array.

Closes #5428 

- Pull branch
- Run app
- Visit http://localhost:4000/components/datagrid/test-xss-prevention.html
- Notice no code is executed when placed inside columns array as value for name property.

### DIFF
--- a/app/views/components/datagrid/test-xss-prevention.html
+++ b/app/views/components/datagrid/test-xss-prevention.html
@@ -62,8 +62,8 @@
         //Define Columns for the Grid.
         columns.push({ id: 'selectionCheckbox', sortable: false, resizable: false, formatter: Formatters.SelectionCheckbox, align: 'center'});
         columns.push({ id: 'productName', name: 'Product Name', sortable: false, field: 'productName', formatter: Formatters.Hyperlink, editor: Editors.Input});
-        columns.push({ id: 'activity', name: 'Activity', field: 'activity', editor: Editors.Input});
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
One small problem with this is we previously do allow some html in here. For example this fix breaks https://main-enterprise.demo.design.infor.com/components/datagrid/test-two-line-header.html (two line header with a `<br>`).

Try `xssUtils.stripTags = function (html, allowed) {` instead and then we can allow only any text elements to be safe. like maybe

```
xssUtils.stripTags(col.name, `<div><span><a><small><svg><i><b><use><br><strong><em>') : '';
```

davidcarlsonberg:
This seems like an odd test failure to me. Any ideas?
https://github.com/infor-design/enterprise/pull/5528/checks?check_run_id=3362038751#step:12:673

tmcconechy:
@davidcarlsonberg that one happens occasionally random. I just reran it and that should work

tmcconechy (approved):

ericangeles (approved):

PR #5534, opened 2021-08-19
5445. Added Changelogs and Fix Eslint

tjamesallen15:

This PR fixes a bug card group toolbar overlaps then disappears after clicking the checkboxes.

Closes https://github.com/infor-design/enterprise/issues/5445

- Go to http://localhost:4000/components/cards/example-group-action.htmll
- Click on the 4th check box then the "Card group action" toolbar overlaps with the Card Title
- Scroll down to the last checkbox then the "Card group action" toolbar disappears

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## v4.55.0 Fixes
 
 - `[Charts]` Fixed a bug where automation ids is not properly rendered on legend, text and slices. ([#5441](https://github.com/infor-design/enterprise/issues/5441))
+- `[Cards]` Fixed a bug card group toolbar overlaps then disappears after clicking the checkboxes. ([#5445](https://github.com/infor-design/enterprise/issues/5445))
 - `[Datagrid]` Fixed an issue where tree list indentation is not left aligned when row has no children and datagrid row height is extra small or small. ([#5487](https://github.com/infor-design/enterprise/issues/5487))
 - `[Message]` Added maxWidth setting to allow message to go full width when title is long. ([#5443](https://github.com/infor-design/enterprise/issues/5443))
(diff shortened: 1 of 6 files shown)


tjamesallen15:
I tried running the functional:ci on my local but I don't get the same error that was displayed in this check.

tmcconechy:
@tjamesallen15 i think this is one that fails "occassionally/randomly" i usually just rerun the checks. So i did that for this case

tmcconechy (dismissed):

tmcconechy (changes requested):
It looks like this fix isnt 100% working for example it broke the layout of http://localhost:4000/components/counts/example-widget-count.html

Aso wondering if it could have other effects on existing widgets? Is there a less impactful way we can fix this

src/components/listview/listview.js
```diff
@@ -212,6 +212,9 @@ ListView.prototype = {
     const cardWidgetContent = this.element.parent('.card-content, .widget-content');
     if (cardWidgetContent[0]) {
       cardWidgetContent[0].style.overflow = 'hidden';
+      cardWidgetContent.css('display', 'flex');
+      cardWidgetContent.css('justify-content', 'flex-end');
+      cardWidgetContent.css('flex-direction', 'column');
```
ericangeles, line 217: You can wrapped all the css attributes on a single `css()` jQuery method something like this,

```javascript
cardWidgetContent.css({
    'display': 'flex',
     'justify-content': 'flex-end',
     'flex-direction': 'column'
});
```

ericangeles (commented):
I also noticed when you click the fourth one it kinda moves a bit?

tmcconechy (approved):
Looks good now 👍🏻  Thanks @tjamesallen15

Issue #5435, opened 2021-07-16
Support for Vertical Scroll-to-End Event in Data Grids


ayoayco:
**Is your feature request related to a problem or use case? Please describe.**
We have a number of use cases where we need to fetch the next set of records from a backend server when the user scrolls to the end of the list. This is both for the standalone Data Grid and the one in the Soho Lookup dialog.

**Describe the solution you'd like**
Ability to listen to the scroll event and/or pass a callback function to trigger the `source(req, response)` when the scroll reaches the end of the list

**Additional context**
And what application do you work on? M3 App Designer & M3 H5 UI Team

tmcconechy:
As a not it should be possible to add an event handler to the DG scroll container and workout this logic. But i don't have a problem adding this to the backlog and providing an official event

ayoayco:
Thank you, Tim! I'm having difficulty doing it, especially with the Soho Lookup dialog DG.

jbrcna:
This has been QA tested and passed on v4.55.0-dev.
https://main-enterprise.demo.design.infor.com/components/datagrid/example-vertical-scroll-to-end-event.html
![image](https://user-images.githubusercontent.com/56722184/131448541-e8e5116d-0396-46de-a053-525a9d8382dc.png)

PR #5539, opened 2021-08-19
5435 vertical scroll to end

ericangeles:

This PR adds `verticalScrollToEnd` property when you reach the end of the Datagrid list. This will only work on the list that has a vertical scrollbar.

Closes https://github.com/infor-design/enterprise/issues/5435

- Go to http://localhost:4000/components/datagrid/example-vertical-scroll-to-end-event.html
- Scroll until you hit the bottom of the list
- A toast message will appear `Hooray! You reached the bottom of the list.`
- Play around with the list, scrolling up and down to test the functionality

### DIFF
--- a/app/views/components/datagrid/example-vertical-scroll-to-end-event.html
+++ b/app/views/components/datagrid/example-vertical-scroll-to-end-event.html
@@ -1,0 +1,39 @@
+<div class="row top-padding">
+  <div class="twelve columns">
+
+    <div id="datagrid" style="height: 300px;">
+    </div>
+
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+
+      var grid,
+        columns = [];
+
+      //Define Columns for the Grid.
+      columns.push({ id: 'productId', name: 'Product Id', field: 'productId', width: 140, formatter: Formatters.Readonly});
+      columns.push({ id: 'productName', name: 'Product Name', sortable: false, field: 'productName', width: 150, formatter: Formatters.Hyperlink});
+      columns.push({ id: 'activity', hidden: true, name: 'Activity', field: 'activity', width: 125});
+      columns.push({ id: 'quantity', name: 'Quantity', field: 'quantity'});
+      columns.push({ id: 'price', name: 'Price', field: 'price', formatter: Formatters.Decimal});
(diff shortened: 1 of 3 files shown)


ericangeles:
cc: @ayoayco

tmcconechy (changes requested):
I dont have huge problem with this way but i was thinking about a better fix.

1) listen to the wrapper scroll event
2) then trigger it on the parent

```
this.element.trigger('scroll', {percentScrolled: 100});
```

Then provide a percent instead so they have more flexibility and pass it in the event.

Theni nthe example they would just do

```
 $('#datagrid').on('scroll', function(e, args) {
     console.log(args.percent + '# scrolled');

**Comment by ayoayco** [comment]
Would it be simpler if you add boolean property in the Datagrid options (something like `triggerSourceOnScrollEnd`) that if set true, the Datagrid will call the `SohoDataGridSourceFunction` callback function when the user reaches the end of the vertical scroll? That would be easier for us as well since we already give the call back function `SohoDataGridSourceFunction`

**Comment by tmcconechy** [comment]
@ayoayco i guess we could add both but generally I feel like the paving is different than scrolling. And I like the idea of the scroll event for other possible use cases (not just this one)

**Comment by ayoayco** [comment]
> @ayoayco i guess we could add both but generally I feel like the paging is different than scrolling. And I like the idea of the scroll event for other possible use cases (not just this one)

Ah, I see! Yeah, it's actually the paging that I want to accomplish.

**Comment by tmcconechy** [comment]
Do you have a pager bar too? As i think that complicates the interaction. I thought the use case was to load more when you get to the bottom? If so i think firing the event and having the percentage scroll down will be the most flexible. Then when you get to 95% or 98% you can load more data into the grid

**Comment by ayoayco** [comment]
No pager bar, just the scroll event (but still called paging in our team somehow). Yeah, the percentage solution sounds good... we can call the source function based on the event. Do you guys have ideas on how we'll be able to use this as well in Angular when it comes out in enterprise-ng?

**Comment by tmcconechy** [comment]
We are thinking we will give the event. And then you could do something like this

```
$('#datagrid').on('scroll', function(e, args) {
     console.log(args.percent + '# scrolled');     if (args.percent > 98) {       
     grid.triggerSource({
          activePage: 2
        }); // Or just get your data and feed it in with other API calls
```

We will have to add the new event to angular as well.

PR #5540, opened 2021-08-20
5519 - Added filter mask options in datagrid

### LINKED ISSUE #5519: Datagrid: Allow a mask to be applied to a column's cell but no mask or a different mask to the filter cell
Issue:
Currently mask options are applied to a column meaning the same mask will be applied when editing a column cell and the same mask will be used in the column's filter row cell.

Request:
Would like the ability to provide a mask for a column's cell and a different mask or no mask to the column's filter cell. 

Alternatives considered:
There are no viable alternates to pursue.

Additional context:
Working on a Landmark feature to allow the user to select or enter multiple lookup select values in the filter row cell for filtering the list.  When the list allows editing a column's cell value the mask applied will hamper the user's ability to enter multiple search values in the filter cell.

yohannahbautista:

Added filterMask and filterMaskOptions in column settings. Column mask will be used on filter if filter mask options are not available.

Closes #5519 

- Go to: http://localhost:4000/components/datagrid/test-filter-mask.html
- Type on the 4 qty rows
- 1st one : Filter Allows: 3 char, Edit in the cell it allows any number
- 2nd one : Filter Allows: 1 char, Edit in the cell Allows: 1 char
- 3rd one : Filter Allows: 3 char, Edit in the cell Allows: 1 char
- 1st one : Filter Allows any number Edit in the cell it allows any number

- [ ] Change in filter mask example page


tmcconechy (changes requested):
The logic for this needs to be such that it still uses col.mask if present or it's going to break for people. The idea is to use the existing mask/logic and add an option to have it different for the filter row if need be.

- if col.mask and no col.filterMask use  col.mask 
- if col.mask and col.filterMask use  col.filterMask for the filter row and col.mask

ericangeles (approved):
Looks good to me now! 🍌

tmcconechy (approved):

tmcconechy (approved):
Approving as this works now. I updated the example and test steps

EdwardCoyle (approved):

ericangeles (approved):

Issue #3618, opened 2020-03-17
Popdown: Provide a "clickOutside" event


edgarinfor:
**Is your feature request related to a problem? Please describe.**
This request is related to issue infor-design/enterprise-ng#760 where we are having popdowns closing when the lookup fields are opened.

In our case, we also have **custom lookup fields** which behave very similar to soho's but, don't use the same exact code.

![wfm_popdown_01](https://user-images.githubusercontent.com/39097500/76870871-b4e00f80-6840-11ea-838d-b4e4a309d701.png)

![wfm_popdown_02](https://user-images.githubusercontent.com/39097500/76870883-b9a4c380-6840-11ea-914c-bba495204f7c.png)

**Describe the solution you'd like**
The soho popdown directive should have a "clickOutside" event that can be combined with **keepOpen** to have more control over when to close or not the popdown even for other things other than lookup dialogs.

**Describe alternatives you've considered**
We came up with a workaround that consists on using **keepOpen** on the popdowns and tracking when any click is done outside the popdowns. We combine that event with checking if a dialog has been opened before or not and decide if it was open from within the popdown and so, we decide to close it or not. 

The workaround was done by extending the **SohoPopDownDirective** with our own and doing all the extra code there; however, this API should belong to the directive because we had to access the popdown internals (e.g, JQuery, JS).

**Additional context**
The non-angular popdown .\src\components\popdown\popdown.js controls the click event (line 335):
```javascript
      if (!self.settings.keepOpen) {
        $(document)
          .on('click.popdown', handleFocusOut)
          .on('focusin.popdown', handleFocusOut);
      }
```
The click.popdown event should also be added when keepOpen is set and extend the handleFocusOut so it can use keepOpen internally and dispatch the clickOutside as needed.

You can refer to our customized directive for reference: [workbrain.popdown.ts.zip](https://github.com/infor-design/enterprise-ng/files/4344211/workbrain.popdown.ts.zip)

tmcconechy:
Also looking at this im wondering if a new pattern like https://master-enterprise.demo.design.infor.com/components/page-patterns/example-search-form.html might be interesting to explore rather than nested modals.

tmcconechy:
Is this still needed?

edgarinfor:
Hi, Tim, yes, we still need it... In our case we display a (custom) lookup field (which opens a dialog) within a popdown.

I tested with the latest changes, and it seems to work fine using an IDS lookup field but, not when opening dialogs from other components.

[...]

PR #5542, opened 2021-08-23
3618 - Popdown Click Outside Event

tjamesallen15:

This PR adds a click outside event in popdown.

Closes https://github.com/infor-design/enterprise/issues/3618

- Go to http://localhost:4000/components/popdown/test-click-outside.html
- Open `Developer Console`
- Click `My Cart`
- Click Outside of the Popdown

### DIFF
--- a/app/views/components/popdown/test-click-outside.html
+++ b/app/views/components/popdown/test-click-outside.html
@@ -1,0 +1,51 @@
+<div class="row">
+    <div class="twelve columns">
+  
+      <button id="popdown-example-trigger" class="btn-tertiary" type="button" data-popdown="true">
+        <svg class="icon" focusable="false" aria-hidden="true">
+          <use href="#icon-cart"></use>
+        </svg>
+        <span>My Cart</span>
+      </button>
+      <div id="popdown" class="popdown hidden" data-automation-id="popdown-automation-id">
+        <div class="widget">
+          <div class="widget-header">
+            <h2 class="widget-title">My Cart <span class="extra">(8 Items)</span></h2>
+          </div>
+          <div class="widget-content">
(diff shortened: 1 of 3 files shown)


EdwardCoyle (approved):

tmcconechy:
@edgarinfor can you check if this will work for your use case? I can see its just a method call you can do to close it but i was wondering if you need an event to fire? The issue seems to indicate an event @tjamesallen15 ?

tmcconechy:
@tjamesallen15 what i am thinking for this is to simply trigger a `clickoutside` event when you click out allowing them to close it or not with the close function

tmcconechy (dismissed):

edgarinfor:
Hi @tmcconechy! I haven't had the chance to test the changes but, looking at the code, it seems that he added a "blur" event instead of "clickOutside" which could be use together with "keepOpen", if that's the case, I think this should work.

@tjamesallen15 , if we set "keepOpen" to "true" and listen to the "blur" event, will we be able to manually close the popover manually if we want to?

tmcconechy:
@edgarinfor i tested it and it seems to work in the new example it fires a clickoutside event as i expect. I think for context the `clickoutside` might be the best name? But is the blur event then redundant?

edgarinfor:
oh, I saw the changes yesterday, now I see there's a "clickoutside" as well... I think that "clickoutside" describes better what we wanted specifically but, "blur" looks like it could be a more generic synonym, unless we can think of an edge case when they can actually behave differently? Now, it seems redundant though.

tmcconechy:
yeah i agree. i think either works but having both is redundant. take your pick @tjamesallen15 maybe just have one

tjamesallen15:
@tmcconechy I chose the `clickoutside` event.

tmcconechy (approved):

PR #5546, opened 2021-08-25
5513 - Added copy option in swaplist

### LINKED ISSUE #5513: SwapList
It would be possible maintain left side content("item") when moving selected item from left side to right.

yohannahbautista:

Added enable copy settings, the default is false. If a list has copy enabled, the item is cloned and added to the target list instead of being moved.

Closes #5513 

- Go to: http://localhost:4000/components/swaplist/example-enable-copy.html
- Select  an item from copy list and click on the arrow button

- [ ] A new example page.

### DIFF
--- a/app/views/components/swaplist/example-enable-copy.html
+++ b/app/views/components/swaplist/example-enable-copy.html
@@ -55,12 +55,6 @@
                 <use href="#icon-left-arrow"></use>
               </svg>
             </button>
-            <button class="btn btn-moveto-right" type="button">
-              <span class="audible">Move to right</span>
-              <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
-                <use href="#icon-right-arrow"></use>
-              </svg>
-            </button>
           </div>
         </div>
         <div class="card-content">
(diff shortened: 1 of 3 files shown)


tmcconechy (commented):
This looks perfect except is `copy` is a bit misleading? What do others think of  the setting to called `keepInList` ? cc @deep7102 @EdwardCoyle

app/views/components/swaplist/example-enable-copy.html
```diff
@@ -0,0 +1,97 @@
...
+        </div>
+      </li>
+    {{/text}}
+    {{/dataset}}
+  </ul>
+</script>
+
+<div class="row">
+  <div class="columns nine">
+    <div class="swaplist one-third" id="example-swaplist-1" data-init="false">
```
deep7102, line 31: Please remove the css class `on-third`, swap list use this internally so by adding here will cause to the mobile view.

deep7102 (changes requested):
Looks good! just two things as
- As @tmcconechy me too go with `keepInList`
- Remove css class `one-third` from example page

Issue #4600, opened 2020-11-11
Split program when a user drops a tab on another tab


marclouisgenedeguzman:
**Is your feature request related to a problem or use case? Please describe.**
Use case: Our classic program have the ability for split screen when a user drops a tab on another tab

**Describe the solution you'd like**
Ability for users to split 2 programs when they drop a tab on another tab.

**Describe alternatives you've considered**
Currently we have styling in place to make it look like the programs are split.
Below is an example of the feature from our classic UI:
![split program when a user drops a tab on another tab](https://user-images.githubusercontent.com/50192038/98809831-2cf6e780-2459-11eb-95c3-415a3a6299c1.gif)

**Additional context**
And what application do you work on? M3 H5 UI Team

Tagging @lucacolumbu

tmcconechy:
We have some of feature called multi tabs. https://master-enterprise.demo.design.infor.com/components/tabs-multi/example-index.html

In this example you can move tabs via the right click menu. I think maybe the drag option to split is missing but thats also not that discoverable. 

For dragging we need https://github.com/infor-design/enterprise/issues/4520 first

katieinfor:
We need to investigate the behavior before moving forward. Thanks!

tmcconechy:
@ericangeles @tjamesallen15 - can you guys add a PR with types for the NG. Thanks!

ericangeles:
@tjamesallen15 please proceed to this.

PR #5612, opened 2021-09-15
4600 - Tabs (Split and Drag)

tjamesallen15:

This PR adds the ability to split tabs.

In this approach, the tabs will
- Tabs can be transferred via drag and drop
- Tabs can be splitted via dragging the tab below and releasing it.

Closes https://github.com/infor-design/enterprise/issues/4600

- Go to http://localhost:4000/components/tabs-multi/example-drag-split.html

**Transfer Tabs from Tab Group to another Tab Group**
- Drag the `Contracts` tab to the another tab Group on the right side.
- The Tab Group where the `Contracts` tab came from will be hidden.

**Tabs Splitting**
- From the example above, drag any of the tabs below and then release it. (In example. `Attachments`)
- The dragged tab will then be split from that Tab Group and will be transferred to another Tab Group.


src/components/tabs-multi/multi-tabs.js
```diff
@@ -8,7 +8,8 @@ const COMPONENT_NAME = 'multitabs';
 
 // Default Settings for MultiTabs
 const MULTITABS_DEFAULTS = {
-  tabContainers: []
+  tabContainers: [],
+  sortableTabs: false
```
tmcconechy, line 12: In other places we uses "sortable" as a setting.

tmcconechy (commented):
Looks like a great start but in terms of behavior

1) When i drag from the right to left the tab "snaps back" and then loads and seems to take a while. I would expect it not to snap back
2) When i drag the one on the left over to the right so there is one set then how can i split tabs. I think this is the use case on the issue where they start with no splits and then drag a tab ON another tab to make it split.

tjamesallen15:
Will work on this one. Thanks for reviewing @tmcconechy

tjamesallen15:
To split the tabs, drag one tab below and then release it. 
015A8D90-D9E2-49DD-B4C7-09522C78531E_1_105_c

tmcconechy:
I think this is looking really good. The only thing i would do is instead of grey when you drop it down the first time on the shadow: try azure 01 #E6F1FD

Otherwise i think this works perfectly.

tmcconechy (approved):

src/components/tabs/_tabs.scss
```diff
@@ -932,6 +932,27 @@ html.is-firefox {
   }
 }
 
+.tab-panel-container {
+  .tab-overlay {
+    position: fixed;
```
ericangeles, line 937: @tmcconechy it looks like that the stylelint doesn't work anymore? I remember that we have lint for alphabetical order in CSS/SCSS.

ericangeles (approved):

src/components/tabs/_tabs.scss
ericangeles: @tmcconechy I like the ordering by group type. I'll keep this in mind. Thanks! :) FYI @tjamesallen15 @yohannahbautista

PR #5627, opened 2021-09-20
NG1102 - Fix the first item selection does not reflect the value immediately

ericangeles:

This PR fixes the selection of the first item on the list that doesn't trigger the `change` event. That event will show/select the value immediately. The condition doesn't meet to trigger it.

It has been reported in the NG wrapper but the fix is on the EP.

Closes https://github.com/infor-design/enterprise-ng/issues/1102

- Go to http://localhost:4000/components/dropdown/test-first-selected-on-list
- Select the first item on the list `Alabama`
- `Selected: Alabama` should immediately show on top of the dropdown element


tmcconechy (approved):
👍🏻

deep7102 (dismissed):

Issue #5640, opened 2021-09-22
Line: If all data points are zero the line is centered


tmcconechy:
**Describe the bug**
 If all data points are zero the line is centered. This regressed between 4.51 and 4.55 so should dig into it from that direction.
 
**To Reproduce**
1. Edit https://github.com/infor-design/enterprise/blob/main/app/views/components/line/example-zero-millions.html#L22-L26 so that all the data is zero i.e.
```
    var dataset = [{
      "data":[
        {"name":"08/30/17","value": 0},
        {"name":"09/29/17","value": 0},
        {"name":"10/30/17","value": 0},
        {"name":"11/30/17","value": 0},
        {"name":"12/28/17","value": 0}],
      name: 'Component C'
    }];
```
3. Run page

**Expected behavior**
Should work as in 4.51 version where line is at the bottom

**Version**
- ids-enterprise: [4.51 to 4.55]

**Screenshots**

![Screen Shot 2021-09-22 at 11 57 06 AM](https://user-images.githubusercontent.com/814283/134379057-181d903e-460a-437a-be9c-877a6124c7c1.png)

**Additional context**
https://jira.infor.com/browse/LIME-7092

tmcconechy:
Looking at the history maybe either this was a change by @deep7102 or @ericangeles 
https://github.com/infor-design/enterprise/commits/main/src/components/line/line.js

Any ideas?

ericangeles:
@tmcconechy I am not sure about this issue before but I can take look at it.

tmcconechy:
Yes please. I would check the history a bit https://github.com/infor-design/enterprise/commits/main/src/components/line/line.js

Deepak told me he doesnt think it was his change? Maybe when we added the combined line / bar chart?

ericangeles:
@tmcconechy I haven't PR yet the combine line and column chart.

tmcconechy:
oh my mistake still might be something in the recent history https://github.com/infor-design/enterprise/commits/main/src/components/line/line.js

CindyMercadoReyes:
This issue is now resolved.

PR #5654, opened 2021-09-24
5640 - Fix line position when all values are zero

ericangeles:

This PR fixes the position of the line chart when all values are zeros. The line chart should not be centered and it should be close to the names.

Closes https://github.com/infor-design/enterprise/issues/5640

- Go to http://localhost:4000/components/line/test-all-data-points-are-zero-look.html
- The line chart should be close below with the names and not be centered

### DIFF
--- a/app/views/components/line/test-all-data-points-are-zero-look.html
+++ b/app/views/components/line/test-all-data-points-are-zero-look.html
@@ -1,0 +1,41 @@
+<div class="row">
+    <div class="two-thirds column">
+        <div class="widget">
+          <div class="widget-header">
+            <h2 class="widget-title">Line Chart: Zero Value</h2>
+          </div>
+          <div class="widget-content">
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
I can see one problem with this is that the transform 38% implies one specific height. If you make the widget a different height (for example class="widget double-height") then it wont be on the bottom.

One idea for a fix is maybe if the left side axis should be shown. So maybe if the scale is 0,0 you can default it to 0,100 and it will automatically work and also show something that way on the left side.

ericangeles:
@tmcconechy I also updated the test page to see both working on standard and double height.

tmcconechy (approved):
Ok this works 👍🏻

tmcconechy (approved):

deep7102 (approved):

PR #5674, opened 2021-09-28
1127- Fix regression of switch in popupmenu

tmcconechy:

Although its an odd case to put a switch in a popupmenu. This is something the team did on #1127 so added support back since it broke. (we may or may not support this on the future 5.x version)

Fixes https://github.com/infor-design/enterprise-ng/issues/1127

- go to http://localhost:4000/components/button/test-menubutton-switch.html
- click the button
- click the link and make sure it toggles


ericangeles (approved):

deep7102:
Switch working fine, like to let know about popup padding -might be just the example page

image

tmcconechy:
@deep7102 this isnt really a supported case was just a fall back situation. So i didn't add any styles to this. The team would do that if they want to do this..

deep7102 (approved):

Issue #5588, opened 2021-09-08
Counts: Mobile Enhancements and style changes


tmcconechy:
**Describe the bug**
Make some style changes to counts and make it work better on mobile.
 
**Fixes**
https://www.figma.com/file/1favYjXB0MGSSfga1Z5TCu/Count?node-id=0%3A1

- [ ] 48 Reg instead of 50 (desktop as well)

**Additional context**
https://jira.infor.com/browse/UXD-2777
Please fix this in the web components too on a seperate PR

ericangeles:
@tmcconechy looks like the WC is now using `48px`.

tmcconechy:
Ok sounds good. Nothing to do there then. I think this is because we already normalized the font sizes on this version

Issue #5566, opened 2021-08-31
Counts: title in RTL is misaligned


jonathanamiran:
**Describe the bug**
When using Counts circle widget in RTL the title sometimes not shown at the bottom of the circle

**To Reproduce**

1. Go to [https://design.infor.com/code/ids-enterprise/latest/demo/components/counts/example-widget-count?theme=uplift&variant=light&locale=he-IL](url)
2. Compare to [https://design.infor.com/code/ids-enterprise/latest/demo/components/counts/example-widget-count?theme=uplift&variant=light&locale=en-US](url)

**Expected behavior**
The title should be shown at the bottom center

**Version**

- ids-enterprise: 4.54.1

**Screenshots**
![image](https://user-images.githubusercontent.com/34129313/131523109-e579c469-754d-4812-8f1e-9bb333a00393.png)

**Platform**

 - OS Version: Windows 10 
 - Browser Name: Chrome/Edge
 - Browser Version: latest

**Additional context**
Thanks!

tmcconechy:
Also in the regular version the text is a little bit high. Not as bad as RTL but should probably fix it there as well.

PR #5706, opened 2021-10-04
5588/5566 - Fix title and icon position in RTL and update font size of count xl

ericangeles:

This PR fixes the title and text position of count in RTL and updates the font size of `xl text` from `50px` to `45px`.

Closes https://github.com/infor-design/enterprise/issues/5588 https://github.com/infor-design/enterprise/issues/5566

- Go to http://localhost:4000/components/counts/example-index.html
- Inspect the `xl text`, and it should be `45px` now
- Go to http://localhost:4000/components/counts/example-widget-count?&locale=he-IL
- Check the count title, it should be centered and below the counts
- Count icon should properly position upper right
- Test also on classic

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,10 @@
 # What's New with Enterprise
 
+## v4.57.0 Fixes
+
+- `[Counts]` Updated the font size of `xl-text` from `50px` to `45px`. ([#5588](https://github.com/infor-design/enterprise/issues/5588))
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
I think we should fix two more things

1) this info tooltip seems out of place. maybe its not supposed to be in the example? If it is we can fix it for RTL
2) the counts are not centered within the widget in RTL

Screen Shot 2021-10-04 at 9 35 36 AM

Also i think an image test was failing

tmcconechy:
The only additional suggestion is i think we should hide that blue info icon on the header at smal breakpoints so it doesnt do this...

Screen Shot 2021-10-05 at 10 11 33 AM

Issue #5442, opened 2021-07-20
Datagrid: Fallback Image when image can not be loaded


hemantg05:
We would like to show a fallback image in case the image url provided to the datagrid image column doesn't load successfully.

**Describe alternatives you've considered**
```
(Soho.Formatters as any).ImageUnavailableFallbackCustomFormatter = (
      row: number,
      cell: any,
      value: any,
      column: SohoDataGridColumn,
      item: any,
      api: SohoDataGridStatic
    ) => {
     return `<img src="`+
      'http://stackoverflow.com/does-not-exist.png' +
      `" onerror="console.log('errror'); this.src='https://cdn.sstatic.net/Img/unified/sprites.svg?v=e5e58ae7df45'"
       alt="" width="35" height="35">`
    };
    
```

 But onerror is removed by IDS for Security concerns (related to XSS attacks).

I referred to this answer for fallback image solution. https://stackoverflow.com/a/7995298

**Additional context**
Infor PIM

bbmuni:
@tmcconechy   Please prioritize this ticket as we have  issue reported by customer related to this functionality

tmcconechy:
@ericangeles for the design we should use the placeholder images / icons here similar to this https://main-enterprise.demo.design.infor.com/components/images/example-index.html

nganotice:
This has been QA tested and passed on v4.57.0-beta.

https://4570-beta0-enterprise.demo.design.infor.com/components/treemap/example-stats.html?theme=new&mode=light
<img width="1261" alt="Screen Shot 2021-11-04 at 2 16 22 PM" src="https://user-images.githubusercontent.com/84306838/140421362-8b1b7afb-c4eb-4ea3-b4bc-a2e4a9e37a52.png">

Moving this ticket to Done.

Thanks!

tmcconechy:
Testing with @hemantg05 if you use add row or re-render then this fails to work. Pushing a new easy fix

jbrcna:
This has been QA tested and passed on v4.61.0-beta.
https://4610-beta0-enterprise.demo.design.infor.com/components/datagrid/test-images-error.html
<img width="1380" alt="image" src="https://user-images.githubusercontent.com/56722184/157383686-e2b73c9a-2bc2-40a8-ba69-b1271cb025ac.png">

PR #5757, opened 2021-10-20
5442 - Datagrid Fallback Image when images cannot be loaded

tjamesallen15:

This PR adds fallback images when images cannot be loaded on datagrid.

Closes https://github.com/infor-design/enterprise/issues/5442

- Go to http://localhost:4000/components/datagrid/test-images-error.html
- See the placeholder

### DIFF
--- a/app/views/components/datagrid/test-images-error.html
+++ b/app/views/components/datagrid/test-images-error.html
@@ -1,0 +1,39 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="readonly-datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+    $('body').on('initialized', function () {
+     //Locale.set('en-US').done(function () {
+      var columns = [],
+        data = [],
+        imgExample = '{{basepath}}/images/natural1.png', // This image doesn't really exist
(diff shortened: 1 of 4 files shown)


tmcconechy:
What about...

```
this.element.find('.datagrid-img').off('error.datagrid').on('error.datagrid', (e) => {
      const targetEl = $(e.target);
      const parentEl = targetEl.parent();

      parentEl.append('<svg class="icon" focusable="false" aria-hidden="true" role="presentation"><use href="#icon-insert-image"></use></svg>');
      targetEl.remove();
    });
```

We could also try to put the logic inline in the formatterr itself `<img onerror="??"`. Also this looks good aside from  that improvement but the icon is not centered still

tmcconechy (approved):

ericangeles (approved):

hemantg05:
@tmcconechy , @ericangeles,  how does this work? Do we app developers have the possibility to provide our own fallback image or is the fallback image provided by IDS and can not be changed by app developers?

tmcconechy:
@hemantg05 we provide the icon that was specified by design

Issue #5791, opened 2021-10-27
Toolbar Search: close button 'x' is not centered


janahintal:
**Describe the bug**
Toolbar Search: close button 'x' is not centered when in responsive view then maximized

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/toolbarsearchfield/example-index.html
and https://main-enterprise.demo.design.infor.com/components/toolbarsearchfield/example-flex-toolbar-align-with-searchfield.html and https://main-enterprise.demo.design.infor.com/components/toolbar/example-selected-event.html
2. Minimize the screen 
3. Then maximize the screen again

**Expected behavior**
'x' button should still be aligned center

**Version**
ids-4.57-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/139046389-a23860f1-2a04-4e37-ad62-32c9cc1c0125.png)
![image](https://user-images.githubusercontent.com/45757016/139047838-cc3caae8-ebb6-41cb-a768-d7109840ec5b.png)

**Platform**
Windows 10 Chrome and all major browsers and OS. 

**Additional context**
N/a

ericangeles:
The only issue that I am seeing is the height of the collapse button on the smaller viewport.

janahintal:
QA Passed. Thanks! 👍 
![image](https://user-images.githubusercontent.com/45757016/141077117-1eaa5211-691b-4a01-9b35-c1da15e9b82a.png)

PR #5809, opened 2021-11-02
5791 - Fix the height of collapse button in non-collapsible searchfield

ericangeles:

This PR fixes the height of the collapse button in the non-collapsible toolbar searchfield on a smaller viewport.

I noticed that this kind of issue is always going back so I created a visual test for it. The issue is either the height of it or the position of the close button.

Closes https://github.com/infor-design/enterprise/issues/5791

- Go to http://localhost:4000/components/toolbarsearchfield/example-flex-toolbar-align-with-searchfield
- Resize the browser to `766px` or below
- Click the `icon search` of Collapsible Searchfield Toolbar Flex
- The searchfield and the collapse button should have the same height
- Click the `icon search` of Non-Collapsible Searchfield Toolbar Flex
- The searchfield and the collapse button should have the same height
- Test in Classic
- Test in different browsers


yohannahbautista:
Looks good! Though I have a question: Is the x button supposed to be that near the edge in the new version? It looks odd when comparing it to the classic one
Screen Shot 2021-11-03 at 5 45 56 PM
Screen Shot 2021-11-03 at 5 46 31 PM

ericangeles:
I am not 100% about that but we can ask @tmcconechy. Is it by design?

tmcconechy:
I think it should be further the right? So maybe need to fix that. Should be the same as the distance of any of the other lookups

yohannahbautista (approved):
Looks good!

tjamesallen15 (approved):

Issue #5509, opened 2021-08-11
ActionSheet: Add "tray" feature


tmcconechy:
**Is your feature request related to a problem or use case? Please describe.**
Design came up with a new feature for action-sheet where the top https://www.figma.com/file/KMCdcFsKVDZjdPFClAS5OF/WFM-Mobile-Components-(Internal)?node-id=1172%3A6097

There are three additions from this design

1. The ability to have it "stuck" on the bottom as a "tray" (call it tray) and when you click it pulls up (call it a header shows)
<img width="861" alt="Screen Shot 2021-08-11 at 1 52 10 PM" src="https://user-images.githubusercontent.com/814283/129079084-6cc0f39e-c22b-40f8-b6ad-6dd8b0d34258.png">

3. The ability to drill into submenus
<img width="683" alt="Screen Shot 2021-08-11 at 1 52 20 PM" src="https://user-images.githubusercontent.com/814283/129079030-f572169b-08c1-4247-850b-1b889ece6c2c.png">

4. The ability to change the color of the tray for status
<img width="773" alt="Screen Shot 2021-08-11 at 1 52 28 PM" src="https://user-images.githubusercontent.com/814283/129079010-e040f718-20b2-4572-86f5-b148c1241958.png">

5. May need swipe up to open... (start with click)

**Describe the solution you'd like**
Add tray, header and submenus to the action-sheet

**Additional context**
https://jira.infor.com/browse/UXD-2120

EdwardCoyle:
@tmcconechy @lucacolumbu Did we ever come to a consensus on the submenus part of this issue?  I vaguely remember discussing the submenus portion of the design and thinking we shouldn't address that immediately?

tmcconechy:
Basically what we said is these arent 100% submenus. They are more of hiding and showing buttons. So the way i would code it in the example is

- Add ALL the links as in the second + third images
- But some are set to hidden
- When you click more... -> it just shows the hidden ones
- If the menu is so big as it takes up the whole view then it gets the X on the top and scrolls

I can demo together if thats easier to explain

CindyMercadoReyes:
This issue is now resolved.

CindyMercadoReyes:
A BDD was created for this ticket. Thanks, Cind
[BDDStory_github5509.docx](https://github.com/infor-design/enterprise/files/7670528/BDDStory_github5509.docx)
y

PR #5861, opened 2021-11-12
5509 - Add actionsheet "tray" feature

ericangeles:

This PR implements a tray feature where this button will stick at the bottom of the page. When hitting the tray button, it will open up the action sheet dialog for the button actions.

Tray will only work on a small viewport (`766px` and below). In desktop view, it will be hidden.

There are couple of settings/options to use such as,
- `tray` to activate the tray button
- `trayOpts.text` to insert text in the tray button
- `trayOpts.icon` to add icon beside the text
- `trayOpts.backgroundColor` can choose in the presets colors `['slate', 'ruby', 'amber', 'emerald', 'azure', 'turquoise', 'amethyst']`. The default value is `slate`

Closes https://github.com/infor-design/enterprise/issues/5509

- Open the test page to mobile or to a smaller viewport `766px` or below http://localhost:4000/components/actionsheet/example-with-tray.html
- You should be able to see the tray button that has the text `Cashier - Clocked In at 8:59 AM` with an icon.
- Click the tray.
- It will show the dialog action sheet.
- Clicking any of the items will close the dialog and pops up a toast message.
- Clicking the trigger button will also open the dialog action sheet.
- Play around with the options like `backgroundColor` and `icon` to see if it's working correctly.


tmcconechy:
Need me to look at this yet @ericangeles ?

ericangeles:
@tmcconechy yes please.

tmcconechy:
@ericangeles can you update the steps? I saw there was a new example
http://localhost:4000/components/actionsheet/example-with-tray.html

I ran it and it looked really good. the only comment i would have is the animation opening and closing doesnt look quite right. It seems like it slides out "seperately". I would work on that a bit.

src/components/actionsheet/actionsheet.js
```diff
@@ -35,6 +42,7 @@ function ActionSheet(element, settings) {
 ActionSheet.prototype = {
   init() {
     this.render();
+    console.log('this', this);
```
tmcconechy, line 45: Remove eventually

ericangeles:
@tmcconechy in desktop/medium screen, I did hide it. It only works on `767px` and below if I am not mistaken? I believe even when you resize it to a bigger screen it will not visible, then toggles back when on a smaller screen.

Thanks for the feedback, Tim! I will work on the animation thing. 👍🏼

tmcconechy:
No thinking about it more i think the desktop behavior is correct. I dont think it should be there then, but we can take feedback when its released.

src/components/actionsheet/actionsheet.js
```diff
@@ -105,9 +112,24 @@ ActionSheet.prototype = {
 
       // Invoke an IDS Button component on each action, if applicable
       $(this.actionSheetElem).find('button').button();
+
+      // Add tray element
+      if (this.settings.tray && breakpoints.isBelow('phone-to-tablet')) {
```
EdwardCoyle, line 117: Maybe this could always render the tray if the setting is enabled, and in `renderTrayElement()` control its visibility based on the breakpoint check

EdwardCoyle (commented):
It's looking great!  Just a comment I have on the behavior so far:

I noticed if you open the page while the viewport is enlarged and shrink the page down, the tray doesn't render.  However it does render when the viewport is small, and hide/show when shrinking/growing works as expected.  Might be fixable below?

Other than that, nice add 👍

src/components/actionsheet/actionsheet.js
ericangeles: I was under impression that the tray will only work on the smaller viewport. So it will be also working on bigger screen sizes?

EdwardCoyle (approved):
Looks great! 👍

tmcconechy (approved):

Issue #3747, opened 2020-04-14
Slider: Showing the tooltip from load


hohnzy:
**Is your feature request related to a problem or use case? Please describe.**
The user is not able to see current value, when the component is loaded. Now the tooltip is seen only after focusing into the component.
![image](https://user-images.githubusercontent.com/11279863/79244484-3db98d80-7e77-11ea-91d1-5556a6c5df8e.png)

**Describe the solution you'd like**
When persistTooltip is set to true, the tooltip should appear immediately after load of the component. Like this:
![image](https://user-images.githubusercontent.com/11279863/79244637-6b063b80-7e77-11ea-98e8-e0c6a1aca6f3.png)

tmcconechy:
This has been discussed. Have to bring up the old conversation i forget at the moment. @EdwardCoyle ?

tmcconechy:
@katieinfor this is an older issue but it requires design feed back. Is the best way to show the value to show it in the tooltip? Maybe there are other options?

katieinfor:
Jira Ticket: https://jira.infor.com/browse/UXD-2909

tmcconechy:
As per https://jira.infor.com/browse/UXD-2909 we should go ahead with this.

a) setting to show the tooltip on page load (basically always shows?)
b) an example with an input next to it showing the value

tmcconechy:
Final designs on https://www.figma.com/file/gHmapzCfCqqPRv5RnFfGcE/Q3-Accessibility-Tests?node-id=44%3A1364

Ready for Dev

jbrcna:
This has been QA tested and passed on v4.59.0-dev.
verified in https://main-enterprise.demo.design.infor.com/components/slider/example-tooltip-onload-and-textbox.html

CindyMercadoReyes:
BDD attached.
[BDDStory_github3747.docx](https://github.com/infor-design/enterprise/files/7737716/BDDStory_github3747.docx)

PR #5874, opened 2021-11-18
3747 - Tooltip on Load

tjamesallen15:

This PR will add support setting for the tooltip to show on load in slider.

Closes https://github.com/infor-design/enterprise/issues/3747

- Go to http://localhost:4000/components/slider/example-tooltip-onload-and-textbox.html
- See the `tooltip`
- Scroll the `sidebar` to see the `textbox` changing value.


tjamesallen15:
I'm not much sure if I got this correctly. Feel free to comment here. I'm thinking at first to make a setting for the textbox but I rather did it only in the example by listening to the events in the slider.

tmcconechy (changes requested):
Can you adjust the centering so its centered as in the design:

Screen Shot 2021-11-18 at 9 09 01 AM

Also the design shows it enabled for entering (but i suppose thats possible either way).

And maybe we can link the input in the component by an id setting so the updating happens as part of the component?

tjamesallen15:
> Can you adjust the centering so its centered as in the design:
> 
> Screen Shot 2021-11-18 at 9 09 01 AM
> 
> Also the design shows it enabled for entering (but i suppose thats possible either way).
> 
> And maybe we can link the input in the component by an id setting so the updating happens as part of the component?

Thank you for the comments @tmcconechy

tmcconechy (commented):
Looks good. But one more thing is i think if you type in the input the expectation is that it would update the slider. Also you could put a number mask on the input.

Thanks James 👍🏻

EdwardCoyle (commented):
@tjamesallen15 This looks really clean!  One thing -- I think before we merge the keyboard usability of the input needs some tweaking:

IDS-Enterprise

Could be we try one of the items below:

tjamesallen15:
Is this okay now, @EdwardCoyle ? :)

EdwardCoyle:
@tjamesallen15 It looks a lot better but I think there's one more piece missing.  When you Tab/Shift+Tab away from the input field, it doesn't commit the value (which I'd expect), but the value in the field becomes mismatched from the value on the slider.  This also happens when the field is blurred, or the user clicks outside the field.  

I think in these cases, the input field should set itself back to the slider's current value.

IDS-Enterprise-_1_

tmcconechy:
@tjamesallen15 i took another look as i saw additions. But the only issue i still see is that if i type in the input and tab it doesnt update the slider.

tjamesallen15:
> @tjamesallen15 i took another look as i saw additions. But the only issue i still see is that if i type in the input and tab it doesnt update the slider.

I've just followed what @EdwardCoyle said that it should revert back to old value. Should I update it once tabbed out?

tmcconechy:
I think we did say the same thing. I.E. just go only in the input and type "50" then hit enter and it updates but go only in the input and type "50" then hit tab and it doesnt

tjamesallen15:
> I think we did say the same thing. I.E. just go only in the input and type "50" then hit enter and it updates but go only in the input and type "50" then hit tab and it doesnt

okay tim! my bad :)

tmcconechy (approved):
LGTM now 🚀

deep7102 (approved):
LGTM 🍌

PR #5898, opened 2021-11-26
5632 - Fix on noSearch dropdown not clearing previous search matches

### LINKED ISSUE #5632: Dropdown: noSearch dropdowns do not clear searchMatches when selecting a different value with mouse or JS
**Describe the bug**
When a noSearch dropdown has values that start with the same character, and the dropdown changes the value by not using the keyboard, the previously typed key is still remembered. That is confusing, but @EdwardCoyle knows what I'm talking about! The steps to reproduce will explain better.

**To Reproduce**

1. Go to /components/dropdown/example-no-search.html
2. Type "t" to select "Two"
3. Use the mouse to select "Five" or use $('#nosearch-dropdown').val('15').trigger('updated') to do it programatically
4. Type "t"
5.  Note that "Three" was selected in the dropdown

**Expected behavior**
"Two" should have been selected

**Version**

- ids-enterprise: 4.51.1

**Platform**

**Additional context**
We have customers hitting this. They had to modify their absence forms to include COVID 19, which they did by prepending C19 to all of their dropdown options. They end up having something like this
* Absence
* C19 - absence
* C19 - extended
* C19 - something else
* Extended
* Something else

After they submit a form, we clear all values to prepare for the next. When a user then goes to that field, and types "C" to selected "C19 - absence", it is actually selecting the next one in the list, causing them to enter a bunch of incorrect data.

The function firstOptionStartingWith in dropdown never clears out searchMatches when selecting other values if not using the keyboard.

yohannahbautista:

The searchMatches aren't cleared after an option is selected, added line of code that deletes current searchMatches if dropdown is noSearch.

Closes #5632 

- Go to: http://localhost:4000/components/dropdown/example-no-search.html
- Type "t" to select "Two"
- Use the mouse to select "Five" or use $('#nosearch-dropdown').val('15').trigger('updated') to do it programmatically
- Type "t"
- "Two" should be highlighted

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - `[Calendar]` Fix the header days where it should be seen when scrolled down. ([#5742](https://github.com/infor-design/enterprise/issues/5742))
 - `[Datagrid]` Tab doesn't go to cells if cellNavigation is false. ([#5734](https://github.com/infor-design/enterprise/issues/5734))
+- `[Dropdown]` Clear search matches after an item is selected. ([#5632](https://github.com/infor-design/enterprise/issues/5632))
 - `[Locale]` Fix issue in parsing date when AM/PM comes first before Hours (a:hh:mm). ([#5129](https://github.com/infor-design/enterprise/issues/5129))
 - `[Modal]` Added option to disable primary trigger on field. ([#5728](https://github.com/infor-design/enterprise/issues/5728))
 - `[Searchfield]` Save input value when searchfield collapses but is not cleared via button click or key. ([#5792](https://github.com/infor-design/enterprise/issues/5792))
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -3235,6 +3235,10 @@ Dropdown.prototype = {
       }
     }
 
+    if (this.settings.noSearch) {
+      delete this.searchMatches;
+    }
+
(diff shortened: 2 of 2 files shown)


ericangeles (dismissed):

EdwardCoyle (changes requested):
@yohannahbautista this does fix the incorrect behavior related to cached searching, but unfortunately it breaks the desired behavior of "cycling" through related values.  Also, the tests that were changed in this PR demonstrate correct behavior needed by one of the teams we support.

The "cycling" behavior occurs when you continuously press the same key while the Dropdown is focused.  For example, if the Dropdown list is closed, pressing the "T" key and waiting should cause the next available item in the list beginning with the letter "T" to become selected.  In the example page, if "Two" is selected, pressing "T" should cause "Three" to become selected next, followed by "Ten", "Twelve", "Thirteen", "Twenty", then cycling back to "Two".  If the dropdown list is opened, the same behavior occurs while keeping the list open (the value isn't committed in this case, just highlighted)

This does currently work on `main` if you want to try: https://main-enterprise.demo.design.infor.com/components/dropdown/example-no-search.html

To get this one finished it needs:
- The tests reverted back
- The search matches should still be cleared, but for example, they should only be cleared for "T" if I try and search for something that isn't "T".  Otherwise, the cycling behavior should remain in-tact.

EdwardCoyle (approved):
🚀  Excellent, works great now

Issue #5719, opened 2021-10-07
SpinBox: Shows ripple effect when disabled


aviralg26892:
**Describe the issue **
Soho Spin Box when we reach to min or max value, disabled buttons stil show the ripple effect making at appear as if its enabled

**Steps to reproduce the behavior:**
1. Go to https://github.com/infor-design/enterprise/blob/4.55.3/app/views/components/spinbox/example-range-limits.html
 https://design.infor.com/code/ids-enterprise/latest/demo/components/spinbox/example-range-limits?theme=uplift&variant=light
2. Click on - or + tabs to reach minimum or maximum value.
3. As we reach min or max value, - or + sections are disabled.
4. Now if we click on disabled area it is showing ripple effect, that eventually leads to click event when applied in project.

**Expected behavior**
On reaching maximum or minimum value, disabled area should not give any response when clicked.

**Version**
- "ids-enterprise-ng": "10.2.1"

**Screenshots**
NA

**Platform**
 - Device (if applicable) NA
 - OS Version: Windows 10 
 - Browser Name :chrome
 - Browser Version: 94.0.4606.71

**Additional context**
Due to this problem when we implement soho spin box in project , even clicking on disabled section triggers to a click event which ultimately leads to enable save feature, even though no changes are done in the component it is used in.

aviralg26892:
Hi @EdwardCoyle , I don't think we have a team here. I am working for Infor and my work id didn't had access to Git earlier. SO I raised issue from my personal Id.

tmcconechy:
@aviralg26892 we just mean whats the name of the app you work on.

aviralg26892:
@tmcconechy  I am working on ION desk application.

nganotice:
This has been QA tested and passed on v4.59.0-dev.

https://main-enterprise.demo.design.infor.com/components/spinbox/example-range-limits

![image0 (1)](https://user-images.githubusercontent.com/84306838/145312695-42bfe8c1-5f86-49c0-962a-40da8c1340d6.gif)

Moving this ticket to Done.

Thanks!

aviralg26892:
Hi @nganotice / @tmcconechy ,

How can I upgrade this in my project? What are the steps and  prerequisites and which angular version is supported in this upgrade?

tmcconechy:
Hi!

[...]

PR #5924, opened 2021-12-06
5719 - Remove ripple effect in spinbox control when it's disabled

ericangeles:

This PR removes the ripple effect in spinbox control when it's being disabled, also added a disabled property to it.

Closes https://github.com/infor-design/enterprise/issues/5719

- Go to http://localhost:4000/components/spinbox/example-range-limits.
- Click on - or + tabs to reach a minimum or maximum value.
- As we reach the min or max value, - or + sections are disabled.
- Clicking the disabled button should not show the ripple effect.
- Ripple effect will only show if the button is not disabled.


tmcconechy (approved):

deep7102 (approved):
🍌

PR #5935, opened 2021-12-10
5632 - Update on clear searchMatches

### LINKED ISSUE #5632: Dropdown: noSearch dropdowns do not clear searchMatches when selecting a different value with mouse or JS

yohannahbautista:

Changed check on clear searchMatches to have it cleared every time an item is selected.

Closes #5632 

- Go to: http://localhost:4000/components/dropdown/example-no-search.html
- Open the top dropdown
- Type '0'
- Select any item besides the 01
- Type '0'
- Should highlight 01


tmcconechy (approved):
Looks good to me

ericangeles (approved):

PR #5949, opened 2021-12-14
1194 - Fix a bug with undefined ids on buttons

tmcconechy:

Due to faulty logic when the toolbar is initialized it is pulling in any buttons with `id="undefined"` when there is a more button with no `aria-controls` on it. This addresses this issue.

TO BE PATCHED in 4.58 once merged

Fixes https://github.com/infor-design/enterprise-ng/issues/1194

- go to http://localhost:4000/components/contextualactionpanel/test-undefined-button
- the page should work as normal - before the button would be moved into the toolbar menu automatically


yohannahbautista (approved):

ericangeles (approved):
🍌

PR #5950, opened 2021-12-15
5793 - Shorten filter delay for single character search entries

### LINKED ISSUE #5793: Dropdown: Tab does not commit a changed value
**Describe the bug**
Clicking on a dropdown, typing a single letter and quickly hitting Tab does not commit the typed letter/option

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/dropdown/test-scenarios.html
2. Click on the 'Set Width Small' dropdown
3. type '7' and quickly follow by pressing TAB

Result:  value in the dropdown is unchanged

**Expected behavior**
Typing a valid value immediately followed by Tab ought to commit the typed value

**Version**

- ids-enterprise: 4.51.1

**Screenshots**

dropdown-quick-tab

**Platform**

**Additional context**

yohannahbautista:

If filter entry is one character and tab is pressed, dropdown will filter list without delay.

Closes #5793 

- Go to: http://localhost:4000/components/dropdown/test-scenarios.html
- Click on the 'Set Width Small' dropdown
- Type '7' and quickly follow by pressing tab
- Value should be selected in dropdown


ericangeles (approved):
Looks good to me!

tmcconechy (changes requested):
@yohannahbautista can the delay go even shorter? or none? im seeing if i do it fast like have two fingers and tap 7 then tab almost at the same time or close it still seems to have the issue. So concerned this isnt going to help fast typers..

s-werking:
Hi @yohannahbautista,

We use noSearch dropdowns in our product. Could this be applied to noSearch dropdowns as well?
I checked out the 5793-dropdown-tab branch and added data-options="{noSearch: true}" as an attribute to the Set Width Small dropdown and the fix doesn't seem to work for this case.

tmcconechy:
I think this one is tough to fix? Im trying the latest and if you hit 7 and then tab fast then it tabs off the page.
Wondering if there is a better solution like ending off the delay and committing the value on the tab key?

yohannahbautista:
> I think this one is tough to fix? Im trying the latest and if you hit 7 and then tab fast then it tabs off the page. Wondering if there is a better solution like ending off the delay and committing the value on the tab key?

@tmcconechy Hello, the current solution I pushed has the dropdown automatically filter when tab is pressed and when the search term is one character. There shouldn't be any delay I think. I press '7' and tab with two fingers as well to test, is this the same way you test it?

tmcconechy:
Yeah thats what im doing.

- click the label on set width small to focus
- put a finger near the 7 and a finger near the tab key
- tap them one after the other but. very fast between

I either see the focus jump to the address bar or to the next field and then back to the same field again

ericangeles:
Is this the bug you're talking about @tmcconechy ?

https://user-images.githubusercontent.com/8839327/147046934-b5e46a48-f432-4aa1-8e54-0b971ab5cda2.mp4

What I did was,
- Clicking the label to focus the dropdown (Set Width)
- Type 7 then tab it very fast

I also noticed that it jumps to the address bar too.

ericangeles:
@tmcconechy, I was wondering, what would be the ideal behavior if these steps were applied?

tmcconechy:
@ericangeles 
yes thats the exact thing im seeing , except for im testing the dropdown below it with 7 in it...(Set Width Small).
i think in this case if the steps are applied then 7 should be saved into the field and the list should close and you should have focus on the next field

ericangeles:
Reviewing it again and it's working correctly now. But am still seeing that it jumps to the address bar, although I test the other example page and it does this too.

yohannahbautista:
> Reviewing it again and it's working correctly now. But am still seeing that it jumps to the address bar, although I test the other example page and it does this too.

@ericangeles What are the steps you do to get this bug? It doesn't focus on the address bar when I do it

ericangeles:
@yohannahbautista same the test I mentioned above, but it's not consistent. Maybe try to speed up a bit when tabbing.

tmcconechy:
@ericangeles maybe your fix on https://github.com/infor-design/enterprise/pull/5950/commits/06e75d98b42ea63b15f7b1727ff8f72e7ad199c4 works as i cant provoke it. Maybe we can pass on to QA now? I'll approve

tmcconechy (approved):

ericangeles (approved):

Issue #5663, opened 2021-09-28
Datagrid: Add option to excel export to use the current locale


NewCoffee1:
**Is your feature request related to a problem or use case? Please describe.**
Hi IDS-Team,

While i'm working with the AppBuilder and the AppHub, i mentioned that the app doesn't take the locale from my browser (as every other application from the CloudSuite does). This is a bigger problem for the export function (to excel). Some formattings are not available (e.g. the Decimal separator (,) and 1000 separator (.) As usual in Germany) and have to be done after the excel export. When i set up different apps, i don't want the user to format the data after export.

Therefore i request you to improve your export functionality to take locale into consideration, so that after that the AppBuilder Team can implement a way for users to change the locale in app builder. 
Right now the export function just takes the data from the grid with no formatting.

The best way would be that the application takes the locale from the browser or i can set it in the app configurations.
I explained the problem in ticket 15714463 on the Infor Concierge Portal.

Please work on this issue because it will increases the attractiveness of the app. Otherwise it becomes unusable in many cases.

Thanks

**Describe the solution you'd like**
Add an option so that the export to excel uses the current locale when exporting dates/numbers.

**Describe alternatives you've considered**
End user has to do it everytime which is a pain.

**Additional context**
AppBuilder and the AppHub

tmcconechy:
When filing issues please use the issue template https://github.com/infor-design/enterprise/issues/new/choose you could edit this issue to form into the markdown template for enhancements?

We do not automatically take the browsers locale as not everyone wants this. But you can easily read the browser locale and call `Soho.Locale.set(navigator.language)`.

I'm also not sure its a very good idea to change the export to use the locale. I would have to be convinced as this is a breaking change. And what we would be doing is formatting dates and numbers to strings. So it would make it difficult to use excel functions on the columns.

I suggest that you just use the excel features to format the numbers and dates.

Will keep this open for now tho for further discussion.

NewCoffee1:
Can i use Soho.Locale.set(navigator.language) i a MultiTenant environment?

It's not possible to format the numbers to a german format in the App. So it's necessary to format it in excel. I think this isn't the way we should work with an CloudSuite. The enduser has to format it every time he uses the export function.

tmcconechy:
The problem is that it takes the unformatted data now. So would have to add an option that takes the data out of the table UI view instead.

jbrcna:
this is now working as expected.
v4.58.0-dev
[...]

PR #5963, opened 2021-12-20
5663 - Updated example page

yohannahbautista:

Updated example page to fix previous bug on locale change.

Addresses comments in #5663 

- Go to http://localhost:4000/components/datagrid/example-export-format.html
- Change locale in dropdown


tmcconechy (changes requested):
@yohannahbautista im ok with the URL changing like this but what im noticing is if you start on http://localhost:4000/components/datagrid/example-export-format.html and pick from the dropdown it doesnt work (when you select from the dropdown nothing happens).

but if you start on http://localhost:4000/components/datagrid/example-export-format.html?locale=en-GB  it works well.

So maybe need a case for default / en-US / no locale?

tmcconechy:
Now on review im not seeing it change the locale. I see an error `gridApi is not defined` in the console. Which is probably why?

tmcconechy (approved):

ericangeles (approved):

PR #5968, opened 2021-12-21
5915: Sink Page: Editor Image not shown

### LINKED ISSUE #5915: Sink Page: Editor Image not shown
**Describe the bug**
when you insert an image in the editor field. it doesn't show

**To Reproduce**

1. Go to https://4580-beta0-enterprise.demo.design.infor.com or https://4580-beta0-enterprise.demo.design.infor.com/components/editor/example-index.html
2. scroll into `Editor`
3. insert image
4. See error;  placeholder image is not shown
5. switch to html view
6. notice that there's an image tag

**Expected behavior**
image

**Version**

- ids-enterprise: v4.58.0-beta0

**Screenshots**
Large GIF (1050x384)

**Platform**
all major browsers

**Additional context**

davidcarlsonberg:

Image is not shown when using image insertion button.

Resolves #5915 

Pull branch, run app, visit http://localhost:4000/components/editor/example-index.html and use the image insertion button and notice the placeholder image should be visible now.


ericangeles:
@davidcarlsonberg can you convert your e2e (protractor) to puppeteer test instead? We are now currently moving all the old tests from protractor to puppeteer. You can check some of the test samples for guidelines and this https://github.com/infor-design/enterprise/blob/main/docs/TESTING.md#puppeteer-e2e-test

test/components/editor/editor.e2e-spec.js
```diff
@@ -97,6 +97,25 @@ describe('Editor example-index tests', () => {
     expect(await element(by.css('.editor img')).isPresent()).toBeFalsy();
   });
 
+  fit('Should insert placeholder image if the image button is clicked and modal confirmed', async () => {
```
tmcconechy, line 100:
```suggestion
  it('Should insert placeholder image if the image button is clicked and modal confirmed', async () => {
```

tmcconechy (commented):
Looks good. You just left in some `fit`s but also agree with @ericangeles comment we should use puppeteer tests as we are working to convert to these

tmcconechy:
Also i think some editor tests might be failing. I restarted the tests to confirm

tmcconechy:
yeah that test is failing https://github.com/infor-design/enterprise/runs/4608596433?check_suite_focus=true#step:12:3008

tmcconechy:
An idea for a test fix here is to make a new page for automation-id's then make this test point to that page instead. Or try and correct the logic so it works if the ID is changed with automation id's...

tmcconechy:
Rechecking - still see failing tests but fixed the conflicts for you

tmcconechy (changes requested):
Looking at this again @davidcarlsonberg i think we should maybe try to fix the logic so that it works if they do change the setting as this could happen.

davidcarlsonberg:
Apologies but I am not receiving your @ mentions to me in GitHub for some reason. I'm not sure what you're referring to when you mention Automation IDs?

tmcconechy:
@davidcarlsonberg by the automation id's i mean that when using the `attributes` setting the developer can assign an id or a `data-automation-id` attribute or whatever they like. This case is really just around the ID. So thinking something like..

- get the current id whatever it is however it is set (if you need to can check settings.attributes['id'] is there...)
- use that in the logic

davidcarlsonberg:
In the latest commit I'm grabbing the ID from attributes. Is that what you had in mind? I can also grab the last item in the attributes array if we want to use whatever the last item is?

tmcconechy:
yeah i think that makes sense 👍🏻 the final proof is in if the tests do pass there then i think it should be good

davidcarlsonberg:
Oof. I've run the tests a couple times and the only consistent error is the Blockgrid one. I am unsure how that is related to the work done in this PR?

ericangeles:
@davidcarlsonberg maybe it has something to do with the chrome version. @tmcconechy made the fixed and it's now merged. Is your branch updated with the main?

davidcarlsonberg:
Ah! Good to know. Updating to `main` now.

tmcconechy:
Yes i fixed this one on a PR recently merged 👍🏻  So update with main should work

tmcconechy (dismissed):

tjamesallen15 (dismissed):

Issue #5611, opened 2021-09-14
Accordion: Mobile Enhancements and style changes


tmcconechy:
**Describe the bug**
Make some style changes to accordion examples and make it work better on mobile. Based on Demo
https://main-enterprise.demo.design.infor.com/components/accordion/example-accordion-panels.html

**Fixes**
https://www.figma.com/file/36JA2H9Z6zRnVuj5A6BfdO/Accordion?node-id=0%3A1

- [ ] 44px hitboxes (non visible/size changing)
- [ ] 16px spacing

**Additional context**
https://jira.infor.com/browse/UXD-2767

tjamesallen15:
Hello @tmcconechy ,

I don't have an access for this one. I'm using my infor email
![image](https://user-images.githubusercontent.com/34049950/146871364-fb33a4ec-e671-4fdf-8abc-006d8fe07451.png)

tmcconechy:
I Requested on https://jira.infor.com/browse/IDS-136 @tjamesallen15

CindyMercadoReyes:
This issue is now resolved.

PR #5975, opened 2021-12-27
5611 - Accordion Mobile Enhancements

tjamesallen15:

This PR will adjust spacing and hitboxes in accordion for mobile enhancements.

Closes https://github.com/infor-design/enterprise/issues/5611

- Go to http://localhost:4000/components/accordion/example-accordion-panels.html
- Open `Developer Console`
- Inspect the `Spacing`
- Inspect  the `Hitboxes`


tjamesallen15:
Will fix the visual regression issues once this PR is approve.

ericangeles (changes requested):
I would rather extend the `hitbox` setting in the button component so it has the option to use it or not. You can check this example page 

http://localhost:4000/components/button/example-button-with-hitbox.html

And make adjustments in the accordion component styling if it's needed. :) Let me know if you need help.

tmcconechy (commented):
I think this would be ok. As long as it doesnt change the height from the old version https://main-enterprise.demo.design.infor.com/components/accordion/example-accordion-panels.html we could keep the hitboxes in the component.

It makes for less settings. The only question i have is that the spacing on the left to the icon vs the spacing to the button doesnt look equal anymore (left and right side). I didnt check but does the design take this into account?

tjamesallen15:
> I think this would be ok. As long as it doesnt change the height from the old version https://main-enterprise.demo.design.infor.com/components/accordion/example-accordion-panels.html we could keep the hitboxes in the component.
> 
> It makes for less settings. The only question i have is that the spacing on the left to the icon vs the spacing to the button doesnt look equal anymore (left and right side). I didnt check but does the design take this into account?

The height is still okay and wasnt change. I did adjust the the buttons for spacing.

tmcconechy (approved):
Personally i think this looks good. For this case i dont think we need a specific setting as it can fit and that way less options to worry about 👍🏻 

If @ericangeles agrees we can just need the tests fixes

ericangeles (dismissed):
Approving! Thanks, @tjamesallen15 !

tjamesallen15:
Will do the visual regression in a while.

tmcconechy (approved):

ericangeles (approved):

Issue #5875, opened 2021-11-18
Modal: Focus issues with fields with inline style of display:none.


JBismonte:
**Describe the bug**
A modal dialog with auto focus is not ignoring input fields are hidden by the inline style display:none. also the mutation observer only recognize the change in focusable elements only when there is something added or removed to the dom, it doesnt update when you change the input elements attributes (tabindex, style etc.) dynamically.

**To Reproduce**

1. replace the example-fullsize-modal.component.html file from here. [example-fullsize-modal.component.zip](https://github.com/infor-design/enterprise/files/7560934/example-fullsize-modal.component.zip)
2. Go to http://localhost:4200/ids-enterprise-ng-demo/modal-dialog
3. Click on full on tablet button
4. it should be focused on the name field upon open of the focus modal.

**Expected behavior**
update/add the following selector when updating the focusable element in DOM.focusableElems() to not include inputs hidden by the inline style display:none, need to also account if there is space between the display:none.
```
'input:not([tabindex="-1"]):not([disabled]):not([style*="display:none;"]):not([style*="display: none;"])'
```

also can we modify the MutationObserver inside the modal.js to include attribute changes like tabindex or inline style?

```
      // When changes happen within the subtree on the Modal, rebuilds the internal hash of
      // tabbable elements used for retaining focus.
      self.changeObserver = new MutationObserver(() => {
        self.setFocusableElems();
      });
      self.changeObserver.observe(self.element[0], 
        { childList: true,
          subtree: true,
          attributes: true // add this argument to trigger the rebuild of internal hash when elements have attribute changes.
```

**Version**

- ids-enterprise: 4.55.3

**Screenshots**
https://user-images.githubusercontent.com/29673378/142371592-95f52681-2900-49cb-8199-3ac6aad74eb4.mp4

**Platform**
 - Infor Application/Team Name: Landmark
 - Device: (if applicable) : NA
 - OS Version: [e.g. Windows 10 or iOS 8]: Windows 10
 - Browser Name: [e.g. chrome, safari, stock browser]: Chrome
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]: Version 95.0.4638.69 (Official Build) (64-bit)

**Additional context**
https://jira.lawson.com/browse/LMCLIENT-38382

janahintal:
QA Passed. Thanks. 
[...]

PR #5979, opened 2021-12-28
5875 - Fix on Modal Focus with Hidden Display None

tjamesallen15:

This PR will fix modal focus issues with inline display none.

Closes https://github.com/infor-design/enterprise/issues/5875 

- Go to http://localhost:4000/components/modal/example-hidden-field.html
- Click `Show Modal`
- Tab around the modal - should skip the hidden field

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -1040,7 +1040,7 @@ Modal.prototype = {
       self.setFocusableElems();
 
       let focusElem = $(self.focusableElems).not('.modal-header .searchfield')
-        .not('[style*="display: none"]').first();
+        .not(':hidden').first();
       if (focusElem.length === 0) {
         focusElem = thisElem.element.find('.btn-modal-primary');
       }
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -1039,7 +1039,8 @@ Modal.prototype = {
       self.changeObserver.observe(self.element[0], { childList: true, subtree: true });
       self.setFocusableElems();
 
-      let focusElem = $(self.focusableElems).not('.modal-header .searchfield').first();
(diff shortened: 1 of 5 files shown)


src/components/modal/modal.js
```diff
@@ -1039,7 +1039,8 @@ Modal.prototype = {
       self.changeObserver.observe(self.element[0], { childList: true, subtree: true });
       self.setFocusableElems();
 
-      let focusElem = $(self.focusableElems).not('.modal-header .searchfield').first();
+      let focusElem = $(self.focusableElems).not('.modal-header .searchfield')
+        .not('[style*="display: none"]').first();
```
tmcconechy, line 1043:
```suggestion
        .not(':hidden').first();
```

Will this work? It could catch more cases like it is invisible with css or css visibility. Lets try to see if it solves?

tmcconechy (changes requested):
Maybe the suggested selector is better. Also can you update the test steps on the PR. I see a new example but wasnt sure how to test

tmcconechy (changes requested):
I see some problems now with this @tjamesallen15 

1. open the modal
2. focus problem then tab multiple times

Notice that the tab stops on the save button and doesnt loop around anymore to the top of the modal. See https://main-enterprise.demo.design.infor.com/components/modal/example-index.html

Also the reverse is wrong
1. open model
2. focus problem field
3. shift tab

Notice it tabs out of the modal. Instead of looping around.

I guess the logic of finding the first field on the form is also wrong or the fix needs to be adjusted?

tmcconechy (approved):
Looks good now 👍🏻

deep7102 (approved):

EdwardCoyle (approved):

Issue #5743, opened 2021-10-14
Calendar event colors should accept color keywords


lbursey-infor:
**Is your feature request related to a problem? Please describe.**
Calendar events currently only accept a hex value color (eg "#00FF00") when overriding the color of an event. Currently we get custom colors for calendar events both by hex and by keyword (eg "papayawhip").

**Describe the solution you'd like**
Calendar event should accept color keywords as well, instead of just accepting hex values.

**Describe alternatives you've considered**
WFM is having to maintain an array of color keywords and convert them to hexes, but this is not ideal.

lbursey-infor:
Hi Tim,

Current workaround is to maintain a map of all color keywords and their hex values (ie this list https://www.w3.org/wiki/CSS/Properties/color/keywords).

Regarding the color palette, our customers have custom colors already set up for certain events (eg vacation/PTO/sick day) and they need to be consistent across the app. That might mean some inconsistency in branding but it's required for some use cases.

jbrcna:
The colors on the event types displayed in the calendar is also reflected in the legend. However, if you create a new event it will not match the legend color of event.
 v4.60.0-dev
Steps 

1. Go to https://main-enterprise.demo.design.infor.com/components/calendar/test-event-custom-colors.html
2. Create a new event (Add event Modal)
3. See error.
4. Try all event types
5. Noticet that the color is always `background-color: powderblue; border-left-color: royalblue;`

CindyMercadoReyes:
BDD attached
[BDDStory_github5743.docx](https://github.com/infor-design/enterprise/files/7927352/BDDStory_github5743.docx)
.

ericangeles:
@tjamesallen15 where are we on this?

jbrcna:
this is now working as expected. v4.60.0-beta1 . thank you
https://4600-beta1-enterprise.demo.design.infor.com/components/calendar/test-event-custom-colors.html
<img width="1459" alt="image" src="https://user-images.githubusercontent.com/56722184/153340520-2b7badb1-bec3-4c0b-82ca-2c20fd2260de.png">

PR #5981, opened 2021-12-29
5743 - Calendar Events Accept Color Keywords

tjamesallen15:

This PR will add a setting in the Calendar for Custom Colors.

Closes https://github.com/infor-design/enterprise/issues/5743

- Go to http://localhost:4000/components/calendar/test-event-custom-colors.html
- See the colors

### DIFF
--- a/app/views/components/calendar/test-event-custom-colors.html
+++ b/app/views/components/calendar/test-event-custom-colors.html
@@ -1,0 +1,312 @@
+<div class="calendar" data-init="false">
+  <div class="calendar-events">
+    <div class="accordion" data-options="{'allowOnePane': false}">
+      <div class="accordion-header is-expanded">
+        <a href="#"><span data-translate="text">Legend</span></a>
+      </div>
+      <div class="accordion-pane">
+        <div class="calendar-event-types accordion-content">
+        </div>
+      </div>
+      <div class="accordion-header is-expanded">
+        <a href="#"><span data-translate="text">UpComing</span></a>
+      </div>
+      <div class="accordion-pane">
+        <div class="calendar-upcoming-events accordion-content">
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="calendar-monthview">
+  </div>
+  <div class="calendar-weekview">
+  </div>
+  <div class="calendar-event-details accordion" data-init="false" data-options="{'allowOnePane': false}">
+  </div>
+  <div class="calendar-event-details-mobile listview" data-init="false">
+  </div>
+</div>
+
+<ul id="calendar-actions-menu" class="popupmenu">
+  <li><a href="#" data-action="delete-event"><span data-translate="text">DeleteEvent</span></a></li>
+  <li><a href="#" data-action="show-event"><span data-translate="text">ShowEvent</span></a></li>
+</ul>
+
+<script>
+$('body').one('initialized', function () {
+  // Get the Event Type and Events to show in the calendar
+  var eventTypes = [];
+  var events = [];
+
+  $.getJSON('{{basepath}}api/event-types', function(res) {
+    eventTypes = res;
+
+    $.getJSON('{{basepath}}api/events', function(res) {
+      events = res;
+      $('.calendar').calendar({
+        month: 6,
+        year: 2020,
+        day: 15,
+        eventTypes: eventTypes,
+        disable: {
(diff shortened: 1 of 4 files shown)


tmcconechy (changes requested):
@tjamesallen15 I think this isnt 100% what was expected.

1. Should change the events to a custom color


Screen Shot 2022-01-03 at 10 25 59 AM

Not the days.

1. We might not need a setting for this. I think what would work is adding a new data set with a custom color like https://github.com/infor-design/enterprise/blob/main/app/data/event-types.json#L2

- test setting color: #80000 or color: papaywhip
- might need a setting for the borderColor as well in the event types


Is that all correct @lbursey-infor

tjamesallen15:
> @tjamesallen15 I think this isnt 100% what was expected.
> 
> * Should change the events to a custom color
> 
> Screen Shot 2022-01-03 at 10 25 59 AM
> 
> Not the days.
> 
> * We might not need a setting for this. I think what would work is adding a new data set with a custom color like https://github.com/infor-design/enterprise/blob/main/app/data/event-types.json#L2
> * test setting color: #80000 or color: papaywhip
> * might need a setting for the borderColor as well in the event types
> 
> Is that all correct @lbursey-infor

Oh my bad on this one! Will do it.

lbursey-infor:
> Is that all correct @lbursey-infor

That is correct, thanks @tmcconechy

tjamesallen15:
I cant fix the e2e test. It shows chromedriver issue.

tmcconechy:
I tried to release the caches if this passes https://github.com/infor-design/enterprise/actions/runs/1660349835 then just pull main on.

If that doesnt work need to make the opposite of this change for that version (you can do it on your branch and try) i.e.e add the version back in https://github.com/infor-design/enterprise/commit/122607e25113e57f40c88264a89ea7fcbaa7c778#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

Otherwise i'll follow up again tomorrow. cc @ericangeles

tjamesallen15:
Waiting for this one to be merge. https://github.com/infor-design/enterprise/pull/5993

tmcconechy (dismissed):
Approved assuming tests pass (merged the fix on)

tmcconechy (approved):

EdwardCoyle:
@tjamesallen15 @tmcconechy are the colors here still supposed to match up with the calendar's legend?  I noticed you can still click the checkmarks on the legend and they still have an effect on the display of the custom events... not sure if that's right?

tmcconechy:
yeah your right @EdwardCoyle i missed that. since the data is in the eventTypes the legends should also show that color...
i think lets request changes?

EdwardCoyle (changes requested):
Yep, sorry for the delay here @tjamesallen15.  The colors on the event types displayed in the calendar should also be reflected in the legend:

Screen Shot 2022-01-07 at 1 28 00 PM

tjamesallen15:
> 

question for @tmcconechy @EdwardCoyle ,

i put them as type DTO in the events.json. Should I add just another DTO but in another color? I checked the eventTypes.json and it was using type as id, for ex. (dto, sickleave)

tmcconechy:
@tjamesallen15 i think we should maybe add a new `eventsCustomColors.json` and use that in just the one color example so the main examples are not affected 👍🏻

tjamesallen15:
> @tjamesallen15 i think we should maybe add a new `eventsCustomColors.json` and use that in just the one color example so the main examples are not affected 👍🏻

Will do this one!

tmcconechy (commented):
I think this looks right now except..

1. can we rename customEvent-Types to maybe `custom-event-colors.json`
2. is there an option to set the `borderColor` as well for the 2 pixel border on the right as the way that color looks for the border could look better. So maybe set that in the data and set it to some color a bit darker than the custom color so it looks nicer

tmcconechy (commented):
I changed one file to kebab-case. But also Im still not seeing the legend colors match (i think they did at one point though if that helps)

Screen Shot 2022-01-13 at 9 29 37 AM

tmcconechy (approved):
cc @EdwardCoyle

ericangeles (approved):

EdwardCoyle (approved):
Looks good, thanks @tjamesallen15 👍

PR #5984, opened 2022-01-03
5846 - Fileupload fix in datagrid

### LINKED ISSUE #5846: Datagrid: File upload editor not working
**Describe the bug**
The example for file upload is not currently working

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/datagrid/test-editable-fileupload.html 
2. Click on first column and select a file
3. See the file is not shown

**Expected behavior**
Should work as before

**Version**
- ids-enterprise: [e.g. v4.56]

**Additional context**
Originated at Ms Teams.
@deep7102 took a brief look and the value is not there https://github.com/infor-design/enterprise/blob/c83478ca43dc69be8f88f927dcc03d7eb8d3625b/src/components/datagrid/datagrid.editors.js#L817 so it always come blank .

If is to maybe so I try to put timeout around 200ms and it show the value

```
setTimeout(() => {
   console.log('input value', this.input[0].value);
}, 200);
```

yohannahbautista:

Added a promise that puts a timeout before getting the value so that value won't be undefined when it reaches datagrid. 

Closes #5846 

- Go to: http://localhost:4000/components/datagrid/test-editable-fileupload.html
- Click on the file upload button on the first column and select a file
- File path should appear on column


tmcconechy (changes requested):
1. This seems to have broken / impacted a lot of tests
1. This seems to work now but i see some improvements that can be made on the UI side
- when clicking the input the button is hidden (try clicking the order date trigger and see how it shows still while loading) - this one i could skip but i think there is a way to keep the editing class on.
- the alignment is off (try all row heights)
Screen Shot 2022-01-03 at 10 32 31 AM

Issue #5842, opened 2021-11-10
Dropdown: Create a Puppeteer Script for Enter key opens dropdown list, when it should only be used to select items within an open list


maezcruz:
Describe the Issue
Create a unit testing for Enter key opens dropdown list, when it should only be used to select items within an open list.

See attached BDD for more information.
[story_github#1239.docx](https://github.com/infor-design/enterprise/files/7512787/story_github.1239.docx)

Related github issue
#1239

maezcruz:
cc: @CindyMercadoReyes

janahintal:
Test covered in #6000. This ticket can now be closed. Thank you.

PR #6000, opened 2022-01-07
5842 - Add puppeteer tests for dropdown custom keystrokes

janahintal:

Dropdown puppeteer test for #5842 

1. pull this branch
2. build and run the app
3. run `npm run e2e:puppeteer dropdown`
4. It should pass.

Steps:
Scenario 1:
Run http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html
Select the dropdown via tab key.
Press enter key
It should not open the list

Scenario 2:
Run http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html
Select the dropdown via tab key.
Open the dropdown list
Hit enter on a highlighted value
Enter key should work even the enter key stroke doesn't allow


test/components/dropdown/dropdown.puppeteer-spec.js
```diff
@@ -81,4 +81,30 @@ describe('Dropdown Puppeteer Tests', () => {
       expect(newVal).toEqual('Validated Dropdown, Required'); // Required text should append to the aria-label
     });
   });
+
+  describe('Dropdown for Enter key opens dropdown list, when it should only be used to select items within an open list', () => {
```
tmcconechy, line 85:
```suggestion
  describe('Supports custom keystrokes', () => {
```

test/components/dropdown/dropdown.puppeteer-spec.js
```diff
@@ -81,4 +81,30 @@ describe('Dropdown Puppeteer Tests', () => {
...
+    const url = 'http://localhost:4000/components/dropdown/test-allow-custom-keystroke.html';
+
+    beforeAll(async () => {
+      await page.goto(url, { waitUntil: ['domcontentloaded', 'networkidle0'] });
+    });
+
+    it('should have the enter key stroke to be disabled  ', async () => {
```
tmcconechy, line 92:
```suggestion
    it('should support disabling the enter key when closed with the onKeyDown function', async () => {
```

test/components/dropdown/dropdown.puppeteer-spec.js
```diff
@@ -81,4 +81,30 @@ describe('Dropdown Puppeteer Tests', () => {
...
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Enter');
+      const ariaEx = await page.$eval('div[class="dropdown"]', el => el.getAttribute('aria-expanded'));
+      expect(ariaEx).toMatch('false');
+    });
+
+    it('should select value when hovered and press enter', async () => {
```
tmcconechy, line 100:
```suggestion
    it('should support using the enter key to select while using the onKeyDown function', async () => {
```

tmcconechy (changes requested):
Just edited some of the test names for clarity. Feel free to adjust as you see fit. Also check out this article that i agree with https://github.com/mawrkus/js-unit-testing-guide#-name-your-tests-properly

ericangeles:
> Just edited some of the test names for clarity. Feel free to adjust as you see fit. Also check out this article that i agree with https://github.com/mawrkus/js-unit-testing-guide#-name-your-tests-properly
> 
> Also the one test seems to fail still

We can apply this approach. Cc @tjamesallen15 @yohannahbautista see the link.

test/components/dropdown/dropdown.puppeteer-spec.js
janahintal: > should support using the enter key to select while using the onKeyDown function

Noted Tim. Working on this. It's weird that the script is failing in headless. Tried it running in Non-headless and it all passed. Still working for a workaround. Thanks!

tmcconechy: @janahintal you could try to do something like this:
```
const element = await page.waitForSelector('div[class="dropdown"], { visible: true });
````      

So it waits for that selector to load. Also this is much more preferable to all the `await page.waitForTimeout(200)` as each of these is slowing down the tests the `waitForSelector` will only wait as long as needed.

ericangeles:
@janahintal I pushed a better fix for getting the aria attributes. Let me know if you still need help.

ericangeles (commented):
A changelog perhaps? @janahintal

tmcconechy:
Looks good now. I like that you fixed the timeouts to do the waitForselector instead. WIll approve if this all passes

ericangeles (approved):

Issue #5937, opened 2021-12-10
Model: Dialog containing a rich text editor has a row height issue


claudenbach:
**Describe the bug:**
When rapidly typing into the rich text editor that is in a modal dialog, the line height of the rows is being changed when the user pauses from typing.

**Steps to reproduce the behavior:**
1. Using https://main-enterprise.demo.design.infor.com/components/editor/test-modal.html
2. Click on Show Modal
3. Clear the existing content with the rich text component
4. Rapidly type into the text editor several characters then the enter key (repeat for a couple of lines) then stop

Notice that the height of the rows within the editor are changed when the user pauses from typing.

I've only noticed this behavior happening when the editor is contained in a modal dialog.

**Version**
 ids 4.59.0-dev
 
 **Additional Context:**
 Have a customer reporting this issue.  The user is doing heads-down data entry, rapidly typing into the editor.

claudenbach:
Example from components/editor/test-modal.html:

![RichTextIssue](https://user-images.githubusercontent.com/39804038/145632645-5c74dfe7-131e-4912-9b8b-49549bf494ca.gif)

claudenbach:
Landmark example:
![LMCLIENT-38399](https://user-images.githubusercontent.com/39804038/145632862-9c4ac537-ddf8-4036-b0a2-844ff350730f.gif)

tmcconechy:
Tips to someone that picks this up:

To clarify we added a setting so it will always wrap in P tags so each time you hit enter it makes a new <p> tag which has space between each one. If you dont want a P tag you can do shift+enter.

But it seems in this one case the container isnt yet wrapped possibly the modal is prevent this from happening https://github.com/infor-design/enterprise/blob/main/src/components/editor/editor.js#L650

tjamesallen15:
@tmcconechy  I tried this at http://localhost:4000/components/editor/example-index.html and it seems behaving exactly with the one inside the modal.

tmcconechy:
@tjamesallen15 i did see this issue - sounds like from our teams chat you were eventually able to reproduce?

tjamesallen15:
> @tjamesallen15 i did see this issue - sounds like from our teams chat you were eventually able to reproduce?

yes. still looking at it.

jbrcna:
this is now working as expected. moving this ticket to Done. thanks
v4.56.0-dev
https://main-enterprise.demo.design.infor.com/components/editor/example-index.html
[...]

PR #6050, opened 2022-01-21
5937 - Rich Text Editor Row Height Issue after typing

tjamesallen15:

This PR will fix a bug which changes the approach intended by the user after typing in editor.

Closes https://github.com/infor-design/enterprise/issues/5937

- Go to http://localhost:4000/components/editor/example-index.html
- Clear the existing content with the rich text component
- Rapidly type into the text editor several characters then the enter key (repeat for a couple of lines) then stop
- Go to http://localhost:4000/components/editor/test-modal.html
- Click on `Show Modal`
- Do the same as the above.


tmcconechy (changes requested):
Sorry i think i see why that code might be there. But maybe there is another place we can add it to update the buttons.

http://localhost:4000/components/editor/example-index.html
- select Cross and make it bold (notice the B button is highlighted)
- select platform  (notice the B button is highlighted still when its not bold)
- make it italic (correct now)
- select Cross again  (notice the I button is highlighted still when its not bold)

Compare to https://main-enterprise.demo.design.infor.com/components/editor/example-index.html

I think maybe some other event logic when selecting still needs to call checkSelect. Either that or we could consider removing the fact it highlights the button? But that seems like it may get pushback.

tjamesallen15:
I brought back the bindSelect but with little bit of twist. @tmcconechy

EdwardCoyle (approved):

ericangeles (approved):

Issue #6083, opened 2022-02-02
Tabs: 'x' icon is not aligned in responsive view


janahintal:
**Describe the bug**
Tabs: 'x' icon is not aligned in responsive view

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/tabs/example-enable-disable.html?theme=classic&mode=light&colors=2578a9
2. Minimize the screen to responsive view

**Expected behavior**
'x' button / icon should be aligned

**Version**
ids-4.60-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/45757016/152144181-e8508ccc-d007-4c29-ba72-2d8c178b631c.png)

**Platform**
All major browsers, OS and devices.

**Additional context**
N/A

janahintal:
QA Failed. Tested in https://main-enterprise.demo.design.infor.com/components/tabs/example-enable-disable.html?theme=classic&mode=light&colors=2578a9
![image](https://user-images.githubusercontent.com/45757016/153198200-b7a1c3c5-8b24-4d28-828a-6b27a50f61ad.png)

tmcconechy:
We had some issues with tests so this wasnt deployed. Please try again

janahintal:
QA Passed. Will now move this ticket to Done. Thanks!
![image](https://user-images.githubusercontent.com/45757016/153392857-6ff38dab-c00c-49f5-b051-bc203ff3fba0.png)

PR #6097, opened 2022-02-06
6083 - Tabs: 'x' icon is not aligned in responsive view

davidcarlsonberg:

This is an issue with the popupmenu associated with tabs overflow.
Close icon should be properly aligned in the popupmenu.

#6083 

- Pull, build, run
- Visit http://localhost:4000/components/tabs/example-enable-disable.html?theme=classic&mode=light&colors=2578a9
- Emulate a responsive environment
- Pull down dropdown (popupmenu) with tab overflow and notice "Dismissible" close icon is vertically centered and right aligned.

- [ ] ~An e2e or functional test for the bug or feature.~

### DIFF
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -116,7 +116,7 @@
               padding-right: 6px;
               position: absolute;
               right: 1px;
-              top: 0;
+              top: 1px;
               transform: translateY(50%);
             }
           }
--- a/src/components/popupmenu/_popupmenu.scss
+++ b/src/components/popupmenu/_popupmenu.scss
@@ -112,12 +112,12 @@
           .icon {
             &.close {
               height: 15px;
-              margin-top: -7.5px;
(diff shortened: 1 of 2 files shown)


src/components/popupmenu/_popupmenu.scss
```diff
@@ -112,11 +112,12 @@
           .icon {
             &.close {
               height: 15px;
+              margin-top: -7.5px;
```
ericangeles, line 115: Can we use `transform: translateY(-50%)` instead `margin-top: -7.5px`? I think it's much accurate than using px value with a decimal value.

tmcconechy:
Screen Shot 2022-02-07 at 9 49 33 AM

A bit of a knitpick but i think it would look better if it was over to the left a touch. Also in RTL mode it should be on the other side http://localhost:4000/components/tabs/example-enable-disable.html?theme=classic&mode=light&colors=2578a9&locale=ar-SA

deep7102 (approved):

Issue #6041, opened 2022-01-19
Contextual Action Panel: Toolbars inside the body of the panel are being displaced


aaronpikkarainen:
**Describe the bug**
All toolbars that exist in the body of the CAP are being moved to the CAP's header. This behavior is unwanted and it is corrupting the display of the CAP by giving it multiple header toolbars. Only the first toolbar in the CAP should be treated as the CAP's header toolbar. Any other toolbars in the CAP should be left in their place. There can be many good reasons for having toolbars in a CAP's body (i.e. one or more grids with their toolbars in the CAP).

**To Reproduce**
The only IDS example I could find of a CAP with a toolbar in the body is https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/test-contextualactionpanel.html  This seems to intentionally move the toolbar for this specific use case of a grid CAP. But there can be many use cases where we have a grid toolbar inside a CAP and we don't want the grid's toolbar moved up into the header. The below solution will fix the problem and still allow for this particular use case of moving the grid's toolbar up into the header (if no other header toolbar exists).

**Solution**
The coding fix for this bug is simple and safe. Only 2 lines of code need tweaked. Please change contextualactionpanel.js so that it only looks for the **first** toolbar and buttonset instead of finding multiple of them...

**line 265:**
`let buttonset = this.panel.find('.toolbar .buttonset, .flex-toolbar .buttonset').first();`

**line 269:**
`this.toolbar = this.panel.find('.toolbar, .flex-toolbar').first();`

**Expected behavior**
The CAP should only have one header toolbar, and all other toolbars should be left where they are.

**Version**
- ids-enterprise: v4.59.1 and earlier

**Platform**
 - Infor Application/Team Name: CSD
 - Device: laptop
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 97.0.4692.71 (Official Build) (64-bit)

Thank you!

tmcconechy:
@aaronpikkarainen i think the change makes sense but i tried to reproduce it and your lines didnt actually fix it.. This is the change i made to test-contextualactionpanel.html

```

<div class="row">
  <div class="six columns">

    <button type="button" id="js-contextual-panel" class="btn-secondary contextual-action-panel-trigger no-init">
      Open Panel with Grid
    </button>
    <div class="contextual-action-panel" style="display:none">
      <div class="full-width">
        <div class="flex-toolbar" role="toolbar">
          <div class="title">
[...]

PR #6135, opened 2022-02-21
6041 - Fix UI issues in CAP when using the toolbar inside of the CAPs body

ericangeles:

This PR fixes UI issues where the toolbar was being moved to the CAP's header instead of staying in its original position.

Closes https://github.com/infor-design/enterprise/issues/6041

- Go to http://localhost:4000/components/datagrid/test-cap-with-toolbar-inside-of-the-body-panel.html
- Open the CAP
- Toolbar in the body should not move to the header toolbar section, it should retain in its position


yohannahbautista (approved):

src/components/contextualactionpanel/_contextualactionpanel-new.scss
```diff
@@ -11,3 +11,15 @@
...
   }
 }
+
+.modal.contextual-action-panel.is-visible {
+  .modal-body-wrapper {
+    .flex-toolbar {
+      padding: 0 5px 5px;
+
+      .title h2 {
+        font-size: 1.8rem;
```
tmcconechy, line 21: Looks good! But wondering why this? I would think it would be 2.2rem like the others but on the UI it looks the same?

ericangeles: I adjusted because of the request in the issue?

tmcconechy (approved):
ok, thanks i missed that. I see it now. I guess since its rems it was getting bumped up

tjamesallen15 (approved):

PR #6147, opened 2022-03-01
6024 - Changed maximum modal width

### LINKED ISSUE #6024: Modal: Dialog content does not fill the space inside modal dialog properly (extra space on the right side) when window size is under 481px
**Describe the bug**

Elements inside a modal dialog that are set to use the full available width inside the modal dialog (i.e. full width of div.modal-body-wrapper minus paddings) suddenly get an extra space on the right side once the window width is shrunk under 481px (i.e. 480px and lower). 

This happens because max-width of div.modal-body-wrapper is set to (window width - 46px) while the parent element div.modal-content width is (window width - 34px), so there is a 12px gap on the right. Thus div.modal-body-wrapper does not use all the available width it's supposed to use.

The underlying cause is that there is a style for .modal (for context: div.modal > div.modal-content > div.modal-body-wrapper) applied when window width is above 480px (@media screen and (min-width: 481px)) that sets max-width of the div.modal to 90%. 

Once the window width goes under 481px, this style is no more applied, but rather the default style for .modal is applied where max-width is set to 100%

The max-width of div.modal-body-wrapper is calculated in the Modal.resize() function. In the case of div.modal having max-width set to 100%, the calculation goes wrong and sets the max-width of div.modal-body-wrapper to a lower value than actually needed.

**To Reproduce**

1. In Chrome, go to 'Modal', e.g. https://design.infor.com/code/ids-enterprise/latest/modal
2. Click on 'View Example' blue icon on the right above the Modal Example area, which takes you to https://design.infor.com/code/ids-enterprise/latest/demo/components/modal/example-index?theme=uplift&variant=light
3. Click on the 'Show Modal' button
4. Open DevTools via F12 and click on the 'Toggle device toolbar' button in the top left
5. Set Dimensions to Responsive and set 481px for width
6. Tweak the size of the 'Problem' input in DevTools so it occupies the whole available width (i.e. width: 100%;). Observe that the input is centered symmetrically inside the modal
7. Change the width to 480px. Observe that the gap on the right is larger than the gap on the left. Hover over div.modal-body-wrapper in DevTools -> Elements and observe 12px extra "margin" highlighted on the right. This is because of the max-width of div.modal-body-wrapper set to a lower value than the parent div.modal-content can accomodate

**Expected behavior**

The content set to fill all the available width inside modal should be centered correctly, regardless of the window width being 481px+ or less.

**Version**

- ids-enterprise: 4.59.0 (initially found on 4.57.0)

**Screenshots**

modal480px

**Platform**

 - Infor Application/Team Name: Infor Workforce Management
 - OS Version: Windows 10
 - Browser Name: chrome
 - Browser Version: 97.0.4692.71 (Official Build) (64-bit)

yohannahbautista:

Maximum width in modal is set to full width so that there won't be a gap on the right side.

Closes #6024 

- Go to: http://localhost:4000/components/modal/example-index?theme=uplift&variant=light
- Change view to mobile > Responsive
- Change width to 480 or lower
- Click on Show Modal
- Inspect modal wrapper
- Should not have the extra gap on the right side


tmcconechy (commented):
I think this looks good. But pinging @EdwardCoyle to test as im not sure what `frameWidth` is for?

tmcconechy (approved):

deep7102 (dismissed):

src/components/modal/modal.js
```diff
@@ -1148,7 +1148,7 @@ Modal.prototype = {
     } else {
       this.element[0].classList.remove('display-fullsize');
       calcHeight = ($(window).height() * 0.9) - this.settings.frameHeight;
-      calcWidth = ($(window).width() * 1) - this.settings.frameWidth;
```
EdwardCoyle, line 1151: I'm not sure about the change here -- This code path with `frameWidth` is for when `fullsize` isn't active, so I don't think it's fully addressing the original issue.  The padding is gone for sure, but this completely removes the `frameWidth` setting from the component which could break some functionality.

I think what might be better here is:
- Definitely remove any `width/min-width/max-width` attributes that were previously applied when the IF block above is hit
- I think we need to keep the size adjustments done in this codepath to prevent any API breakage.  If there are still padding issues on non-fullsize, maybe we need to consider defaulting `frameWidth` to 0.

EdwardCoyle (changes requested):
@yohannahbautista The fix in this PR does address the excess padding, but I think it might need some more work.  See my comment below:

tmcconechy (approved):

tjamesallen15 (approved):

EdwardCoyle (approved):
Looks great now, thanks @yohannahbautista 🎉

Issue #5113, opened 2021-04-20
Dropdown: Backspace key not setting a noSearch dropdown to "blank" value


awbuboltz:
**Describe the bug**
In previous versions, the backspace key could be used on a dropdown (with it closed), to set the value to blank when using a noSearch dropdown. The delete key still works to do this, but backspace is limited to only OSX. 

**To Reproduce**

1. Go to http://localhost:4000/components/dropdown/example-no-search-lsf.html
2. Tab to the first dropdown, type "5" to select 5, then hit backspace.
3. Note that it does not set the dropdown to blank.
4. Press "Delete" key and note that it does set the dropdown to blank.

**Expected behavior**
A noSearch dropdown that contains a blank value should clear out when using backspace. A customer reported this change in behavior to us.

**Version**

- ids-enterprise: 4.51-dev

**Platform**
 - Infor Application/Team Name: Infor Lawson
 - Device: Browser
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 90

**Additional context**
Talked about this one with @EdwardCoyle. In dropdown.js -> handlePseudoElemKeydown, I think this line ` if ((!isOSX && key === 'Delete') || (isOSX && key === 'Backspace')) {` could be changed to `if ((!isOSX && key === 'Delete') || (isOSX && key === 'Backspace') || key === 'Backspace' && this.settings.noSearch) {`.

awbuboltz:
Hey all, any timeline on this one? We've got a customer looking for the fix.

tmcconechy:
I'll put in for next sprint planning - will probably be able to add it. Then it will be done end of July for Aug release.

janahintal:
QA Passed. Thank you. 
Verified in https://main-enterprise.demo.design.infor.com/components/dropdown/example-no-search-lsf.html
![image](https://user-images.githubusercontent.com/45757016/125402167-abd86180-e3e6-11eb-9660-8aecde385e62.png)

PR #6154, opened 2022-03-01
6069 - Dropdown: Typing Capital Letters in an Open NoSearch Dropdown …

### LINKED ISSUE #6069: Dropdown: Typing capital letters in an open noSearch dropdown does not focus a value
**Describe the bug**
Typing a capital letter to select a value works when the dropdown is closed, but not when the dropdown is open. This happens on dropdowns with the noSearch option set to true.

**To Reproduce**
1. Go to http://localhost:4000/components/dropdown/example-no-search-lsf.html
2. Click on the top dropdown to open it
3. Type "T"
Expected Result: "T - Thor"  should be focused
Actual Result: The blank option is focused

**Expected behavior**
When a capital letter is typed in an open dropdown with noSearch turned on, the corresponding option should receive focus just as it would when the lowercase letter is typed.

**Version**
- ids-enterprise: v.4.60.0-dev

**Platform**
 - OS Version: Windows10
 - Browser Name: chrome
 - Browser Version: Version 97.0.4692.99 (Official Build) (64-bit)
Related Lawson JIRA Ticket: [LSF-50656](https://jira.lawson.com/browse/LSF-50656)

davidcarlsonberg:

In Windows10, using Chrome, some users are reporting that upon initial open of the dropdown and then immediately pressing 'Shift + T', the T input is not focused/selected.

Fixes #6069 

Visit on Windows10 and Chrome http://localhost:4000/components/dropdown/example-no-search-lsf.html
Click the noSearch dropdown to open (the first/top one)
Press Shift + T
The T entry should be focused.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -23,6 +23,7 @@
 - `[Datagrid]` Fix a bug in datagrid where text is align right when using mask options in filter. ([#5999](https://github.com/infor-design/enterprise/issues/5999))
 - `[Datagrid]` Fix a bug in datagrid where datepicker range having an exception when having values before changing to range type. ([#6008](https://github.com/infor-design/enterprise/issues/6008))
 - `[Datepicker]` Fix on the flickering behavior when range datepicker is shown. ([#6098](https://github.com/infor-design/enterprise/issues/6098))
+- `[Dropdown]` Fix an issue specific to Windows 10 and Chrome where entering a capital letter (Shift + T, e.g.) after opening the dropdown does not focus the entry associated with the letter pressed. ([#6069](https://github.com/infor-design/enterprise/issues/6069))
 - `[Dropdown]` Fix on dropdown multiselect where change event is not triggered when clicking X. ([#6098](https://github.com/infor-design/enterprise/issues/6098))
 - `[Editor]` Fix a bug in editor where CTRL-H (add hyperlink) breaks the interface. ([#6015](https://github.com/infor-design/enterprise/issues/6015))
 - `[Spinbox]` Spinbox should update to correct value when Enter is pressed. ([#6036](https://github.com/infor-design/enterprise/issues/6036))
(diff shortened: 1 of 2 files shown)


tmcconechy:
@s-werking look like its working to you?

tmcconechy (dismissed):

s-werking:
The only issue I see with this solution now is that the backspace and delete keys no longer focus the blank option when the dropdown is open.

tmcconechy:
Backspace appears to work for me (but im mac). Also looking again at the fix seems like it would basically just prevent Ctrl key from doing anything? So not sure i even understand the fix.

Any other ideas for a fix?

s-werking:
I checked on our product for the hotkeys that use the different control keys on open dropdowns and they still seems to work alright, this is more to prevent the control keys from showing up in the autocomplete search as a character.

The backspace and delete buttons are included in the list of control keys so that is why those don't function properly. The fix for the backspace/delete key on noSearch was put in handlePseudoElemKeydown() which is not hit in this scenario because the input is readonly (https://github.com/infor-design/enterprise/issues/5113).

 Having the backspace/delete key fix above additionally put into a place like handleKeyDown() where an open noSearch dropdown would hit and keeping the original fix that @davidcarlsonberg put in would hopefully fix the issue. We can have a call about this if it still doesn't make much sense, thanks!

s-werking:
I added a section of code for when backspace or delete is pressed and the noSearch dropdown is open. Here's the patch for it:
[Add_backspace_delete_key_handler_for_noSearch_dropdowns_when_open.zip](https://github.com/infor-design/enterprise/files/8164699/Add_backspace_delete_key_handler_for_noSearch_dropdowns_when_open.zip)

I checked that this would not change the behavior for dropdowns that are not noSearch.

tmcconechy (approved):

ericangeles (approved):

PR #6201, opened 2022-03-16
6048 - Updated modal close method

### LINKED ISSUE #6048: Modal: Modal close is being blocked when clicking a close button while a popupmenu or dropdown is open
**Describe the bug**
Closing a modal is being blocked while a popupmenu or dropdown is open. This is causing modals to not respond/close when clicking their buttons. It is also blocking the destroy process of a Contextual Action Panel in our Angular application (I believe one fix will resolve both issues). There is logic in the modal.js close function which aborts the close if there are any open sub components, which causes some issues (see comments below).

**To Reproduce**

1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/popupmenu/test-modal.html
2. Open the modal
3. Click one of the Menu Buttons to open the Popupmenu
4. With the popupmenu still open, click the Cancel or Save buttons, and notice the modal doesn't get closed. (A 2nd click with the popupmenu closed will allow the modal to close.)
5. The same problem can also be seen at the following example when a Dropdown field is open: https://design.infor.com/code/ids-enterprise/latest/demo/components/modal/example-close-btn.html

**Expected behavior**
Clicking a modal close button should work even when a popupmenu or dropdown is open.

**Version**
- ids-enterprise: v4.59.1

**Screenshots**

image

image

**Platform**
 - Infor Application/Team Name: CSD
 - Device: Laptop
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: 97.0.4692.71 (Official Build) (64-bit)

yohannahbautista:

Edited close method in modal, calling close method will close modal even if there are subcomponents opened.

Closes #6048 

- Go to: http://localhost:4000/components/popupmenu/test-modal.html
- Open the modal
- Click one of the Menu Buttons to open the Popupmenu
- With the popupmenu still open, click the Cancel or Save buttons


tmcconechy (changes requested):
@yohannahbautista - I think we should fix this within the existing `close` function and not add a new method. So this always works correctly. So i think we just need to alter the behavior to work not make a new thing you have to remember to call.

cc @EdwardCoyle for suggestion

tmcconechy (approved):

ericangeles (approved):

tmcconechy (approved):

PR #6239, opened 2022-03-29
NG1261 - Added legend load for datepicker

tjamesallen15:

This PR will add legend load for datepicker. This will partially fix the issue in NG but it still needs to be applied in NG after this this one is packaged.

Closes https://github.com/infor-design/enterprise-ng/issues/1261

- Go to http://localhost:4000/components/datepicker/test-legend-loaded.html
- Open datepicker
- Change the month to `February`

### DIFF
--- a/app/views/components/datepicker/test-legend-loaded.html
+++ b/app/views/components/datepicker/test-legend-loaded.html
@@ -30,7 +30,11 @@
     console.log('listopened', e);
   }).on('beforemonthrendered', (e) => {
     console.log('beforemonthrendered', e);
-    loadLegend();
+    if (datepickerAPI.currentMonth !== undefined) {
+      loadLegend();
+    }
+  }).on('listclosed', (e) => {
+    console.log('listclosed', e);
   });
 
   var datepickerAPI = $('#date-field').data('datepicker');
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
From this example i think maybe 

a) that the beforemonthrendered should also fire the first time the list opens as a month is rendering there
b) lets add the listclosed too so we can see it in the test page

Will need the types added laster as well to NG?

tjamesallen15:
> From this example i think maybe
> 
> a) that the beforemonthrendered should also fire the first time the list opens as a month is rendering there b) lets add the listclosed too so we can see it in the test page
> 
> Will need the types added laster as well to NG?

yes i will add the types later to NG after this one was merged. I'm having issues on the NG side because its not reflecting if I paste it directly on its dependencies. cc @ericangeles

tmcconechy (approved):

ericangeles:
@tmcconechy, we've been encountering that issue for quite some time now, and for some reason, the way we test the EP fix to NG doesn't work anymore even if it's just a CSS solution.

yohannahbautista (approved):

PR #6285, opened 2022-04-07
NG1262 - Fix on dropdown not moving when pressing key down

yohannahbautista:

Add an event handler in body that watches for keydown in dropdown.

Addresses issue in: https://github.com/infor-design/enterprise-ng/issues/1262

- Go to: http://localhost:4000/components/dropdown/test-label
- Click on a dropdown
- Use arrow up and down
- Selection should update


tmcconechy (changes requested):
@yohannahbautista this case works but seems to have broken this functionality on this test https://github.com/infor-design/enterprise/runs/5865980895?check_suite_focus=true#step:8:2868

Also the alignment looks bad on the example they gave us -> i know its an example but maybe we can make it look a bit better so QA doesnt complain

tmcconechy (dismissed):

tjamesallen15 (approved):

tmcconechy (approved):

Issue #628, opened 2018-08-17
Refactor all inline styles out.


tmcconechy:
**Describe the bug**

One problem is that our current CSP policy uses insafe inline for styles. We still need to do this because of the following reasons.. #256 Fixed many inline styles but the remaining issues exist as a continuation. We need to:

- [ ] need to remove any inline styles from all samples (fx display: none -> class hidden)
- [ ] need to remove any other inline styles from all samples (move to a script tag which has a nonce or use class)
- [ ] datagrid column widths use inline style for width. this needs to be refactored to use CSSDOM (and may make it slower)
- [ ] editor pasting html needs to somehow use CSSDOM (may be tricky)
- [ ] dragging/dropping code needs to be refactored to use CSSDOM (and may make it slower)
- [ ] search for `style=` in all .js files and fix those components (completion-chart, column, editor, tree should be simple to use CSSDOM)
- [ ] search for `style=` in all .html files 
- [ ] Consider a "bypassSecurity" setting for things like editor to avoid stripping.

**To Reproduce**
1. Go to csp-handler.js
2. Change it to `style-src': [ 'self', `nonce-${res.opts.nonce}`]`
3. Restart server
4. Go to any page fx kitchen sink, datagrid or 

**Expected behavior**
All pages should run with no csp errors without unsafe-inline styles.

**Workaround**
For now, running with this limitation of needing unsafe-inline.

EdwardCoyle:
two things:

> need to remove any inline styles from all samples (fx display: none -> class hidden)

Some of the demos have inline styles that are specific to said demo.  The demoapp has SASS files inside of `app/src/sass` that are built and served whenever the app's run.  All important inline styles should probably make their way into this folder, with the following guidelines:
   - Try to generalize styles if possible, for reuse between tests.
   - If styles can't be generalized, make some CSS comments above these specific styles to point out where they are used by the app (use the URL of the demo/test to label it)

> need to remove any other inline styles from all samples (move to a script tag which has a nonce or use class)

Can `<style>` tags also contain a nonce?  Is that valid?  If so, maybe we should try that?

tmcconechy:
Yes, i made style tags have a nonce in the Veracode PR. I started down this path but found it was much bigger than expected so its about 50% done on that PR. But general idea is if styles are needed in the demo they should go in style tag in that page or replaced with a class if we have one that will do it.

karinpb:
Can we use nonce with IDS? Or do we need to wait for this ticket?

[...]

PR #6286, opened 2022-04-08
6128 - Tooltip in Datagrid Disabled Rows

tjamesallen15:

This PR will fix a bug where tooltip is not displayed even when settings is turned on in disabled rows.

Closes https://github.com/infor-design/enterprise/issues/628

- Go http://localhost:4000/components/datagrid/test-tooltip-disabled-rows.html
- Hover the first row

### DIFF
--- a/app/views/components/datagrid/test-tooltip-disabled-rows.html
+++ b/app/views/components/datagrid/test-tooltip-disabled-rows.html
@@ -1,0 +1,42 @@
+{{={{{ }}}=}}
+
+<div class="row">
+  <div class="twelve columns">
+
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+
+    Locale.set('en-US').done(function () {
+        var grid,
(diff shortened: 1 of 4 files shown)


tmcconechy (approved):
This looks good - but as another optional improvement maybe we could make the tooltip come down a bit as it sits pretty far up to what its poining to on

http://localhost:4000/components/datagrid/example-tooltips.html
and
http://localhost:4000/components/datagrid/test-tooltip-disabled-rows.html

Can probably use the tooltip settings and maybe make the arrow sit approx 2px below the top line. (would also test this in all 4 row heights).

If you think thats too much and we can make a new issue, just let me know

tjamesallen15:
Did I get this one correctly? @tmcconechy, I adjusted the distance between the tooltip and top line.

tmcconechy (commented):
Screen Shot 2022-04-11 at 9 50 13 AM


Thinking more like this... I bit down more but not sure about the first row if its visible over the header?

yohannahbautista (approved):

PR #6295, opened 2022-04-11
NG1240 - Fix on too wide minimum width when close button is enabled

yohannahbautista:

Change minimum width only if screen size is greater than 400.

Addresses issue in: https://github.com/infor-design/enterprise-ng/issues/1240

- Use any device with screen width more narrow than 400 pixels. Or Inspect element and switch to mobile
- Go to: http://localhost:4000/components/modal/example-close-btn.html
- Click Show Modal (Full)
- Content should fit in screen

### DIFF
--- a/app/views/components/modal/example-close-btn.html
+++ b/app/views/components/modal/example-close-btn.html
@@ -28,6 +28,33 @@
 			</div>
 		</div>
 
+    <!-- Modal Full Example -->
+    <div id="modal-add-context-full" class="hidden">
+      <div class="form-responsive row">
+        <div class="full column">
+          <div class="field">
+            <label for="subject">Problem</label>
+            <input type="text" id="subject-full" name="subject" />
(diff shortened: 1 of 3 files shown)


tmcconechy (commented):
Looks good. But maybe we can make all the fields stretch across the modal usisng the respsonsive form classes https://main-enterprise.demo.design.infor.com/components/form/example-inputs-simple.html

Also see one unrelated test fail? Maybe it just needs an image update to pass?

tmcconechy (changes requested):
One thing i noticed with this solution is now if you go down below 400px you get the same issue. This is around 345px. We probably should flex it so it works in either case

Screen Shot 2022-04-12 at 9 27 18 AM

tmcconechy (dismissed):

tjamesallen15 (approved):

Issue #6253, opened 2022-04-01
Validator: Error message on toolbar button still appears after error is removed


s-werking:
**Describe the bug**
When calling removeMessage() on a toolbar button with an error and error message on it, the error is removed but the message remains visible.

**To Reproduce**
1. Please use the zipped html file: 
[test-error-on-toolbar-with-addmessage (2).zip](https://github.com/infor-design/enterprise/files/8399566/test-error-on-toolbar-with-addmessage.2.zip)
2. Click on "Show Error" button
3. Note that an error icon appears on the toolbar and the toolbar button with the error is red and has an error tooltip
4. Click on "Hide Error" button
5. Note that the error icon is removed and the toolbar button is no longer red, but the error tooltip is still visible.

**Expected behavior**
The error message should be removed once Hide Error is pressed.

**Version**
- ids-enterprise: v4.60.3

**Screenshots**
![Visible error message on remove message](https://user-images.githubusercontent.com/76002134/161307997-56ed2a10-a76f-4a46-8bf9-84ba2c3b8b14.gif)

**Platform**
 - Infor Application/Team Name: lawson
 - OS Version: Windows11
 - Browser Name: chrome
 - Browser Version: 100.0.4896.60 (Official Build) (64-bit)

tmcconechy:
I didnt know the icon got onto the toolbars. Is thats something you implemented?

s-werking:
I think it's only when inline is set to false for addMessage()

tmcconechy:
Would it be better if it didn't put it on the toolbar at all? I thought it was just for tabs where the tab could be hiding the message on another page

s-werking:
Yeah maybe removing the icon would be best

tmcconechy:
Will do i think in this pattern its a mistake

jbrcna:
QA Passed. v4.64.0-dev
https://main-enterprise.demo.design.infor.com/components/toolbar/test-error-on-toolbar-with-addmessage.html
https://main-enterprise.demo.design.infor.com/components/datagrid/test-editable-paging-validation.html
<img width="1354" alt="image" src="https://user-images.githubusercontent.com/56722184/170418243-95cb790e-ba2a-419b-84cc-8fc23ee17dd3.png">

iamjamesreyracho:
QA Passed. v4.64.0-dev. But I noticed that the show/hide error button was moved.
[...]

Issue #6290, opened 2022-04-08
Datagrid with validators and paging creates a toolbar and it is never removed


AnaLauraMartinez:
**Describe the bug**
Datagrid has validators and paging. When there is an error on the first page and the user goes to page 2 a toolbar appears with an error icon. When the user returns to page 1 and fixes the error, the error icon disappears but the toolbar remains.

**To Reproduce**
In a datagrid with paging and validators, datagrid doesn't have a toolbar, follow these steps:

1. Enter invalid data, notice the cell is marked with an error icon but there is no toolbar
2. Move to page 2, notice there is now a toolbar on the top with an error icon indicating there is an error on page 1
3. User moves back to page 1, notice the toolbar still there with an error icon
4. User fixes the error and the cell error icon disappeared, the toolbar is cleared from the error but the white space remains.

**Expected behavior**
Toolbar should disappear when the error is corrected

**Version**
"ids-enterprise-ng": "13.0.3"

**Screenshots**
![image](https://user-images.githubusercontent.com/17933529/162528062-a240bdc5-df82-4e6f-876d-35daea609c8b.png)
![image](https://user-images.githubusercontent.com/17933529/162528122-10a572e7-bf4a-4fbe-8e81-67d49ebc98e1.png)

**Platform**
 - Infor Application/Team Name: [e.g. Infor XYZ, Infor ABC]
 - Device: (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name: Chrome
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

When debugging the code that is adding the toobar is this and it doesn't remove it after the errors are cleared.
  /**
     * Show all non visible cell errors
     * @private
     * @returns {void}
     */
    showNonVisibleCellErrors: function showNonVisibleCellErrors() {
      var _this19 = this;

      // Create empty toolbar
      if (!this.toolbar) {
        this.settings.toolbar = {
          title: ' '
        };
        this.appendToolbar();
      }

PR #6393, opened 2022-05-06
6253 - validation error message toolbar button

tjamesallen15:

This PR will fix a bug where toolbar error message still appears after error is removed.

Closes https://github.com/infor-design/enterprise/issues/6253

- Go to http://localhost:4000/components/toolbar/test-error-on-toolbar-with-addmessage.html
- Click `Show Error`
- Click `Hide Error`
- go to http://localhost:4000/components/datagrid/test-editable-paging-validation.html
- edit a cell and make it blank
- go to page 2 -> error should show on the toolbar


tmcconechy (changes requested):
@tjamesallen15 the fix i was expecting would be that it doesnt even add the icon to the toolbar anymore. I think it should only add it to tabs (because you can switch the tab and need to know the hidden tab has errors)

But it should also work in this case:
http://localhost:4000/components/datagrid/test-editable-paging-validation.html
- edit a cell and make it blank
- go to page 2

In this case nothing is hidden. So is it possible to change the logic further so it only adds to tabs and datagrid toolbar?

tmcconechy:
There is a couple things wrong here.

1. an error: https://github.com/infor-design/enterprise/runs/6383440778?check_suite_focus=true
2. to clarify on http://localhost:4000/components/toolbar/test-error-on-toolbar-with-addmessage.html it should not show the icon on the toolbar but on datagrid it still should show it http://localhost:4000/components/datagrid/test-editable-paging-validation.html (on the datagrid its not)
3. on http://localhost:4000/components/toolbar/test-error-on-toolbar-with-addmessage.html when you click the button the error tooltip is showing in the wrong place?

tmcconechy:
@tjamesallen15 i. think the only thing wrong is something on this example isnt right the error is not showing on the field. This is what i expect:

http://localhost:4000/components/toolbar/test-error-on-toolbar-with-addmessage.html
- click show error or tab out of the field or tab out to get a validation error
- error should show on the field but not on the toolbar
- hide error should remove it from the field

tmcconechy (approved):

ericangeles (approved):

s-werking:
Hi @tmcconechy @tjamesallen15 , I think there is a misunderstanding about this ticket. I still would like the ability to add an error message to a toolbar to let a user know that they need to select something from a toolbar menu. With these changes in place I am unable to add/remove an error message on a toolbar. Is there any way this can still be an option?

tmcconechy:
The way i thought about it was that if you don't change the page there is no reason to put the validation message in two places (on the field and also on the toolbar). But in the datagrid example it makes sense? Plus we had these problems https://github.com/infor-design/enterprise/issues/6290

What exactly do you want to do? Echo the validation error in two places? Or manually add some error to the toolbar? If its just manually adding an error / maybe we just need an example for that as i dont think echoing the validation message would be the right way to do it.

tmcconechy:
Ok i was'nt aware that worked. @tjamesallen15 can you take another look? I think we should:

- modify / copy the example to just have a show message/hide message button you call that it sets the message on button and that works
- keep what we modified before - as its separate issue 

So just need an example of an error on a button?

tjamesallen15:
> Ok i was'nt aware that worked. @tjamesallen15 can you take another look? I think we should:
> 
> * modify / copy the example to just have a show message/hide message button you call that it sets the message on button and that works
> * keep what we modified before - as its separate issue
> 
> So just need an example of an error on a button?

Will do.

Issue #6326, opened 2022-04-22
100% Stacked Bar Chart - Component shows 99.999 instead of 100%


amittapally:
I am using 100% Stacked Bar Chart component for PLM for Fashion Critical Path widget. In the below screenshot all the tasks in the last milestone are in 'Not started' status but the tooltip shows Total 99.9999 instead of 100.

This issues happens only in some use cases 
Example :  When total Not started count is 28 it shows 99.9999 and when total Not started count is 36 it shows 100.

![image](https://user-images.githubusercontent.com/89181007/164611343-3e7fd3bf-7502-43a0-b377-f42e30fe7f1e.png)

**To Reproduce**
We are able to reproduce the issue. Please let me know if anything need to be changed in widget code to avoid this issue. This issue occurs only in some use cases. I used 28,0,0 to reproduce the issue.  [Ids Quick Start 1331 (forked) - StackBlitz](https://stackblitz.com/edit/ids-quick-start-1331-vlh8le?file=src/app/app.component.ts)

 [Ids Quick Start 1331 (forked) - StackBlitz](https://stackblitz.com/edit/ids-quick-start-1331-vlh8le?file=src/app/app.component.ts)

**Expected behavior**
Stacked Bar Chart should display total 100% in all the cases

**Version**
Infor_HomepagesWidgetSDK_3.9.0_20210922

**Screenshots**

![image](https://user-images.githubusercontent.com/89181007/164611114-ec10886b-e511-454c-beac-5ebfb6bbef97.png)

**Platform**
 - Infor Application/Team Name:  M3 T2V Team
 - Device: 
 - OS Version: Windows 11
 - Browser Name: Chrome
 - Browser Version: 100.0.4896.127

GlenlieMaye:
Changes have already been applied. However in the label, the 'Value' is in bold and the 'Total' is not, can we make it consistent?

![image](https://user-images.githubusercontent.com/102151352/169432121-29b83ffb-bba0-4f5d-8c7b-10df83f5a8a5.png)

PR #6396, opened 2022-05-10
6326 - 100% Stacked Bar Chart: Component shows 99.999 instead of 100%

sesbrenosienna:
This PR fixes a bug where the chart tooltip total shows 99.999 instead of 100 on the 100% Stacked Bar Chart.

Closes https://github.com/infor-design/enterprise/issues/6326

- Go to http://localhost:4000/components/bar-stacked/example-stacked-100
- Hover over the 2014 bar 


tmcconechy:
Looks good - but im just wondering if the total should be the "part" of the bar. I.E. on http://localhost:4000/components/bar-stacked/example-index.html if you hover the blue part it shows that total (and the percents) and hover the green part it shows that total.

I also tested http://localhost:4000/components/bar/example-index.html and http://localhost:4000/components/bar-grouped/example-negative.html and those still work. So thats good :)

sesbrenosienna:
Hi Sir Tim! I clicked on the link http://localhost:4000/components/bar-stacked/example-index.html and tried hovering over the blue and light blue parts, but it shows the total for both parts. I also tried using this link https://main-enterprise.demo.design.infor.com/components/bar-stacked/example-index.html, but it still showed the same. :) 
image

tmcconechy:
hi @sesbrenosienna - no need for sir with me 😄 

What i mean is actually the opposite:
1) hover the blue part it shows a tooltip for that part / hover the green part it shows a tooltip for that part 
2) so the total should be just for each part (because its hard with the lines to see the exact value

We could do something like: 
234: Value
358: Total

yohannahbautista (approved):

ericangeles (approved):

PR #6428, opened 2022-05-17
6218 - Fixed bug where lookup still appears when modal closes

### LINKED ISSUE #6218: Modal/Lookup: Closing a modal before an async lookup that's in it is loaded breaks the UI
**Describe the bug**
When there's an async lookup inside a modal and you close the modal while waiting for the lookup to load, the modal backdrop doesn't disappear which prevents you from interacting with the page and you need to reload the page to fix it.

**To Reproduce**
The following code can be used to reproduce this:

```
// Button to open the modal
@Component({
	selector: "demo-button",
	template: `<button soho-button="primary" (click)="openDialog()">Open Dialog</button>`
})
export class DemoButtonComponent {
	constructor(private readonly dialogService: SohoModalDialogService) { }

	openDialog(): void {
		const dialog = this.dialogService.modal(DemoModalComponent);
		dialog.buttons([{ text: "Close", click: () => dialog.close() }]).open();
	}
}

// Modal
@Component({
	selector: "demo-modal",
	template: `<input soho-lookup [beforeShow]="beforeShow"/>`
})
export class DemoModalComponent {
	beforeShow = (_: SohoLookupStatic, response: SohoLookupBeforeShowResponse) => {
		setTimeout(() => response(), 2000);
	}
}
```

1. Click the button to open the modal.
2. Click the lookup button.
3. Close the modal either by clicking the "Close" button or the "Esc" key BEFORE the lookup has been loaded. Have to be quick before the modal opens
4. The lookup doesn't open and the modal closes, but the modal backdrop is still there which prevents you from interacting with the page.

**Expected behavior**
The lookup shouldn't open and the modal should close, but the backdrop should also disappear.

**Version**
- Observed and reported in ids-enterprise@4.60.0 / ids-enterprise-ng@13.1.1

**Screenshots**
dialog_lookup

**Platform**
 - Infor Application/Team Name: Homepages
 - Device: All
 - OS Version: All
 - Browser Name: All
 - Browser Version: All

**Additional context**
Tested the same scenario with a second modal instead of a lookup but didn't observe the issue in that case.

yohannahbautista:

Added check to see if lookup input element is visible before opening lookup. If it's not visible, lookup should not show.

Closes #6218 

- Go to: http://localhost:4000/components/modal/test-async
- Click on Show Modal
- Click on lookup icon
- Lookup should show after around 2 seconds
- Close modal and click on Show Modal again
- Click on lookup icon and either press "Esc" or close the modal right after
- Modal should close, lookup should not show 

### DIFF
--- a/app/views/components/modal/test-async.html
+++ b/app/views/components/modal/test-async.html
@@ -1,0 +1,81 @@
+<div class="row top-padding">
+	<div class="twelve columns">
+
+		<button class="btn-secondary" type="button" id="show-modal">Show Modal</button><br/><br/>
+
+		<div id="example-modal" class="hidden">
+      <div class="container">
+        <div class="form-responsive row no-indent">
+          <div class="full-width column">
+              <div class="field">
+                <input id="example-lookup" data-init="false" class="lookup" name="example-lookup" type="text"/>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+	</div>
+</div>
+
+<script>
+const modals = {
+    'show-modal': {
+      'id': 'my-id',
+      'content': $('#example-modal')
+    }
+  };
+  
+  let timer, data = [], columns = [];
+
+  columns.push({ id: 'productId', name: 'Product Id', field: 'productId'});
+  columns.push({ id: 'productName', name: 'Product Name', field: 'productName', formatter: Formatters.Hyperlink});
+  columns.push({ id: 'activity', hidden: true, name: 'Activity', field: 'activity'});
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
Seeing some cases where i cant open the modal anymore but its a little hard to reproduce.

This seems to work
- open modal by clicking the button -> hit escape after it loads
- repeat this and after 3/4 times it will not be possible to open the modal anymore

ericangeles (approved):

tmcconechy (approved):

Issue #6074, opened 2022-01-31
Bar Chart: VPAT Accessibility issues


JBismonte:
**Describe the bug**
Bar charts do not provide textual alternatives.
Information in bar charts including x and y axes and bars is contained in an svg without text alternative. Text fill values are announced in some browser/screenreader combinations, but cannot be relied upon as a textual alternative. The actual values for each bar, displayed in a tooltip on mouse hover (e.g. "Category B 372", "Category C 236.35") is not announced to screen reader users. 

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/bar/example-index.html
2.  bar chart cannot be focused by the keyboard and when using a screen reader, its tooltips are not announced when the mouse is hovered in each of the bar.

**Recommendation by VPAT**
Developers must ensure that information conveyed in charts and graphs is provided textually.
Consider also providing the information in a table alongside the chart. This approach allows all users to see the information presented in a more structured way.
If this is not possible, the table alternative could be visually hidden, and only presented to screen reader users.
If it is necessary to instead try to make the SVG fully accessible, refer to the following resources that may be of use:
* https://css-tricks.com/accessible-svgs/
* http://simplyaccessible.com/article/7-solutions-svgs/

**Version**

- ids-enterprise: 4.60

**Screenshots**
![image](https://user-images.githubusercontent.com/29673378/151765605-15829018-e3ff-4612-a7b2-01f68252d506.png)

**Platform**
 - Infor Application/Team Name: Landmark Client
 - Device: (if applicable) : NA
 - OS Version: Windows 10
 - Browser Name: Firefox
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**
https://jira.lawson.com/browse/LMCLIENT-34217

tmcconechy:
Plan for this:

- solve as a list in WC version
- provide example here with the same (might need help from @ericangeles )
- will implement for just bar code so that it can be verified before adding to any other components here

tmcconechy:
If you see what i did on https://main.wc.design.infor.com/ids-bar-chart/example.html with aria i think we should implement the same in the EP bar chart.

- a title in the svg
- a parent element with role="list" and an aria-label
- group each bar in a <g> with role="lisitem"inside that have the bar but also an offscreen text element with all the info in it
- Then when you try it with a screen reader you navigate to the chart with the virtual cursor then can read through the chart data like a list

[...]

PR #6531, opened 2022-06-02
6074/6541 - Fix VPAT accessibility in bar chart

ericangeles:

This PR enhances the VPAT accessibility in Bar Chart.

Closes https://github.com/infor-design/enterprise/issues/6074
Closes https://github.com/infor-design/enterprise/issues/6541

- go to http://localhost:4000/components/bar/example-index.html
- turn on voice over (or any screen reader)
- caps lock + left/right arrow (or screen reader nav key)
- do it until you hit the page
- For #6541, you should be able to tab from the toolbar element in the header to the main elements


tmcconechy:
@ericangeles not quite try https://main.wc.design.infor.com/ids-bar-chart/example.html

- Using voice over the sequence is to
  - Navigate to the parent element above it  or parent page
  - Hold <kbd>caps lock + left/right arrow</kbd>
  - You will hear the title , followed by number of items and then each list item
  - Proceed to use <kbd>caps lock + left/right arrow</kbd> will move through the list items announcing the values

You will see it announces that its a list/group and then you can arrow through the list items (the bars).
But this one has two issues:

1) it just says "bar graphic" not list so might need to remove that part
2) the numbers at the bottom are navigable (to fix this i added `aria-hidden="true"` in the WC one

tmcconechy:
@ericangeles still having troubles with this i raised https://github.com/infor-design/enterprise/issues/6541 as i cant get from the toolbar down.

But if i select the title text above the card and navigate from there using caps-lock+left/down then all i hear is "bar Graphic empty" im not hearing the list?

ericangeles:
@tmcconechy am having a hard time going to the bar section (from the title of the card) but I can confirm that the "bar graphic empty" is mentioned in the voice-over.

tmcconechy (commented):
Looking a Lot better. 

a) does it still work without the tabindex? Problem with that is either we have to style the focus state or take it off and just use the virtual cursor like: https://main.wc.design.infor.com/ids-bar-chart/example.html
b) i think maybe we need a aria-label on the group as it just goes to "list" but doesnt tell you what the list topic is like https://main.wc.design.infor.com/ids-bar-chart/example.html

ericangeles:
> Looking a Lot better.
> 
> a) does it still work without the tabindex? Problem with that is either we have to style the focus state or take it off and just use the virtual cursor like: https://main.wc.design.infor.com/ids-bar-chart/example.html b) i think maybe we need a aria-label on the group as it just goes to "list" but doesnt tell you what the list topic is like https://main.wc.design.infor.com/ids-bar-chart/example.html

It does work  but had hard time to tab to the chart. I knew it will have a focus style that's why I added css fix for that.

tmcconechy:
@ericangeles when i turn off the voice over. and tab to it I'm not seeing anything. If we put that on and add an additiional aria-label for the chart title i think this will work?

ericangeles:
@tmcconechy what aria-label value should we put?

tmcconechy:
@ericangeles what i did was:

a) add a chart title (notice two titles this is so the browser tooltip doesnt show)
b) add an aria-label that is the legend text (so you know the chart items are about that)
Screen Shot 2022-06-13 at 11 28 47 AM

tmcconechy (approved):
Ok lets move this further to test. I think its working pretty well 👍🏻

tjamesallen15 (approved):

PR #6544, opened 2022-06-06
NG1305 - Monthview legend fix

yohannahbautista:

loadLegend method will also update the dates to add legend color in them

Addresses this issue: https://github.com/infor-design/enterprise-ng/issues/1305

- Go to: http://localhost:4000/components/monthview/test-legend-loaded.html
- Go to next month
- Legends should change
- Click on Update Legend
- Legends should change


tmcconechy (changes requested):
There is part of this issue thats not 100% clear but the team needs to be able to do an ajax call to get the data. When you do that the events in the calendar dont update.

To see this I added a setTimeout to simulate this in the example. If you go to next month now you see the events are not updated.

So we ideally need to account for this somehow.

Also wasnt sure if it should go in beforemonthrendered or monthrendered so went for monthrendered. And the example no longer has a button so could remove that click code?

yohannahbautista:
> There is part of this issue thats not 100% clear but the team needs to be able to do an ajax call to get the data. When you do that the events in the calendar dont update.
> 
> To see this I added a setTimeout to simulate this in the example. If you go to next month now you see the events are not updated.
> 
> So we ideally need to account for this somehow.
> 
> Also wasnt sure if it should go in beforemonthrendered or monthrendered so went for monthrendered. And the example no longer has a button so could remove that click code?
@tmcconechy 
Hi, I changed the test page and it should work like this:
- There are two legend lists that are used. One when going to the next month and one when clicking Update Legend
- Legends should update whenever Update Legend is clicked or whenever month is changed
- Added timeout for update legend to simulate ajax calls
- Legend settings from the api are printed below the monthview and should be updated as well

tmcconechy (approved):
Great! Works with the timeout now. I think this should work

PR #6577, opened 2022-06-15
NG1300 - Datagrid Strange Focusing Behavior

tjamesallen15:

This PR will fixed a bug in datagrid where focus is not behaving properly when inlineEditor is set to true.

Closes https://github.com/infor-design/enterprise-ng/issues/1300

- Go to http://localhost:4000/components/datagrid/test-cellchange-event.html

Case 1:
- Click into the input field in row 1
- Mark the existing number in row 2 without clicking into the input field before
- Change the number

Case 2:
- Mark the existing number in row 2 without clicking into the input field before
- Change the number
- The cursor stays in row two and the number seems to be changed
- Leave the input field
- Click on input field in row 2 and the value should not change back to the original value


tmcconechy (dismissed):
Looks good to me - can you also test this branch @jriedel-osp and confirm?

jriedel-osp:
Also looks good to me. Thank you for the solution!

tmcconechy (approved):

Issue #6434, opened 2022-05-17
Treegrids: cannot expand a row via keyboard when there is an editable attribute in the grid options


JBismonte:
**Describe the bug**
when a treegrid's option has an editable attribute to true. we cannot expand the row via the space button when you are focused at the cell with the expand button, this also doesn't work if the cell with the expand button is editable or not.

**To Reproduce**

1. Go to [here](https://main-enterprise.demo.design.infor.com/components/datagrid/test-tree-with-checkbox.html)
2. focus on the cell with the expand button
3. try clicking space
4. the expand doesn't work.

**Expected behavior**
it should expand. I also found another way on how they expand here:
https://www.w3.org/TR/wai-aria-practices/examples/treegrid/treegrid-1.html

where you are in a expandable cell, you can click the right arrow key to expand, after it is expanded, you can click the right arrow key again to move to the next cell, and you can move back again with the left arrow key, and collapse the expanded cell with the left arrow key.

**Version**

- ids-enterprise: 4.64

**Screenshots**

**Platform**
 - Infor Application/Team Name: Landmark UI
 - Device: (if applicable) : NA
 - OS Version: Windows 11
 - Browser Name: Firefox
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**
[Jira Ticket](https://jira.lawson.com/browse/LMCLIENT-34746)

janahintal:
The issue still exists in https://main-enterprise.demo.design.infor.com/components/datagrid/test-tree-with-checkbox.html?theme=classic&mode=light&colors=2578a9 for both windows and macOS browsers. 

Steps: 

1. Go to https://main-enterprise.demo.design.infor.com/components/datagrid/test-tree-with-checkbox.html?theme=classic&mode=light&colors=2578a9
2. Focus on the cell with the expand button
3. Hit space
4. It should expand the row of treegrid

Thanks

PR #6578, opened 2022-06-15
6434 - Fix treegrid not expanding a row when using keyboard when editable is set to true

ericangeles:

This PR fixes the treegrid not expanding a row when using the space keyboard when the editable setting is true.

Closes https://github.com/infor-design/enterprise/issues/6434

- Go to http://localhost:4000/components/datagrid/test-tree-with-checkbox.html
- Focus on the cell with the expand button
- Hit space
- It should expand the row of treegrid


tmcconechy (commented):
Not sure if this is worth addressing (or can be addressed). But if you tab in to the first cell on the first row and hit space it does expand properly but seems to jump and scroll down.

src/components/datagrid/datagrid.js
```diff
@@ -9585,6 +9585,12 @@ Datagrid.prototype = {
       // For Editable mode - press Enter or Space to edit or toggle a cell,
       // or click to activate using a mouse.
       if (self.settings.editable && key === 32) {
+        const btnExpand = target?.find('.datagrid-expand-btn');
+        if (btnExpand) {
+          btnExpand.trigger('click.datagrid');
+          e.preventDefault(); // This will prevent scrolling down when the list is overflowing.
```
tmcconechy, line 9591: 👍🏻

yohannahbautista (approved):

PR #6599, opened 2022-06-23
1274 - Datagrid: Shift-click on the always selects the first row

sesbrenosienna:
This PR fixes the bug where using shift-click to multi-select on Datagrid with treeGrid setting = true selects from the first row until the bottom row.

Closes https://github.com/infor-design/enterprise-ng/issues/1274

Case 1: 
- Go to http://localhost:4000/components/datagrid/test-tree-select-multiple.html
- Click on the row with id = 4
- Do a shift-click on the row with Id=8

Case 2:
- Go to http://localhost:4200/ids-enterprise-ng-demo/datagrid-treegrid-lazy
- Click on the row with id = 11
- Do a shift-click on the row with Id=14


tmcconechy (changes requested):
I dont think the behavior is 100% correct here on http://localhost:4000/components/datagrid/test-tree-select-multiple.html

Try clicking id 4 and then shift clicking id 2 (it selects one wrong row). The way to think about it is the behavior should be the same as if you clicked each row individually 2-4

tmcconechy (approved):
Looks good now..

yohannahbautista (approved):

Issue #6580, opened 2022-06-15
Datagrid multiselect filter dropdown overflows when at edge of the screen


ntscheel:
**Describe the bug**
When filtering on a datagrid, clicking a column filter with a multiselect dropdown that is near the edge of the screen can cause the dropdown to open and overflow the page container, pushing everything to the left. Note that this does not happen with dropdowns that are **not** multiselect. See gif below.

**To Reproduce**

1. Go to [https://master-enterprise.demo.design.infor.com/components/datagrid/example-filter.html](https://master-enterprise.demo.design.infor.com/components/datagrid/example-filter.html)
2. Shrink the screen width so that "Activity" or "Activity (Multi)" are near the right edge of the screen 
3. Click open the dropdown and see that the page shifts left.

**Expected behavior**
Dropdowns with multiselect should function like dropdowns with single select, not pushing the screen around.

**Version**

- IDS Enterprise Components - v4.51.0-dev

**Screenshots**
![SoHo_DatagridFilterDropdownOverflow_Bug](https://user-images.githubusercontent.com/25674227/173854947-49a3f869-b6c1-48d5-a2c0-599d19a0b406.gif)

**Platform**
 - Infor Application/Team Name: Infor Landmark
 - OS Version: Windows 11
 - Browser Name: Chrome
 - Browser Version: 102.0.5005.115 (Official Build) (64-bit)

**Additional context**
Issue was reported by Infor HRT QA, but I didn't bother including that information since it is replicable on the soho components demo. Will add that additional information if requested.

jbrcna:
QA PAssed
v4.66.0-dev
https://main-enterprise.demo.design.infor.com/components/datagrid/example-filter.html
https://main-enterprise.demo.design.infor.com/components/datagrid/example-filter?theme=new&mode=light&colors=0066D4&locale=ar-SA

PR #6643, opened 2022-07-08
6580 - Fix overflowing of dropdown in datagrid when at edge of the screen

ericangeles:

This PR fixes the overflowing of the multiselect dropdown on the page and pushes the container near the screen's edge.

Closes
- #6580 

- Go to http://localhost:4000/components/datagrid/example-filter.html
- Shrink the screen width so that "Activity" or "Activity (Multi)" is near the right edge of the screen
- Click open the dropdown
- Dropdown should be placed inside of the datagrid container and the page should not add white space
- Test in RTL position too http://localhost:4000/components/datagrid/example-filter?theme=new&mode=light&colors=0066D4&locale=ar-SA

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@
 ## v4.66.0 Fixes
 
 - `[Datagrid]` Fixed redundant aria-describedby attributes at cells. ([#6530](https://github.com/infor-design/enterprise/issues/6530))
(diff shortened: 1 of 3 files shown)


tmcconechy (commented):
This is a minor thing but i wonder if the popup can be moved so its more over the left side of the field instead of flush with the right side of the screen.

Screen Shot 2022-07-08 at 9 35 41 AM

If not no worries, seems minor.

yohannahbautista (approved):

tmcconechy (approved):

PR #6647, opened 2022-07-11
6588 - Fix on edit outline in textarea not filling the entire cell

### LINKED ISSUE #6588: Datagrid: Editor does not outline the cell when a text area cell has multiple lines
**Describe the bug**
The editor of a cell will not outline the full cell when in the same row as a text area cell with multiple lines. It keeps the defined row height as an outline of the cell.

**To Reproduce**
1. Please use this zipped html file: 
[test-text-area-with-editor.zip](https://github.com/infor-design/enterprise/files/8930899/test-text-area-with-editor.zip)
2. Click on the cell that says "Assemble Paint" to edit
Result: The editor keeps the defined row height and does not outline the entire cell.

**Expected behavior**
It is expected that when editing a cell, the editor outlines the cell.

**Version**
- ids-enterprise: v4.64.0-dev

**Screenshots**
Text Area Editor Outline

**Platform**
 - OS Version: Windows 11
 - Browser Name: chrome
 - Browser Version: 102.0.5005.115 (Official Build) (64-bit)

yohannahbautista:

Border is moved to gridcell instead of wrapper so it appears in the entire cell.

Closes #6588 

- Go to: http://localhost:4000/components/datagrid/test-text-area-with-editor.html
- Click on first Activity cell
- Border should be in the whole cell


ericangeles (dismissed):

tmcconechy (changes requested):
I think the reason i had the border on the inner part is because of this problem you can see on http://localhost:4000/components/datagrid/example-editable.html when editing the last row.
Screen Shot 2022-07-12 at 1 49 54 PM

Also on http://localhost:4000/components/datagrid/test-text-area-with-editor.html if you edit the first cell the editing goes from centered vertically to top vertically. Maybe we can add a class for this type of editor to the editing input.

So i think we should fix the editing alignment and maybe find a different fix that leaves the border on the inside?

tmcconechy:
@yohannahbautista yeah i see the complication with centering the text. I noticed with your fix for that it isnt 100% centered and if you open the cell with enter it is not centered.

Maybe just revert that part and we can leave it as it was?

tmcconechy (dismissed):

tmcconechy (dismissed):

ericangeles (approved):

Issue #6490, opened 2022-05-25
Datagrid: Add ability to have an icon on the datagrid header with a tooltip


pwpatton:
**Is your feature request related to a problem or use case? Please describe.**
In landmark we have a new feature for fields called "mouse over text". When a customer defines this text we place an icon next to the fields label. When the user hovers over that icon we build the text (lazily) and display the tooltip. We'd like to be able to have this for datagrid headers as well.
Here is an example of it on a form
![image](https://user-images.githubusercontent.com/4737815/170274397-0cbe45e1-cc37-40af-ad52-7cb31b6edeea.png)

**Describe the solution you'd like**
- We'd like to be able to place an icon on each header in a datagrid that has our "mouse over text" to work in the same manner. This "mouse over text" should only display when hovering over the icon.
- Have a beforeOpen callback to be able to lazily build our mouse over text to display in a tooltip.
- Have the ability to place a class on the header element so we can scope some necessary css changes.

![image](https://user-images.githubusercontent.com/4737815/170277864-b307a66b-a523-4a8a-84c7-765c320c5bc1.png)

A note on the mouse over text feature: it's NOT to be considered help text, and hence no content here is delivered, but just something the customer can configure for "important" fields where they want to have some "instructive" text.

**Describe alternatives you've considered**

**Additional context**
This is being asked for from Koch
And what application do you work on? Landmark

pwpatton:
Related issue is https://github.com/infor-design/enterprise/issues/4596

tmcconechy:
Similar to https://github.com/infor-design/enterprise/issues/4596

tmcconechy:
Question: What if there is a header tooltip already? Can that happen? I.E. how should we integrate this with the tooltips on header?

pwpatton:
@tmcconechy I was thinking when we did a mouse over on the icon in the table header we'd see this special "mouse over text".  Mouse over anywhere else and see the normal table header tooltip. What do you think?

tmcconechy:
Yeah i guess thats the only way but i think its bad from a usability standpoint perhaps to have two tooltips for the same small area. Anyway you dont have two tooltips just one or the other?

tmcconechy:
Kinda like that! - but whats in your header tooltips? Anything or just the header when it is doing the .... thing?

pwpatton:
Just the header name when it's doing the ... thing

tmcconechy:
Ok then i like the solution you suggested...

[...]

PR #6661, opened 2022-07-18
6490 - Add ability to have an icon on the datagrid header with tooltip

ericangeles:

This PR allows adding an icon in the column header with a tooltip and adding a custom class.

For this to achieve, there are a couple of column settings you need to set,
- `headerIcon` configurable by adding icon name e.g., `info`, `alert`, `notes` etc.
- `headerIconTooltip` the contents

For the custom class in the column header, you just need to set
- `headerCssClass` name of the custom class

FYI - Header Title/Text will be automatically appended in the tooltip content by default (for header info icon only).

Closes
- #6490 

- Go to http://localhost:4000/components/datagrid/example-header-icon-with-tooltip
- Hover on the info icon beside the `Product Name` column header
- It should show the tooltip info with the header text
- Resize the column header until the text gets ellipsis
- Hover the Info icon
- It should still show the tooltip info with the header text


tmcconechy (changes requested):
Hi @ericangeles just a couple of things but looks great!

1) I think the tooltip should show in two cases a) if you hover the icon and b) if you hover the column AND the text ellipsis is on (as that is the behavior for ellipsis text)
2) Looks like the styling could use more top and bottom padding (make same as the tooltip code?). Also can we add more line height for the <p> text so its more spaced
3) Can we change the options so the icon is totally configurable? I suggest:

headerIcon (any sting like 'icon-info' or 'icon-anything' instead of hasHeaderIcon if it exists there is an icon added and can be any icon)
headerTooltip (as it may also show on the ellipsis text)
headerCssClass (instead of customHeaderClass to match the existing cssClass for cells)

tmcconechy:
@ericangeles the only thing confusing to me is the "Header Tooltip" that sometimes pops up? How that be disabled for this example? I think we just want it in these two cases:

a) if you hover the icon and 
b) if you hover the column AND the text ellipsis is on

But that could be used on other columns.

tmcconechy (approved):

Issue #6521, opened 2022-06-01
Could not perform copy and paste function of any text of a PDF file opened in Acrobat Reader to our rich text editor component. 


rvilleno:
**Describe the bug**
Could not perform copy and paste function of any text of a PDF file opened in Acrobat Reader to our rich text editor component. 

**To Reproduce**

1. Copy any text from a PDF file opened in Acrobat Reader
2. Go to http://localhost:4000/components/editor/example-index.html
3. Paste on editor

**Expected behavior**
Copied text will be pasted on editor.

**Version**

- ids-enterprise: IDS Version: 4.64.0-dev.20220504

**Screenshots**
![LMCLIENT-37739](https://user-images.githubusercontent.com/10869544/171329209-4ad7de44-34ad-410e-84a9-94925a461f4c.gif)

**Platform**
 - Infor Application/Team Name: Landmark UI
 - OS Version: Windows 11
 - Browser Name: Chrome
 - Browser Version: Version 102.0.5005.63 (Official Build) (64-bit)

**Additional context**
If the PDF file is opened on a browser, the copy and paste function is working.

tmcconechy:
I couldn't reproduce this. I found a simple pdf online http://www.africau.edu/images/default/sample.pdf
Copied and then pasted and it worked.

![Screen Shot 2022-06-01 at 9 26 08 AM](https://user-images.githubusercontent.com/814283/171415555-7f482058-74f2-495c-9107-fd92f508ae99.png)
![Screen Shot 2022-06-01 at 9 25 55 AM](https://user-images.githubusercontent.com/814283/171415558-747a878b-8443-440b-8f91-ed145d23ff70.png)

This being said i dont think every single thing will work when pasting you might have to format things a bit. Or copy different parts.

tmcconechy:
Also read the comment on the other issue and installed Adobe reader and tried it - same thing it works. Except for sometimes the text isn't exact same font which is expected you may need to format a bit.

Workaround for problematic Pdfs is paste into note pad? Or can you paste into note pad either?
Did you try more than one PDF? Can the test pdf be attached thats bad.

Not confident we can change anything here.

rvilleno:
That is weird, I tried the same PDF file you used located in http://www.africau.edu/images/default/sample.pdf performing the same steps, I still encountered the same issue as the customer.

[...]

PR #6664, opened 2022-07-19
6521 - Editor Paste not working when copied from Windows PDF Reader

tjamesallen15:

This PR will fix bug where paste function is not working on editor when copied from Windows Adobe Reader.

Closes https://github.com/infor-design/enterprise/issues/6521

- Use a Windows PC
- Go to http://localhost:4000/components/editor/example-index.html
- Download the pdf in http://www.africau.edu/images/default/sample.pdf
- Open the PDF File
- Copy any line from the PDF File
- Paste it in the Editor


tmcconechy (changes requested):
@tjamesallen15 it seems like this breaks the ability to paste formatted word docs in the editor.
Try to make a word doc that has formatting like bold/fonts and copy paste and you can see its not working anymore.

- is there some way to detect the PDF text in the paste? 
- or figure out why its getting "stripped"?
- we did already have this option to paste as text http://localhost:4000/components/editor/test-paste-as-plain-text.html so what you added seems like the same thing as turning that on. last case we could tell them that they need this on?

tjamesallen15:
> 

Hello @tmcconechy ,
When I try to copy from Acrobat Reader (Windows PDF Reader) all I get is text/html and text/rtf types, there are no text/html types which we need to get the bold/italics. See the image below.
copied from acrobat

I tried copying from MS Word and I get  the text/html fine. 
copied from word

So what I did is just check if the text/html is there. Now it is working with bold text copied from other sources.

tmcconechy (dismissed):
Ok i think this will work. Thanks!

docs/CHANGELOG.md
```diff
@@ -17,6 +17,12 @@
 - `[Datagrid]` Fixed the overflowing of the multiselect dropdown on the page and pushes the container near the screen's edge. ([#6580](https://github.com/infor-design/enterprise/issues/6580))
 - `[Datagrid]` Fixed unselectRow on treegrid sending rowData incorrectly. ([#6548](https://github.com/infor-design/enterprise/issues/6548))
 - `[Datagrid]` Fixed incorrect rowData for grouping tooltip callback. ([NG#1298](https://github.com/infor-design/enterprise-ng/issues/1298))
+- `[Editor]` Fixed a bug where paste function is not working on editor when copied from Windows Adobe Reader. ([#6521](https://github.com/infor-design/enterprise/issues/6521))
```
ericangeles, line 20: I see some duplicate logs. Can you fix it @tjamesallen15 ?

ericangeles (approved):

Issue #6663, opened 2022-07-18
Datagrid: Render issue with editable alerts


claudenbach:
**Describe the bug**
A render issue with a column that contains an alert and whose editor has a trigger icon .

**To Reproduce**
1. Using the attached 'test-editable-alerts.html'
2. Look at the Action column - notice the alert icon is not visible
3. Mouse over a value in the Action column - notice the hover action

**Expected behavior**
Expect the alert icon to be visible and not having a hover (or mouse over) change occurring.
Seeing this same behavior on any column editor that has a trigger icon with an alert.

**Version**
- ids-enterprise: v4.66.0-dev

**Additional context**
Have a Landmark regression ticket attached to this issue.  Apparently this was last working in v10.14.0 (v4.61).

claudenbach:
[EditableAlerts.zip](https://github.com/infor-design/enterprise/files/9132813/EditableAlerts.zip)

claudenbach:
Landmark example were is was working...

![Alert-Icon-Trigger-Not-Working](https://user-images.githubusercontent.com/39804038/179551684-193c5764-5310-4017-a38a-58a13d6f7405.gif)

tmcconechy:
@ericangeles where you able to determine which issue this is on? Or have a suggestion? I can look but when i tried i really wasnt sure which issue might have caused this.
@tmcconechy Note to self: patch with https://github.com/infor-design/enterprise/pull/6655 into 4.65.x

ericangeles:
@tmcconechy not yet but am looking now.

claudenbach:
Noticed when using the attached sample, if I change the row height to Extra Small or Small (which is what Landmark uses), then the error icon's margin-left property becomes a margin-left: -26px.  This doesn't seem correct?  

![Icon-Error-Margin-Left](https://user-images.githubusercontent.com/39804038/179825407-5211d220-59e2-47ad-84e3-4852ade8e6a2.gif)

claudenbach:
I've updated the test to be more realistic...

[EditableAlertsUpdate.zip](https://github.com/infor-design/enterprise/files/9151756/EditableAlertsUpdate.zip)

tmcconechy:
Closing but working through patches

tmcconechy:
Can you help to fix this on all branches? For some reason this doesn't work on 4.62.x and maybe others.

[...]

PR #6671, opened 2022-07-20
6663 - Fix visiblity of datagrid alert icon in datagrid trigger cell

ericangeles:

This PR fixes the visibility of datagrid alert icon when the trigger cell is active and moving cell content when hovering over.

Closes
- #6663 

- Go to http://localhost:4000/components/datagrid/test-editable-alerts
- Check the `Action` Column
- Alert icon should be visible
- Hover on the cell
- It should not move when hovering over
- Test on different row heights

### DIFF
--- a/app/views/components/datagrid/test-editable-alerts.html
+++ b/app/views/components/datagrid/test-editable-alerts.html
@@ -13,6 +13,13 @@
         columns = [],
         data = [];
 
+
+      var myCustom = function (row, cell, value, col) {
+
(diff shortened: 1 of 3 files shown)


ericangeles:
The only thing that I noticed was the alert icon in `Action` column has 3 pixels away from the edge of the padding. Here's the difference from `Action` column and `Alert w Text` column.

`Action` Column
action

`Alert w Text` Column
alert-w-text

PR #6714, opened 2022-08-08
6490 - Fix header icon tooltip not working when changing row heights

ericangeles:

This is a follow-up fix where the header tooltip icon doesn't work when changing row heights.

Closes 
- #6490 

- Go to http://localhost:4000/components/datagrid/example-header-icon-with-tooltip.html
- Test first the header icon tooltip after the page load
- Hover on the info icon under `Product Name` column header
- It should show the tooltip content with a `Product Name` header
- Change row height to small
- Hover again on the info icon
- It should show the tooltip content with a `Product Name` header

~~- [ ] A note to the change log.~~

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -1500,7 +1500,7 @@ Datagrid.prototype = {
       // If header text is center aligned, for proper styling,
       // place the sortIndicator as a child of datagrid-header-text.
       const svgHeaderIcon = `
-        <svg class="icon datagrid-header-icon" focusable="false" aria-hidden="true" role="presentation">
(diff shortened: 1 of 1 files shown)


ericangeles:
Patch for 4.66?

tmcconechy (approved):

tmcconechy:
@ericangeles Yeah i put this on the 4.66 branch for the next patch in case it is needed (it might be).

yohannahbautista (approved):

pwpatton:
@tmcconechy This doesn't appear to add changes for lazy loading of the tooltip. With a soho-tooltip we have the ability to set a beforeShow() callback function that lets us set the title text before the the tooltip renders.

tmcconechy:
OK, That was not specified...  I think it uses its own logic to accomplish it so not that easy.. I would suggest making a new issue to make this work with lazy loading and we can add that next.

PR #6728, opened 2022-08-11
6697 - Updated datagrid header CSS height

### LINKED ISSUE #6697: Datagrid: Render issue with frozenColumns and a column with textOverflow of 'ellipsis' in height of 'extra-small'
**Describe the bug**
A render issue when using frozenColumns and a column has a property of textOverlow: 'ellipsis' when using a row height of 'extra-small'.

**To Reproduce**
1. Using the attached 'example-frozen-columns.html'
2. Notice that the height of the header is incorrect for the columns that are scrollable

**Expected behavior**
Expect the grid header has a consistent height.

**Version**
- ids-enterprise: v4.64.4

**Screenshots**
image

**Additional context**
Have a Landmark regression ticket attached to this issue.  Problem started in v4.64.x.

yohannahbautista:

Updated css for datagrid header height to fix the uneven headers in frozen columns

Closes #6697 

- Go to: http://localhost:4000/components/datagrid/test-frozen-columns-text-overflow.html
- Minimize screen if needed to show scroll
- Change the row heights, check each size
- Header height should be even

### DIFF
--- a/app/views/components/datagrid/test-frozen-columns-text-overflow.html
+++ b/app/views/components/datagrid/test-frozen-columns-text-overflow.html
@@ -17,11 +17,11 @@
         columns.push({ id: 'selectionCheckbox', sortable: false, resizable: false, formatter: Formatters.SelectionCheckbox, align: 'center' });
         columns.push({ id: 'productId', name: 'Product Id', field: 'productId', formatter: Formatters.Text, sortable: true});
         columns.push({ id: 'productName', name: 'Product Name', field: 'productName', formatter: Formatters.Hyperlink, href: 'http://www.google.com', target: '_blank'});
-        columns.push({ id: 'activity', name: 'Activity', field: 'activity'});
+        columns.push({ id: 'activity', name: 'Quantity', field: 'activity'});
 
         columns.push({ id: 'quantity', name: 'Quantity', field: 'quantity', singleline: true, textOverflow: 'ellipsis'});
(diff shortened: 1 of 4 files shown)


tmcconechy (changes requested):
@yohannahbautista I can reproduce the original issue only in 4.65.x and 4.64.x in extra small row height.

So i think we will need a fix against one of them. You can branch off 4.65.x and do a fix that works in all row heights. And we can push that back (or better if you try both to ensure the fixes are right)

Also i noticed in this branch the quantity column is not aligned with the other columns (more obvious in medium but a little off in all of the row heights).
Screen Shot 2022-08-10 at 4 24 05 PM

tmcconechy (dismissed):

tmcconechy:
@yohannahbautista UPDATE: I fixed this issue here on the PR, you may need to adjust some visual tests?
I also put the fix on 4.64 and 4.65 as this was time sensitive to finish today due DTO...

Thanks!

tmcconechy (approved):

ericangeles:
Are we going to patch it today @tmcconechy ?

tmcconechy:
Yep Done.

ids-enterprise@13.5.6 -> with 4.64.5

ids-enterprise@14.1.2 -> with 4.65.4

it has a very similar fix as this one, should be cool, but we can continue to test this PR as normal no need to fix anywhere but here. this is probably good to merge
 for me

ericangeles (approved):

PR #6758, opened 2022-08-18
Cd/release env

corpulent:
Environment deployment pipeline.

https://jira.infor.com/browse/HLO-302


.github/workflows/ci.yml
```diff
@@ -1,73 +1,122 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
-name: Build Check
+name: Deploy environment
 
 on:
   push:
-    branches: [ main ]
+    branches:
```
tmcconechy, line 5: Is this right? Seems like that means any branches we push will deploy?

tmcconechy (dismissed):
Would like to discuss the issue if I read it right every single one here would deploy https://github.com/infor-design/enterprise/branches and no one would need any of them

Maybe we only need to do this on main or via some branch pattern? Either that or need to communicate NOT to push your branch for no reason. I dont think i see the need for every branch... But willing to discuss

.github/workflows/ci.yml
corpulent: Yes this is correct.

.github/workflows/ci.yml
```diff
@@ -1,73 +1,121 @@
...
```
tmcconechy, line 5: Are we going to change this then? Not thinking its right too just release everything we push

danielortiz1982 (dismissed):

tmcconechy:
The only problem i have with this PR now are some of the rules on the package.json. It is common that you will need a few other changes to support the changes in the package.json and this check doesn't seem to allow that.

For example https://github.com/infor-design/enterprise/pull/6731 would not be possible, some things change in package json that need other changes. Because of that this seems too restrictive. Maybe we can do this as a pr review note/manual check?

danielortiz1982 (approved):

tmcconechy (approved):

Issue #6748, opened 2022-08-17
Bar Charts Grouped: Identical aria labels for bar chart


JBismonte:
**Describe the bug**
the aria labels for bar chart grouped is the same, the label for the first bar is the same for all bars.

**To Reproduce**

1. Go to http://localhost:4000/components/bar-grouped/example-formatter.html
2. Inspect element for each of the bars.
3. See that they have identical aria labels.

**Expected behavior**
aria label should be different for each bar.

**Version**

- ids-enterprise: 4.67

**Screenshots**
See screen recording 

https://user-images.githubusercontent.com/29673378/185085568-0a7dee15-d64d-48e1-bc6e-69b1ca39f507.mp4

**Platform**
 - Infor Application/Team Name: Landmark UI
 - Device: (if applicable: NA
 - OS Version: Windows 11
 - Browser Name: Firefox
 - Browser Version: 103.0.2 (64-bit)

**Additional context**
Infor Jira Link
https://jira.infor.com/browse/LMWEB-1972

glenlieorillo:
QA tested and passed.
http://localhost:4000/components/bar-grouped/example-formatter.html

https://user-images.githubusercontent.com/102151352/193491666-869c4146-c483-4813-94ae-a5200219b69a.mov

PR #6838, opened 2022-09-21
6748 - Bar Charts Identical Arias

tjamesallen15:

This PR will fix a bug in bar charts grouped, where arias are identical to each series.

Closes https://github.com/infor-design/enterprise/issues/6748

- Go to http://localhost:4000/components/bar-grouped/example-formatter.html
- Inspect element for each of the bars.
- See that they have identical aria labels.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## v4.68.0 Fixes
 
+- `[Bar Chart]` Fixed a bug in bar charts grouped, where arias are identical to each series. ([#6748](https://github.com/infor-design/enterprise/issues/6748))
 - `[Datagrid]` Fixed alignment issues in trigger fields. ([#6678](https://github.com/infor-design/enterprise/issues/6678))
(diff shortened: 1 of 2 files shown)


tmcconechy (changes requested):
Re-reading the issue the issue is that they all should not have the same label. Also i think having "item value" for the aria label isnt very useful. I think it would be better to put the value in there (same value in the tooltip i.e. 4.58, 2.46 ect). That way this is what the screen reader will tell you and it will tell you the bar info.

tmcconechy:
@tjamesallen15 i think its better to be an aria-label because aria-describedby would be an ID that should point to something https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

I also looked into this and i think it should be aria-label = "Schedule Adherence By Quantity 4.5" 
I.E. the legend label as well as the value for each bar.

ericangeles:
@tjamesallen15 let me know if you need help on this, let's have a quick call tomorrow.

yohannahbautista (approved):

tmcconechy (approved):

PR #6843, opened 2022-09-21
6530 - add setting ariaDescribedBy in the column to override the aria-describedby value

### LINKED ISSUE #6530: Datagrid: redundant aria-describedby attributes at cells.
**Describe the bug**
Table cells each have an aria-describedby attribute which references their column header. This information is redundant. 

**To Reproduce**

1. Go to http://localhost:4000/components/datagrid/test-landmark.html
2. Inspect the cell, and check that the td attribute has a redundant aria-describedby attribute

**Expected behavior**
The datagrid cells don't need a redundant aria-describedby attribute.

**Version**

- ids-enterprise: 4.64

**Screenshots**
image

**Platform**
 - Infor Application/Team Name: Landmark 
 - Device: (if applicable): NA
 - OS Version: Windows 10
 - Browser Name: Firefox
 - Browser Version: 100.0.2 (64-bit)

**Additional context**
https://jira.lawson.com/browse/LMCLIENT-40109

ericangeles:

This PR allows overriding the `aria-describedby` value in the cells by adding the `ariaDescribedBy` setting in the column.

Closes #6530 

- Go to http://localhost:4000/components/datagrid/test-landmark
- Inspect the cells in `Product Name` column `td` element
- The `aria-describedby` values of all cells should be `id-row-1-cell-1 `and incrementing +1 e.g., `id-row-2-cell-1`
- Check also the `Manufacturer` Column (the first one)
- It should also be the same as the `Product Name` column but the incrementing part is in the cell, eg., `id-row-1-cell-1`, `id-row-1-cell-2`, etc.
- The rest of the column cells should start in `test-landmark`

### DIFF
--- a/app/views/components/datagrid/test-landmark.html
+++ b/app/views/components/datagrid/test-landmark.html
@@ -5,6 +5,9 @@
 
 <script>
   $('body').one('initialized', function () {
+    var ariaDescribedBy = function (row, cell) {
+      return `id-row-${row}-cell-${cell}`;
+    }
     var settings = {
       "actionableMode": true,
       "cellNavigation": true,
@@ -35,7 +38,8 @@
             "align": "left",
             "hideable": true,
             "textOverflow": "ellipsis",
-            "formatter": Formatters.Readonly
+            "formatter": Formatters.Readonly,
+            "ariaDescribedBy": ariaDescribedBy
           },
         {
           "id": "Manufacturer0",
@@ -45,7 +49,8 @@
           "name": "Manufacturer (Ed, Ell)",
           "filterType": "lookup",
           "textOverflow": "ellipsis",
-          "editor": Editors.Lookup
+          "editor": Editors.Lookup,
+          "ariaDescribedBy": ariaDescribedBy
         },
         {
           "id": "Manufacturer1",
--- a/app/views/components/datagrid/test-landmark.html
+++ b/app/views/components/datagrid/test-landmark.html
@@ -6,7 +6,7 @@
 <script>
   $('body').one('initialized', function () {
     var ariaDescribedBy = function (row, cell) {
-      return `id-row-${row}-cell-${cell}`;
+      return `test-landmark-datagrid-${row}-header-${cell}`;
     }
     var settings = {
       "actionableMode": true,
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,8 +11,7 @@
 
 - `[Datagrid]` Fixed a bug in datagrid where dropdown filter does not render correctly. ([#6834](https://github.com/infor-design/enterprise/issues/6834))
 - `[Datagrid]` Fixed alignment issues in trigger fields. ([#6678](https://github.com/infor-design/enterprise/issues/6678))
-- `[Datagrid]` Allow beforeCommitCellEdit event to be sent for Editors.Fileupload. ([#6821](https://github.com/infor-design/enterprise/issues/6821))
-- `[Datagrid]` Add null guard in tree list when list is not yet loaded. ([#6816](https://github.com/infor-design/enterprise/issues/6816))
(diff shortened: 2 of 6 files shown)


ericangeles:
@corpulent how do we fix the issue in File check / Test changed-files here?

tmcconechy (commented):
Just a note that the pacakge-lock.json was accidentally included (maybe from the node work so should probably remove that from the PR).

Also i didnt check it yet but are the aria-described-by's possible the exact same as before (thats going to be important for them)

ericangeles:
@tmcconechy yeah, I reverted to the original, and the setting overriding works fine.

tmcconechy (changes requested):
@ericangeles just so we make sure its possible can you make the example show the exact same aria-described by's as before i was testing https://4590-enterprise.demo.design.infor.com/components/datagrid/test-landmark and its different

Screen Shot 2022-09-23 at 10 30 04 AM

Just have to maybe use the same idea as before in the example, taking the page url ect..

tmcconechy:
right, but they might want to see that you can actually make them the same as before. the problem. being that the test scripts are broken and they need all the IDs to be the exact same as before

ericangeles:
@tmcconechy all good!

tmcconechy (dismissed):
Cool looks good. I think Landmark will need the type added to NG for the new setting. So maybe go ahead and PR that as well?

tjamesallen15 (approved):

tmcconechy (approved):

vonnyw:
@ericangeles @tjamesallen15 how would one remove the ariaDescribedBy attribute? The initial ask for 508 compliance is to remove the attribute, due to references in test code, we need the setting to toggle it back and forth.

ericangeles:
@vonnyw we reverted back it to the original behavior and added a setting to override the aria-describedby value.

vonnyw:
We don't wish to override the aria-describedBy value, but remove it based on the original ticket entered https://github.com/infor-design/enterprise/issues/6530.

tmcconechy:
@vonnyw i don’t really understand. On the original fix we removed the aria. But you said you needed it for testing still. So we added. It back as a setting you can use because it is incorrect we don’t want it by default. But. Ow you can either use the setting or not. So you have the choice and could maybe toggle it as a setting or something. I don’t see what we could change, you have the choice to use it or not.

vonnyw:
@tmcconechy is there a setting to remove the attribute all together if we don't wish to use it?

tmcconechy:
Yes but you either use it or don’t use it so you would need to add to to the columns of the datagrid or not add it. (So you could do a setting and add or remove the column setting)

ericangeles:
@tmcconechy I remember we reverted back the original behavior of it (having aria-describedby property by default). The original fix was removing the property without any setting to add the aria-describedby property.

tmcconechy:
@ericangeles so its not a column setting its just on all the time now?

ericangeles:
Its a column setting but we reverted back the original behavior of aria-describedby which is by default it adds in the row. The first fix that we did was to remove it and we reverted it back with the addition of column setting to override it.

tmcconechy:
ok that was my understanding. Either way Landmark can use it or not (so effectively its a setting)

ericangeles:
I think they want to get rid of the aria-describedby and we don't have any ability to do that.

vonnyw:
@ericangeles yes, this is what I am asking for: we want to get rid of the aria-describedby and we don't have any ability to do that.

tmcconechy:
I guess I don’t understand then. Looking at this code you can specify ariaDesribedBy on the column https://github.com/infor-design/enterprise/pull/6843/files#diff-13c32b0c73b40992aa6d376fb22d6746845a3b7d6d2545e19bd30687bcabb5a5R4861

wouldn’t that constitute a setting? So you just change the column config and reload the grid ?

PR #6881, opened 2022-10-07
6145 lookup disablefix

### LINKED ISSUE #6145: Lookup: Input label stays disabled when enable API is called on Lookup.
``**Describe the bug**
Utilizing as soho-label with a soho-lookup when inside of a `<div class="field">` does not enable the soho-label correctly after disabling, then re-enabling the soho-lookup. This happens when the .field is not the first parent of the soho-lookup. 

```
<div class="field">
  <field-chooser>
    <label soho-label>Test</label>
    <input soho-lookup>
 </field-chooser>
</div>
```

**To Reproduce**
https://stackblitz.com/edit/ids-quick-start-1311-ipgosj?file=src/app/app.component.html

1. Navigate to url above.
2. Notice the Label and Lookup are disabled.
3. After 3 seconds the Lookup will become enabled. However the label will stay disabled.

**Expected behavior**
The Label and Lookup should stay in sync with disabling then re-enabling.

**Version**
4.60.0

**Screenshots**
image

**Platform**
 - Infor Application/Team Name: Infor WFM

**Additional context**
There is a small inconsistency in lookup.js enable(). enable is using .parent() instead of .closest('.field') like disable is using. I will produce a Pull Request shortly.

White-NW:
Fixed the bug where a field stays disabled when the field is not a direct parent of the lookup. Took the same logic to find the parent field from the disable function to the enable function.

Fixes #6145

Wasn't sure how to create a functional test for the issue since each functional test uses the same lookup configuration but there are multiple scenarios.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@
 - `[Searchfield]` Visual fixes on go button in searchfield toolbar. ([#6686](https://github.com/infor-design/enterprise/issues/6686))
 - `[Textarea]` Fixed a bug in textarea where validation breaks after enabling/disabling. ([#6773](https://github.com/infor-design/enterprise/issues/6773))
 - `[Typography]` Updated text link color in dark theme. ([#6807](https://github.com/infor-design/enterprise/issues/6807))
+- `[Lookup]` Fixed where field stays disabled when enable API is called ([#6145](https://github.com/infor-design/enterprise/issues/6145))
 
 ## v4.67.0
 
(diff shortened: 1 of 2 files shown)


LilBiggs:
@tmcconechy - We weren't sure on the recommended way to test this. We could add an example to one of the /lookup/test-*pages. How do you normally test something like this?

The scenario came from Angular because we wrap the lookup in an Angular component so the HTML output looks like:
`

<div class="field">
   <field-chooser type="fileType">
      <input lookup/>
  <field>
</div>`

tmcconechy (approved):

ericangeles (approved):

PR #6883, opened 2022-10-10
6844 - Added additional checks for destroying toolbar

### LINKED ISSUE #6844: Toolbar Flex: Incorrect destroy function for 'toolbarflexitem' component
**Describe the bug**
The change in <https://github.com/infor-design/enterprise/pull/6439> fixed a memory leak problem, but create new issue.
The 'toolbarflexitem' component is not destroyed correctly.
Line 22 in https://github.com/infor-design/enterprise/blob/4.67.x/src/components/toolbar-flex/toolbar-flex.item.jquery.js, it should be corrected from
```js
$.removeData(this, COMPONENT_NAME);
```
to
```js
if (this.element) $.removeData(this.element, COMPONENT_NAME);
```
It is because the data value is set in toolbar button object, not in 'instance' object. The original code `$.removeData(this, COMPONENT_NAME);` will always do nothing.

Because the component is not destroyed correctly, the toolbarflexitem cannot be re-initialised again. See line 10 in <https://github.com/infor-design/enterprise/blob/4.67.x/src/components/toolbar-flex/toolbar-flex.item.jquery.js#L10>, the instance variable will always have value. The expected result is the instance variable should be undefined after being destroyed.

After I correct the line 22 in toolbar-flex.item.jquery.js, there are other issues at line 728 and 731 in <https://github.com/infor-design/enterprise/blob/4.67.x/src/components/toolbar-flex/toolbar-flex.item.js#L728>. The 'originalButtonAPI' variable could be undefined after being successfully destroyed. When destroying the 'More' button, it will try to unlink those predefinedItems that has already been done before. Need some null point check on the two lines like below:
```
line 728:   if (originalButtonAPI) originalButtonAPI.actionButtonLink = null;
line 731:   if (originalButtonAPI && originalButtonAPI.type === 'menubutton') {
```

**To Reproduce**

Sorry that I cannot use the pure html page to reproduce the problem. I can only reproduce the issue in our GWT projects.

**Expected behavior**
The 'toolbarflexitem' component should be destroyed correctly.

**Version**

- ids-enterprise: [v4.67.1]

**Screenshots**
N/A

**Platform**
 - Infor Application/Team Name: [Infor SunSystems]
 - Device: (if applicable) [N/A]
 - OS Version: [Windows 11]
 - Browser Name: [chrome]
 - Browser Version: [105.0.5195.127 (Official Build) (64-bit) (cohort: Stable)]

**Additional context**
N/A

yohannahbautista:

Added additional checks and deletes for destroying toolbar flex.

Closes #6844 

Issue occurs in other projects (mentioned in #6844), not sure how to reproduce it in IDS.


ericangeles (approved):

tjamesallen15 (approved):

docs/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - `[Modal]` Fixed a bug where suppress key setting is not working. ([#6793](https://github.com/infor-design/enterprise/issues/6793))
 - `[Searchfield]` Visual fixes on go button in searchfield toolbar. ([#6686](https://github.com/infor-design/enterprise/issues/6686))
 - `[Textarea]` Fixed a bug in textarea where validation breaks after enabling/disabling. ([#6773](https://github.com/infor-design/enterprise/issues/6773))
+- `[Toolbar Flex]` Added additional checks for destroying toolbar. ([#6844](https://github.com/infor-design/enterprise/issues/6844))
```
ericangeles, line 31: @yohannahbautista please move this too in 4.69.0

ericangeles (approved):

tmcconechy (approved):

PR #6909, opened 2022-10-21
NG1365 datagrid value input field inline editor

tjamesallen15:

This PR will fix a bug in datagrid where inline editor input text is not being selected upon clicking.

Closes https://github.com/infor-design/enterprise-ng/issues/1365

- Go to http://localhost:4000/components/datagrid/test-editable-with-inline-editor.html
- Click on one of the input fields under Column Activity
- Text should be highlighted

### DIFF
--- a/app/views/components/datagrid/test-editable-with-inline-editor.html
+++ b/app/views/components/datagrid/test-editable-with-inline-editor.html
@@ -1,0 +1,198 @@
+
+<div class="row">
+  <div class="twelve columns">
+    <div role="toolbar" class="toolbar">
+
+      <div class="title">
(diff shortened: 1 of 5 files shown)


ericangeles:
@tjamesallen15 can you also create an e2e for this?

tmcconechy (changes requested):
only issue i see is the styling when there is an error... (clear the field). I think we need some special css for errors in inline editors.

Screen Shot 2022-10-21 at 9 37 09 AM

tmcconechy (approved):

ericangeles (approved):

Issue #6850, opened 2022-09-23
Datagrid: Cell color mismatch when editing a cell with alternate shading


s-werking:
**Describe the bug**
When editing a cell with the darker color of alternate shading, the cell has a mismatch of coloring. This happens for the following theming:
- New Light
- New Contrast
- Classic Contrast

**To Reproduce**
1. Add the setting alternateRowShading: true to http://localhost:4000/components/datagrid/example-editable.html
2. Go to http://localhost:4000/components/datagrid/example-editable.html
3. Click on a cell with shading
4. See that the cell is not one color when editing
5. Click on a cell without shading
6. See that the cell is all one color when editing

**Expected behavior**
When a user clicks on a cell with alternate shading to edit, it is expected that the cell is all one color.

**Version**
- ids-enterprise: v4.68.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/76002134/191962690-acbc2b25-2c20-490c-8c2a-290d5c75e519.png)

**Platform**
 - Infor Application/Team Name: lawson
 - OS Version: Windows 11
 - Browser Name: chrome
 - Browser Version: 105.0.5195.127 (Official Build) (64-bit)

**Additional context**
Is it possible to fit this in the 4.68 release?

tmcconechy:
Probably not as there is a big chunk left of work to do for 4.68 and only less than a week left. But if it all gets done we can try. Slotting it in for next version

glenlieorillo:
QA tested and passed.


**Windows 11 - Chrome (BrowserStack)**

https://user-images.githubusercontent.com/102151352/195321478-3e82fa06-7552-4c5f-bcb2-3c4c79e94a99.mov


**Mac Chrome**



https://user-images.githubusercontent.com/102151352/195321650-db3caf37-6610-4f98-8b12-653bc2aa94e7.mov

s-werking:
Hi, we have encountered an issue related to this ticket after implementing this fix.

Steps to reproduce:
1. Go to http://localhost:4000/components/datagrid/test-alternate-row-shading.html
2. Click on the date cell with shading on the 4th row
3. Click outside of the table
4. See that the color around the datepicker icon is different than the color of the cell
5. To see in other themes, change the theme, refresh the page, then repeat steps 2-4

Gif for reference:
![Different color background around icon](https://user-images.githubusercontent.com/76002134/196765985-7b6438b9-6191-405a-a492-2febfa28d16e.gif)

[...]

PR #6913, opened 2022-10-24
6850 - datagrid alternate row shading (2)

tjamesallen15:

This PR will fix a bug in datagrid where row shading is not rendered properly.

Closes https://github.com/infor-design/enterprise/issues/6850

- Go to http://localhost:4000/components/datagrid/test-alternate-row-shading.html
- Click on a cell with shading

### DIFF
--- a/src/components/datagrid/_datagrid.scss
+++ b/src/components/datagrid/_datagrid.scss
@@ -2346,6 +2346,7 @@ $datagrid-small-row-height: 25px;
 
         &.is-editing {
           background-color: $datagrid-cell-editing-bg-color;
+
           div.datagrid-cell-wrapper {
             svg.icon {
               @include trigger-icon-background($datagrid-cell-editing-bg-color);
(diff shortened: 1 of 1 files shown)


tmcconechy (dismissed):
@s-werking what do you think?

s-werking:
Hi @tjamesallen15, I'm seeing a darker color around the datepicker's icon on non-shaded cells in the Classic High Contrast and Classic Dark themes. This can also be seen by clicking a lookup cell on this sample (http://localhost:4000/components/datagrid/test-editable-lookup.html?theme=classic&mode=contrast&colors=133c59).

Here is a screenshot for reference:
image

tjamesallen15:
> Contributor

I think that's a box-shadow already. Correct me, if I'm wrong @tmcconechy

tmcconechy:
Is it the `ripple-effect`? It is hard to see maybe we could disable it in this case.. With a rule like

```
  .ripple-effect {
    background-color: transparent;
  }
  ```

Or similar? That targets the button in this case

tmcconechy (approved):

deep7102 (approved):

Issue #6781, opened 2022-08-25
DataGrid expanded row cannot be edited after two editing scenarios on the parent row


guymid:
**Describe the bug**
When the parent row of an expanded row contains editable fields they interfere with the editability of inputs in the expanded row. This is a defect with the state handling where the exit states from fields in the parent row are incorrect. If line 107486 of sohoxi.js is commented out the editability of the input in the expanded row is correct, but the input in the parent row is still in edit mode. The line is in commitCellEdit() and contains this code:

cellNode.removeClass('is-editing is-editing-inline has-singlecolumn'); // Editor.destroy

**To Reproduce**
This Stackblitz shows both scenarios (with steps) where editability is broken by the parent row. https://stackblitz.com/edit/ids-quick-start-1411-6rgabz?file=src/app/app.component.html

**Expected behavior**
Editable fields in the parent row should not break the editability of inputs in the expanded row.

**Version**

v4.59.2

**Platform**
 - Infor Application/Team Name: Orchestrator

tmcconechy:
Nice bug report. I think we can fix this. But there was'nt the general idea that editing would take place in the expandable row.
Might be better to put that editor in a column (not knowing the full use case)

guymid:
In this case the expandable row is for viewing and editing the description of an item. As the description can be quite long it doesn't fit in the main grid.

jbrcna:
QA Passed.v4.70.0-dev
https://main-enterprise.demo.design.infor.com/components/datagrid/example-expandable-row-editable.html

PR #6984, opened 2022-11-21
6781 - datagrid expanded row editable

tjamesallen15:

This PR will fix a bug in datagrid where expandable row input cannot edit the value.

Closes https://github.com/infor-design/enterprise/issues/6781

- Go to http://localhost:4000/components/datagrid/example-expandable-row-editable.html
- Click on one of the + icons and edit the value in the expanded row.
- Now click on either Quantity or Price and now try to edit the value in the expanded row.

### DIFF
--- a/app/views/components/datagrid/example-expandable-row-editable.html
+++ b/app/views/components/datagrid/example-expandable-row-editable.html
@@ -1,0 +1,228 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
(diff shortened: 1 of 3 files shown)


tmcconechy (approved):
I think this is working right but the PR is still draft.

1. could add some safety checks `e.currentTarget?.parentElement?.classList.contains('datagrid-expandable-row')`
2. would remove the word "dirty cell" from the PR as im not seeing the dirty indicator in this example but it doesn't seem like part of the issues

tmcconechy (approved):

Issue #6977, opened 2022-11-17
Secondary Button: Change from Slate to Solid with Border


tmcconechy:
**Describe the bug**
Based on feedback would like to change the secondary button to a blue border to a slate button.

- Classic Leave it
- Dark =- TBD
- Contrast - darker azure like the primary

**Expected behavior**

**Screenshots**
See Designs link https://www.figma.com/file/VcRDGPFqlp3MJ46mEWkGPf/Buttons?node-id=701%3A1222&t=2eRdbDDXoBLYCBwt-1

**Implementation Details**

- all button styles change to this new style
- updates can seen no https://main-enterprise.demo.design.infor.com/components/button/example-index.html?colors=7928E1 go across the examples
- personalization color changes the border
- use a class for destructive action button and show it and show it on https://main-enterprise.demo.design.infor.com/components/button/example-index.html?colors=7928E1

jbrcna:
QA failed.
`disabled primary button` text is hard to read in `Amethyst`, `Azure`, `Grapihite` using high contrast. observe in light and dark mode. 
<img width="326" alt="image" src="https://user-images.githubusercontent.com/56722184/209770190-a1d17e6c-2c61-460f-b49a-9f83bcd1ff35.png">
<img width="307" alt="image" src="https://user-images.githubusercontent.com/56722184/209770226-f35e6aa9-2191-4bda-bf93-30a73a0bb85b.png">
<img width="268" alt="image" src="https://user-images.githubusercontent.com/56722184/209770276-6b027310-2d8e-4de5-a1d5-ee7bfad83073.png">

tmcconechy:
@tjamesallen15 @ericangeles I see this as well. I think we just need to adjust the css in personalize-styles.js for the new buttons it should get the colors and look readable.

ericangeles:
@tjamesallen15 please check at this too.

janahintal:
Hi Team, some buttons still not that visible/readable in Dark mode. 

Dark mode amethyst 
![image](https://user-images.githubusercontent.com/45757016/211303619-21ef129e-6ae2-4a54-a0f4-b1bcdbbdf802.png)

Dark mode Graphite 
![image](https://user-images.githubusercontent.com/45757016/211303788-a4b8e43d-e88b-417f-83ca-64943b7abf5a.png)

Thanks, 
Janah

ericangeles:
@tmcconechy yes please

tmcconechy:
[...]

PR #7024, opened 2022-12-12
6977 - Style Changes for Existing Buttons and New Destructive Buttons

tjamesallen15:

This PR will change style and features of existing buttons and added destructive buttons.

Closes https://github.com/infor-design/enterprise/issues/6977

- Go to http://localhost:4000/components/button/example-index.html

Please refer to the requirements on:
- https://github.com/infor-design/enterprise/issues/6977
- https://www.figma.com/file/VcRDGPFqlp3MJ46mEWkGPf/Buttons?node-id=701%3A1222&t=2eRdbDDXoBLYCBwt-1


yohannahbautista:
Hello, there might need to be an adjustment in hover color when it's in different color modes as the text is hard to read
Screen Shot 2022-12-14 at 3 03 45 PM

tmcconechy (changes requested):
Looks pretty good.

- minor but it sort of looks like you added a while set of rules. Maybe some of the previous rules should actually be removed. But this might take too long so just a note.

the only issue i see is
- the hover state is missing on the enabled icon button
- the hover state is missing on the "btn-link" example on the tertiary icon
- when you look at the buttons on http://localhost:4000/ the new labels look out of place and out of the grid alignment. Maybe make it use the same classes as "Form Action Buttons" and contain it in the responsive grid:

Screen Shot 2022-12-14 at 10 40 28 AM

- also looking at http://localhost:4000/ -> the menu button should work like the new "tertiary"

tmcconechy (approved):
This looks very good to me :) Thanks @tjamesallen15

deep7102 (approved):

tmcconechy:
@tjamesallen15 can you just check the two tests. If something broke? Or if they need updating? Or maybe we just skip them?

ericangeles:
I noticed this from a mobile perspective. I don't think we have a design for this yet. This is where the visual test fails.

Screen Shot 2022-12-17 at 4 06 20 AM

ericangeles:
Think we can skip it for now.

ericangeles (approved):

tmcconechy (approved):

PR #7056, opened 2022-12-27
4319 - Added modal settings to lookup

### LINKED ISSUE #4319: Lookup: Add option to use full screen in small browsers
A regular modal has the option to be used in full screen mode in smaller browsers. We would like to have this option for Lookups as well. Currently when using a lookup component and seeing it on a mobile screen the Lookup content is often to wide to show it in portrait mode and in landscape mode no content can be seen. If the Lookup was shown in full screen mode I believe it would be easier to use on a phone or other device with a small screen or lower resolution.  

I think the Lookup component would work better for small screens if an option to view it in full screen was added similar to or equal to the Modal options with the fullsize='responsive' option and the breakpoint setting. I think it's very good if it has the same responsive behaviour as in the Modal.

Tim also suggested using a Custom Modal in the Lookup, but I believe it would be easier to use with a Lookup option than with a custom modal, specially if there are more than one Lookup to update.

**Additional context**
This is not urgent from our side, but there are some application teams that use Lookups in their widgets in Homepages and I believe this option would be good to improve the user experience on phones and small tablets.

And what application do you work on? Homepages

yohannahbautista:

Added modal settings to lookup

Closes #4319 

- Go to: http://localhost:4000/components/lookup/example-index.html
- Minimize screen width to around 767px or lower
- Open lookup
- Should be full size 

### DIFF
--- a/app/views/components/lookup/example-index.html
+++ b/app/views/components/lookup/example-index.html
@@ -52,6 +52,10 @@
       field: 'productId',
       autoApply: true,
       autoFocus: false,
+      modalSettings: {
+        fullsize: 'responsive',
+        breakpoint: 'phone-to-tablet'
+      },
       options: {
         columns: columns,
         dataset: data,
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## v4.71.0 Features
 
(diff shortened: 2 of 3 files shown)


tmcconechy (approved):

deep7102 (approved):
Looks good 🍌

PR #7062, opened 2022-12-29
NG1401 - Added paste event listener for textarea

yohannahbautista:

Added paste event listener for textarea

Closes https://github.com/infor-design/enterprise-ng/issues/1401

- Go to: http://localhost:4000/components/textarea/example-index
- In 'Notes (maxlength)' textarea, copy-paste the following string:
```
Lorem ipsum dolor sit amet, consectetur
adipiscing elit. Donec

rutrum hendrerit nunc sed mollis...
```
- Should not paste the entire string
- Delete a character and type it back
- Should be able to type


deep7102:
Looks good, one thing saw when text area already have some value, then paste new stuff if its value reach to max. It replace the old content thinking it should let it paste the new content at its caret place then check if need to trim it should begin trim from the end of the value.

tmcconechy:
Seems good - aside from @deep7102 's comment. Maybe its just a logic issue here test/baseline-images/diff/

deep7102 (approved):
Looks good 🍌

ericangeles (approved):

tmcconechy:
Some tests seem a little suspicions. I would check them but feel free to disable if irrelevent.

PR #7129, opened 2023-01-25
7128: Use dataRowIdx also for summary rows.

### LINKED ISSUE #7128: Summary will be selected
**Describe the bug**
When having a data grid with summary row and enabled option to select a row, then the summary row will also be selected, when you select the first row in data grid. 

**To Reproduce**
1. git clone https://github.com/infor-design/enterprise.git
2. cd enterprise
3. change the file `app/views/components/datagrid/example-summary-row.html` and add selectable option
```
@@ -35,6 +35,7 @@
         columnReorder: true,
         dataset: data,
         summaryRow: true,
+        selectable: 'single',
         toolbar: {title: 'Compressors', actions: true, rowHeight: true, results: true }
       });
     });
```
4. npm ci
5. npm run start
6. When IDS server is running open a browser with the shown URL and append `components/datagrid/example-summary-row.html`to open the changed HTML file.
7. Select and unselect the first row and maybe other rows

**Expected behavior**
The summary should never be selected.

**Version**
- ids-enterprise: v4.70.0

**Screenshots**
image

**Additional context**
This bug was introduced by solving #6746.

swuendri:
This PR solves the issue #7128 by using dataRowIdx for summary rows instead of actualIndex, which is null in that case.

Fixes #7128

See the steps in issue, you should not be able to reproduce the bug.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -29,6 +29,7 @@
 - `[Datagrid]` Fixed a bug in datagrid where flex toolbar is not properly destroyed. ([NG#1423](https://github.com/infor-design/enterprise-ng/issues/1423))
 - `[Datagrid]` Fixed a bug in datagrid in datagrid where the icon cause clipping issues. ([#7000](https://github.com/infor-design/enterprise/issues/7000))
 - `[Datagrid]` Fixed a bug in datagrid where date cell is still in edit state after editing when using Safari. ([#6963](https://github.com/infor-design/enterprise/issues/6963))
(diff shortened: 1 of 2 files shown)


tmcconechy (dismissed):
Looks good, thanks for the contribution.

- can you just add a note in https://github.com/infor-design/enterprise/blob/main/docs/CHANGELOG.md in the bugs/4.71.0 section

🍻

swuendri:
CHANGELOG is also update now @tmcconechy. Thank you for the hint!

tmcconechy (approved):

ericangeles (approved):

PR #7132, opened 2023-01-26
7041 - Added left and right key traversal for searchfield

### LINKED ISSUE #7041: Header: Keyboard not working
**Describe the bug**
At some stage keyboard stopped working on the header component. Code might be in the toolbar flex component and the header example should use the toolbar-flex.

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/header/example-index.html
2. Tab into the header
3. try to use arrow keys to navigate (see expected behavior)

**Expected behavior**
For that you should test as follow:
- tab into the header
- next tab goes out of the header
- once in focus on a button in the header left and right arrows should move between the header buttons
but if you move into a search field because you can use arrows in the field up and down arrow should move next/previous. up/down arrow can also be used as left/right on buttons
- header examples can use toolbar flex iinstead

**Version**
- ids-enterprise 4.70

yohannahbautista:

Include left and right keys in traversal for searchfield

Closes #7041 

- Go to: http://localhost:4000/components/header/example-flex-toolbar.html
- Tab into the header
- Try to use arrow keys to navigate

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,7 @@
 - `[Colors]` Re-add yellow alerts.([#6922](https://github.com/infor-design/enterprise/issues/6922))
 - `[Chart]` Added 'info' and theme color options in settings.([#7084](https://github.com/infor-design/enterprise/issues/7084))
 - `[Icons]` Added three new icons: `icon-paint-brush, icon-psych-precaution, icon-observation-precaution`. ([#7040](https://github.com/infor-design/enterprise/issues/7040))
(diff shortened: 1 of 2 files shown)


tmcconechy (changes requested):
@yohannahbautista looks good except for the search field...

a) when you focus the search field it "changes size and moves"
b) because of this fix you cant use the left and right arrow when in the search input (to move around the text). To solve this we before had ignored the left and right arrow key when in the field. Instead you use up arrow and down arrow to move out (up goes left and down goes right)

tmcconechy:
Still see it changing size when focusing the search - the keyboard behavior looks good 👍🏻

EdwardCoyle (approved):

tmcconechy (approved):

PR #7155, opened 2023-02-02
7005 - Fix on inverse colors not showing in popupmenu in masthead

### LINKED ISSUE #7005: Popupmenu: Inverse is not applied in masthead if parent element doesn't have a class
**Describe the bug**
When using a popupmenu inside a masthead, the "inverse" class is not added to the popupmenu-wrapper element unless the direct parent of the popupmenu has a `class="..."` attribute.

This is caused by the `containerClass !== undefined` check in popupmenu.js: https://github.dev/infor-design/enterprise/blob/6b5ca490227a44d06a275db844c025611dfe2e4b/src/components/popupmenu/popupmenu.js#L374

**To Reproduce**
When using the Angular directives, the following code should reproduce the problem:

```html
<nav class="masthead">
  <!-- The following will correctly create a dark popupmenu (a .popupmenu-wrapper element with .inverse) -->
  <div class="does-not-matter"> <!-- The class attribute can contain anything. -->
    <button soho-menu-button menu="menu-dark"></button>
    <ul soho-popupmenu id="menu-dark">
      <!-- items... -->
    </ul>
  </div>

  <!-- The following will incorrectly create a light popupmenu missing the .inverse class. (assuming light theme) -->
  <div> <!-- <--- Note the absence of a class attribute -->
    <button soho-menu-button menu="menu-light"></button>
    <ul soho-popupmenu id="menu-light">
      <!-- items... -->
    </ul>
  </div>
</nav>
```

**Expected behavior**
A popupmenu that is declared inside a .masthead should always be "inverse", regardless of whether the parent element has a class or not.

**Version**

- ids-enterprise: 4.69.0

**Screenshots**
Issue:
image

Workaround:
image

With Workaround
image

**Platform**
All

**Additional context**
The issue was found when the popupmenu was declared directly inside an Angular component template. This component does not use classes for styling, so the host element does not naturally have a class attribute. Although we added one as a workaround for this issue.

Note that we have only experienced this issue when using `ids-enterprise-ng`, but the actual bug is in `ids-enterprise` so I created the issue here.

yohannahbautista:

Removed containerClass undefined check

Closes #7005 

- Go to: http://localhost:4000/components/popupmenu/test-masthead
- Click on "No Class"
- Click on "Has Class"
- Both popups should use inverse colors

### DIFF
--- a/app/views/components/popupmenu/test-masthead.html
+++ b/app/views/components/popupmenu/test-masthead.html
@@ -1,0 +1,26 @@
+<nav class="masthead">
+  <div class="buttonset" style="display: flex">
+    <div>
+      <button id="no-class" class="btn-menu">
+        <span>No Class</span>
+      </button>
+      <div class="popupmenu-wrapper">
+        <ul class="popupmenu">
+          <li><a href="#">Item One</a></li>
+          <li><a href="#">Item Two</a></li>
+        </ul>
+      </div>
+    </div>
+    <div class="field">
+      <button id="has-class" class="btn-menu">
+        <span>Has Class</span>
+      </button>
+      <div class="popupmenu-wrapper">
+        <ul class="popupmenu">
+          <li><a href="#">Item One</a></li>
+          <li><a href="#">Item Two</a></li>
+        </ul>
+      </div>
+    </div>
+  </div>
+</nav>
--- a/app/views/components/popupmenu/test-masthead.html
+++ b/app/views/components/popupmenu/test-masthead.html
@@ -1,5 +1,5 @@
 <nav class="masthead">
-  <div class="buttonset" style="display: flex">
+  <div class="buttonset" style="display: flex; padding-top: 2px">
     <div>
       <button id="no-class" class="btn-menu">
         <span>No Class</span>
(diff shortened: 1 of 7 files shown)


tmcconechy (changes requested):
@yohannahbautista the popup looks good now but the colors on the button are incorrect for some reason?

1. text is barely visible
2. the click state is showing the default button state - should match the darker color
3. also i see on http://localhost:4000/components/masthead/test-popups-on-each-side.html there is a little blue line on the button.
4. also on http://localhost:4000/components/masthead/test-popups-on-each-side.html the search field looks to close to the bottom

There is some styles for it on https://www.figma.com/file/ok0LLOT9PP1J0kBkPMaZ5c/IDS-Component-File-v4.6-(Draft)?node-id=119%3A148&t=GVN3StvDkRg9yOkh-0 

But i think we should try slate 01 or 02 and then white on the hover...

tmcconechy:
@yohannahbautista tested again...see a couple things..

1. On http://localhost:4000/components/popupmenu/test-masthead is the colors the same as http://localhost:4000/components/masthead/test-popups-on-each-side.html ? because they look passable on the later but the first one they dont..
2. On http://localhost:4000/components/popupmenu/test-masthead try changing theme /mode - some modes are invisible like contrast. For themes this component just stays the same and shouldn't change
3. Some alignment issues on http://localhost:4000/components/masthead/test-popups-on-each-side.html
- the button sits a bit low
- the search isnt centered 100% when expanded

tmcconechy:
- i still see the issues on http://localhost:4000/components/masthead/test-popups-on-each-side.html
- i tried to fix the change log conflict but the online git tool wasnt working. So maybe also fix the conflict locally

tmcconechy (approved):

ericangeles (approved):

PR #7213, opened 2023-02-14
NG 1431 - Fixed bug in Safari where dynamically switching from RTL to LTR doesn't update all the alignments

yohannahbautista:

Updated dir to ltr instead of removing dir attribute when changing locales to make Safari update alignments.

Closes https://github.com/infor-design/enterprise-ng/issues/1431

- Open Safari
- Go to: http://localhost:4000/components/datagrid/test-rtl-swap
- Click on Arabic button
- Check alignments. Click on editable fields and check alignments
- Click on English button
- Check alignments. Click on editable fields and check alignments

### DIFF
--- a/app/views/components/datagrid/test-rtl-swap.html
+++ b/app/views/components/datagrid/test-rtl-swap.html
@@ -1,0 +1,87 @@
+
+<div class="row">
+  <div class="twelve columns">
+    <div role="toolbar" class="toolbar">
+      <div class="title">
+        Data Grid Header Title
+        <span class="datagrid-result-count">(N Results)</span>
+      </div>
+
+      <div class="buttonset">
+        <button id="btn-eng" class="btn btn-secondary" type="button">English</button>
+        <button id="btn-ar" class="btn btn-secondary" type="button">Arabic</button>
+      </div>
+    </div>
+
+    <div class="contextual-toolbar toolbar is-hidden">
+      <div class="title selection-count">1 Selected</div>
+      <div class="buttonset">
+        <button class="btn-icon" type="button" id="remove-btn">
+          <svg class="icon" focusable="false" aria-hidden="true" role="presentation">
+            <use href="#icon-delete"></use>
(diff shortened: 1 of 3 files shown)


tmcconechy (commented):
@yohannahbautista just wondering if there is any other solution (maybe with css) because the attribute really shouldnt need to be there if its LTR since a) there is no css against that and b) its the default?

I tried without it and the only issue i noticed with the datagrid column dividers? Is that the only issue?

tmcconechy (changes requested):
Wondering if there is another fix other than adding the attribute when its not need. Since its its already the default it doesn't seem like it should be needed. Maybe a css solution or more component specific solution?

ericangeles:
It looks like for some reason in Safari, when switching back to English, the datagrid doesn't go back to its LTR look. As a result, the `showLabel` column remains in the left section. That's why the current CSS rules does not work.

ericangeles:
@tmcconechy checking the example page and the issue is reproducible there. I don't see any way to fix it via CSS only (as I've mentioned above the behavior of current issue) and am okay with the fix, not unless it's a breaking change.

Or we might want to just isolate the adding attr `ltr` in safari only.

e.g.,

```
else if (self.isSafari) {
  html.attr('dir', 'ltr');
}
```

Open for discussion.

yohannahbautista:
Hello, it's not just datagrid that doesn't render properly, toolbar has an issue too. If I run the page without the change:
(added another toolbar to test other components)

Original page
image

After clicking Arabic, then English
image
The toolbars don't fully go back to its previous alignment.

tmcconechy:
Ok Wierd. @ericangeles @yohannahbautista if thats the only way I guess its minor that it adds the attribute. Safari is wierd, i just felt like it shouldnt be needed since its the default but if thats the only way I'll approve

tmcconechy (approved):

deep7102 (approved):

PR #7237, opened 2023-02-17
7232 - Fix hover color in is list/mixed mode

### LINKED ISSUE #7232: Datagrid: Hover color is wrong when in new/light theme in an is-list datagrid with tall rows
**Describe the bug**
In a datagrid with a card view where the dastagrid is an `is-list` the hover color is incorrect, should be the same as the activated color. 

**To Reproduce**
1. Put example in place in components/datagrid (see attached example below)
2. open example http://localhost:4000/components/datagrid/test-card-cell-is-list.html?theme=new&mode=light
3. Select a row and the hover over it
4. Notice a 2nd color is bleeding through

image

**Expected behavior**
Hover color should be the same as the activated color. 

**Example**
[test-card-cell-is-list.html.txt](https://github.com/infor-design/enterprise/files/10760959/test-card-cell-is-list.html.txt)

**Version**
- ids-enterprise: v4.80.x, 4.81.x

**Screenshots**
These rules seem to be the culprit.
image

tmcconechy:

When rows are very tall the hover color and selection color for mixed mode would blend incorrectly. This is a fix

Fixes #7232 

- go to http://localhost:4000/components/datagrid/test-card-cell-is-list.html
- hover the rows
- then click the row (not the checkbox) to activate
- should be no bleed through of either state
- test also http://localhost:4000/components/datagrid/test-tree-mixed.html

### DIFF
--- a/app/views/components/datagrid/test-card-cell-is-list.html
+++ b/app/views/components/datagrid/test-card-cell-is-list.html
@@ -1,0 +1,137 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid"></div>
+  </div>
+</div>
+
+<style>
+  .custom-card {
+    padding: 18px;
+  }
+
+  .custom-card h3 {
+    color: #1a1a1a;
+    font-size: 16px;
+    line-height: 18px;
+    padding-bottom: 14px;
+  }
+
+  .custom-card dl>div {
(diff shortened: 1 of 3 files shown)


pwpatton:
Approved

ericangeles (approved):

PR #5440, opened 2021-07-19
986 - Fix selection for server side and paging with lookup

deep7102:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
Fixed selection for server side and paging was not working with Lookup.

**Related github/jira issue (required)**:
Fixes infor-design/enterprise-ng#986

**Steps necessary to review your pull request (required)**:
Pull this branch and build/run the demo app
Navigate to: http://localhost:4000/components/lookup/test-multiselect-description-only.html

scenario 1
- Refresh the page
- Open lookup modal and select anything
- Click on `Show Selected` and see log in console should show results

scenario 2
- Refresh the page
- Open lookup modal and select single record on any page
- After modal close, click on `Show Selected` should show one record in array
- Open again modal and unselect the same record
- After modal close, click on `Show Selected` should show empty array

scenario 3
- Refresh the page
- Open lookup modal and select first record of the first page
- Go to 2nd page and select first records and click on `Apply`
- Now again open modal and uncheck the first page record and click `Apply`
- Click on `Show Selected` and see log in console should show results

**Included in this Pull Request**:
- [x] A note to the change log.

PR #7245, opened 2023-02-21
NG1409 - lookup fix selection async multiple

tjamesallen15:

This pull request will fix a bug in lookup where items are not selected for async data.

Closes https://github.com/infor-design/enterprise-ng/issues/1409

- Go to http://localhost:4000/components/lookup/test-lookup-source-async.html
- Open the `Lookup with source`
- Notice that the correct items are selected in the dialog
- Open the `Lookup with async source`


tjamesallen15:
This is still WIP. I'm still checking it. Thanks!

tjamesallen15:
Hello Team, @tmcconechy

I managed to fix this but happen to stumble an old ticket in relation to this. Here is the ticket.
https://github.com/infor-design/enterprise/pull/5440

@ericangeles and I discussed to test it thoroughly and both the "main" and my "local" where my current fix for https://github.com/infor-design/enterprise-ng/issues/1409 resides. Upon checking my solution didn't break the solution that was made on that ticket.

But I saw an bug in the link while testing it
http://localhost:4000/components/lookup/test-multiselect-description-only.html

As you can see below, I just only followed what the instructions are listed and I produce a bug that when I select a `Product` after clicking `Apply` and `Show Selected` then open up the modal, we can see that there are 4 soxtest that was already clicked even though I only clicked one previously with an additional error in the logs.

image

I suggest that we make a separate ticket for it.

Anyway, my PR is ready for review. Thank you.

ericangeles (approved):

tmcconechy (approved):
Looks good 👍🏻

Issue #6528, opened 2022-06-01
ListView: Better translations for selected label


javenturina:
We are using the List view component with the single select configuration in our widget. Its working  and translations are being applied as well (by default). However our translation team noted that the translation are not correct in the context of our application. With this, can we control the source of the text via code/setup?

**What are you trying to do**
customize the value of the "Selected" text from component based(default) to code controlled (via variable?)

**Version**
 IDS Enterprise 4.62.1
 IDS Enterprise NG 13.3.

**Screenshots**
English:

![image](https://user-images.githubusercontent.com/91868168/171373531-295d2738-a208-4ee5-b6cc-7243c2130b20.png)

Japanese

![image](https://user-images.githubusercontent.com/91868168/171373637-e12256bf-26a0-4028-92cc-c92f31ce710f.png)

cpinatel7:
Currently "Selected" is appended to a number therefore translations are incorrect. A variable is required e.g.

%1 Record(s) Selected
Selection: %1

Thanks

tmcconechy:
We can be default change this to turn toggle between single and more than one selected. Would be a change around here https://github.com/infor-design/enterprise/blob/main/src/components/listview/listview.js#L1129

You can also try to manually add the title 
```
<div class="contextual-toolbar toolbar is-hidden">
      <div class="title selection-count">0 Selected</div>
```

But i think it will get overwritten so we probably need a change

javenturina:
thanks @tmcconechy  we will wait for the update.

tmcconechy:
So the solution is to use `Selection: %1` instead of `1 Selected`? This is anywhere with contextual toolbar.

cpinatel7:
Selection: %1
(where %1 is a number)
or
%1 Record(s) Selected

Thanks

tmcconechy:
Ok lets go with `{0} Record(s) Selected` for the new string.

cpinatel7:
Thanks for "{0} Record(s) Selected". If you can, please add a comment indicating that {0} is a number. Thanks again

cpinatel7:
@tmcconechy 
Is this ticket ready for retest? Are translations available?
Thanks

tmcconechy:
@cpinatel7 should it be one label `{0} Record(s) Selected` or two labels `{0} Record Selected` vs `{0} Records Selected`

cpinatel7:
[...]

PR #7247, opened 2023-02-21
6528 - Listview Translation

tjamesallen15:

This pull request will add an additional translation for records selected in listview.

Closes https://github.com/infor-design/enterprise/issues/6528

- Go to http://localhost:4000/components/listview/example-multiselect.html
- Select an item
- It should be `Record(s) Selected shown`


deep7102 (approved):

src/components/locale/cultures/en-US.js
```diff
@@ -449,6 +449,7 @@ Soho.Locale.addCulture('en-US', {
     Range: { id: 'Range', value: 'Range', comment: 'Range for tooltip' },
     RecordsPerPage: { id: 'RecordsPerPage', value: '{0} Records per page', comment: 'Shows an actual number of pages in the pager' },
     RecordsPerPageNoToken: { id: 'RecordsPerPageNoToken', value: 'Records per page', comment: 'Shows as a title list of values for page number (without the slot/token value)' },
+    RecordsSelected: { id: 'RecordsSelected', value: 'Record(s) Selected', comment: 'text describing a selected objects' },
```
tmcconechy, line 452:
```suggestion
    RecordsSelected: { id: 'RecordsSelected', value: 'Record(s) Selected', comment: 'Text the number of items selected in a list component' },
```

tmcconechy (changes requested):
@tjamesallen15 see this comment
https://github.com/infor-design/enterprise/issues/6528#issuecomment-1438727974
We will need to check the count and show the correct label based on the count.

tjamesallen15:
> @tjamesallen15 see this comment [#6528 (comment)](https://github.com/infor-design/enterprise/issues/6528#issuecomment-1438727974) We will need to check the count and show the correct label based on the count.

copy!

ericangeles (dismissed):

tmcconechy (changes requested):
@tjamesallen15 i think we need to use 3 strings exactly as follows:

`0 Records Selected` -> (In English, "0 Records" is plural but in most languages, 0 is singular)
`1 Record Selected`
`{0} Records Selected` -> more than one

We can't just reuse the same one because in some languages 0, 1 or multiple "things" are translated different. So expect the logic to have three things and use the right one if its 0, 1, or more.

tmcconechy (approved):

ericangeles (approved):

Issue #5673, opened 2021-09-28
Badge/Tag - Update colors for better accessibility contrast


tmcconechy:
**Describe the bug**
Make some style changes to badge and tag colors. 

**Fixes**
https://www.figma.com/file/gHmapzCfCqqPRv5RnFfGcE/?node-id=0%3A1
- [ ] Change colors in all 3 variants (new)
- [ ] Update the badge and tag examples to match
- [ ] add status colors to https://main-enterprise.demo.design.infor.com/components/colors/example-index.htm
- [ ] Make changes in the web component too
- [ ] Make sure tokens are in sync

**Additional context**
https://jira.infor.com/browse/UXD-1865

janahintal:
QA Failed. 
Light UI Expected color for Status 04 is #1F9254
Result: #1C7F49

Figma: https://www.figma.com/file/gHmapzCfCqqPRv5RnFfGcE/Q3-Accessibility-Tests
![image](https://user-images.githubusercontent.com/45757016/149758568-139dc5a5-1492-465a-a2f4-20e55a069d5b.png)

ericangeles:
@janahintal can you please provide the steps for this test? Thanks.

ericangeles:
@tmcconechy @kevinwhitedesign can we confirm the color that we should use for the New Light? In the example page, we used emerald09 (#1C7F49) but in the Figma file it's #1F9254 (emerald08).

CindyMercadoReyes:
This issue is now resolved.

tmcconechy:
I will reverse some of these changes for now due to https://github.com/infor-design/enterprise/issues/6114
The reason being the size and font changes look abrupt and out of place and want to avoid complaints and breaking.

The biggest problem being the size change on the fields that will look out of place on forms with different fields
![Screen Shot 2022-02-10 at 2 38 29 PM](https://user-images.githubusercontent.com/814283/153483829-4b1d33d0-339e-4f11-84fc-a8b395e02c17.png)

And the tag size changes in general can cause disruption. So will kick this back for now

Approved with @inforandy

kevinwhitedesign:
I believe reverting from this change will end in failing accessibility latest WCAG guidelines. Correct me if I'm wrong @elizabethhartley

tmcconechy:
So according to a scan on https://4580-enterprise.demo.design.infor.com/components/tag/example-index.html only one thing fails...

New Light:
[...]

PR #7296, opened 2023-03-07
7234 - Added example page for widget count with color background

### LINKED ISSUE #7234: Widget Counts: Make an example showing it in a colored banner
**Describe the Feature**

We want to ensure that the newer counts work in a banner widget. So take the example
and implement a copy of this page https://main-enterprise.demo.design.infor.com/components/personalize/test-classes-short.html?colors=BB5500 but replace the counts with these ones https://design.infor.com/code/ids-enterprise/latest/demo/components/counts/example-widget-count.html?colors=BB5500

We may need some design feedback on color schemes.

**Expected behavior**
A new example where you can change colors that looks good with widget counts.

**Additional context**
Email / HomePages

Example should look similar to this one. But make sure the counts are centered and the text is centered and everything looks as good as ours  https://design.infor.com/code/ids-enterprise/latest/demo/components/counts/example-widget-count.html?colors=BB5500

image

yohannahbautista:

Added filled class that can be added in instance-count: Sets instance-count icon background color to inherit so it can get the background color of parent class. Also included setting font color to lighter colors so that text can be readable.

Closes #7234

- Go to: http://localhost:4000/components/counts/example-header.html
- Change theme color
- Alert icon background should inherit color

### DIFF
--- a/app/views/components/counts/example-header.html
+++ b/app/views/components/counts/example-header.html
@@ -1,0 +1,84 @@
+<div class="is-personalizable">
+  <div class="card" style="height: 155px; min-height: 155px; border: none; border-radius: 0;">
+    <div class="card-content l-center personalize-header" style="height: 155px; min-height: 155px;">
+      <div class="instance-count">
+        <span class="count ruby06">5352</span>
+        <svg class="icon icon-alert ruby06" focusable="false" aria-hidden="true" role="presentation">
+          <use href="#icon-alert"></use>
+        </svg>
+        <span class="title">High</span>
+      </div>
+
+      <div class="instance-count">
+        <span class="count amber06">10021</span>
+        <svg class="icon icon-alert" focusable="false" aria-hidden="true" role="presentation">
+          <use href="#icon-alert"></use>
+        </svg>
+        <span class="title">Medium</span>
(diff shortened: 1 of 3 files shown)


tjamesallen15 (changes requested):
image

http://localhost:4000/components/counts/example-header.html?theme=new&mode=dark&colors=default
On Dark Mode, The right most circle can't be seen.

ericangeles (approved):

tmcconechy:
Was just thinking about @tjamesallen15 comment. Maybe we can make the standard ones change the color to the standard color per theme. I.E. Slate on normal, white on dark mode, darker slate on contrast.

Some of the other alert colors do not work well. Maybe they do need to change colors in themes? Like https://main-enterprise.demo.design.infor.com/components/notification/example-index.html

tmcconechy:
Yeah. I think it needs more thought. Especially around https://github.com/infor-design/enterprise/issues/5673 ~ there is a few issues where we want to define alert colors across themes.

So i think we could solve it when this is defined. But it is true it cant be seen in dark mode. So thinking for non contrast ones we could make it work with a light color for the "standard" one.

Issue #5913, opened 2021-12-01
Datagrid row status icon not correct when first column is a dropdown.


dmcodrum:
**Describe the bug**

In a datagrid component, if the first column is a dropdown, the status indicator/icon doesn't show correctly. See attached screenshot.

**To Reproduce**

1. In the IDS-NG app, modify the file '..\src\app\datagrid\datagrid-demo.service.ts' per the attached file 'datagrid-demo.service.txt' It's essentially modifying the demo to show a dropdown in column 1.
2. Run the IDS-NG demo, select 'Datagrid - Dynamic'.
3. In the demo app, click the Add Row button.
4. Note the appearance of the status icon, it's not correct.

**Expected behavior**
The status icon should appear correctly.

**Version**
- ids-enterprise: 4.57

**Screenshots**

![Capture](https://user-images.githubusercontent.com/39279254/144294532-2e571552-eaf6-4541-b14a-ffc02ba1fc9f.JPG)

**Platform**
 - Infor Application/Team Name: GLT
 - Device: PC
 - OS Version: Windows 10 
 - Browser Name: chrome
 - Browser Version: 96.0.4664.45 

**Additional context**
Demo source file:

[datagrid-demo.service.txt](https://github.com/infor-design/enterprise/files/7636453/datagrid-demo.service.txt)

glenlieorillo:
Tested and passed.

https://main-enterprise.demo.design.infor.com/components/datagrid/test-selected-rows-addnew.html

https://user-images.githubusercontent.com/102151352/228188925-376089ea-7fdd-4f7a-b291-14926f8ab84e.mov


https://user-images.githubusercontent.com/102151352/228189604-117e1b4b-9905-4f7e-9f88-97d2ac3b7819.mov

PR #7341, opened 2023-03-21
NG5913 - Fix on styling in row status icon when first column is not a select column

yohannahbautista:

Updated styling on datagrid icon. 

Closes https://github.com/infor-design/enterprise/issues/5913

- Go to: http://localhost:4000/components/datagrid/test-selected-rows-addnew.html
- Add row
- Row status icon should be shown right


tmcconechy (changes requested):
@yohannahbautista can you test it in the 4 row heights. It seems like only normal is rendering nicely for the smaller row heights maybe we just keep it teh same size no matter what row height is on.

We may redesign this in the future

tmcconechy (approved):
Looks good now. Thanks!

ericangeles (approved):

PR #7348, opened 2023-03-22
7094 - Listview Search in Card

tjamesallen15:

### DIFF
--- a/app/views/components/listview/test-modal.html
+++ b/app/views/components/listview/test-modal.html
@@ -8,7 +8,8 @@
   <div class="twelve columns">
 
     <button class="btn-secondary" type="button" id="list-tasks">List of Tasks</button><br><br>
-    <button class="btn-secondary" type="button" id="list-companies">List of Companies</button>
+    <button class="btn-secondary" type="button" id="list-companies">List of Companies</button><br><br>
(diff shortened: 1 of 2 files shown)


yohannahbautista (approved):

ericangeles (changes requested):
The only issue that I am seeing is on mobile view. See the width of listview searchfield.

You'll see it around 410px and below width. You can also use the mobile devices (I used iPhone 12 Pro)

Screen Shot 2023-03-22 at 8 04 06 PM

ericangeles (approved):

tmcconechy (changes requested):
@yohannahbautista the problem i see is that when i type "ab" one by one the modal gets smaller and smaller and changes sizes each key. Maybe we need to set a maintain constant width for this case? Maybe by checking the search field width?

tjamesallen15:
> @yohannahbautista the problem i see is that when i type "ab" one by one the modal gets smaller and smaller and changes sizes each key. Maybe we need to set a maintain constant width for this case? Maybe by checking the search field width?

Will check on this one

EdwardCoyle (approved):
Not seeing any of aforementioned issues here, looks great @tjamesallen15

Issue #6929, opened 2022-10-27
Datagrid: Undefined tooltip when headerIconTooltip setting is not defined


s-werking:
**Describe the bug**
When the column setting headerIconTooltip is not included on a column with a header icon and the icon is hovered, the tooltip is shown as undefined.

If I want a header icon to not have a tooltip I need to include the column's headerIconTooltip setting and set it to an empty string. Instead, could it be possible to only use the headerIconTooltip setting if a tooltip is needed?

**To Reproduce**
1. Remove the headerIconTooltip setting from the 'productName' column in http://localhost:4000/components/datagrid/example-header-icon-with-tooltip.html
2. Go to http://localhost:4000/components/datagrid/example-header-icon-with-tooltip.html
3. Hover over the icon on the Product Name column header
4. See that the tooltip is shown as undefined

**Expected behavior**
If the headerIconTooltip setting is not included on a column with a header icon, upon hovering over the icon there should not be a tooltip.

**Version**
- ids-enterprise: v4.68.0-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/76002134/198306723-f21e4556-f34f-4cbb-b229-e4ea7dd1130a.png)

**Platform**
 - Infor Application/Team Name: lawson
 - OS Version: Windows 11
 - Browser Name: chrome
 - Browser Version: 107.0.5304.63 (Official Build) (64-bit)

jbrcna:
QA Passed.
<img width="1389" alt="image" src="https://user-images.githubusercontent.com/56722184/228431636-74d07281-c0b7-4889-872d-7df4ae60c880.png">

PR #7351, opened 2023-03-22
6929 - datagrid undefined headicon

tjamesallen15:

This pull request will fix header icon tooltip showing when undefined.

Closes https://github.com/infor-design/enterprise/issues/6929

- Go to http://localhost:4000/components/datagrid/test-empty-header-icon.html
- Header Icon tooltip should not be shown

### DIFF
--- a/app/views/components/datagrid/example-header-icon.html
+++ b/app/views/components/datagrid/example-header-icon.html
@@ -1,0 +1,80 @@
+<div class="row" style="margin-top: 40px;">
+  <div class="twelve columns">
+
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<style id="special-test-rules">
+  .tooltip-extra-class {
+    background-color: #488421;
+    /* Emerald09 */
+    border: 1px solid #488421;
+  }
+
+  .tooltip-extra-class.top .arrow::after {
+    border-top-color: #488421;
+    /* Emerald09 */
+  }
+
+  .tooltip-extra-class.bottom .arrow::after {
+    border-bottom-color: #488421;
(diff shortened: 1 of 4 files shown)


s-werking:
Hi @tjamesallen15, I was hoping that we could have a header icon shown without setting a headerIconTooltip option, but the header icon is not shown in the sample. Could this be possible?

s-werking:
> @tjamesallen15 i think reading the issue its a bit confusing but you want to only show the icon in the header and no tooltip? Is that right @s-werking ?

Correct

tmcconechy (changes requested):
See inline comment and suggestion. Think the example needs a tweak

EdwardCoyle:
@tjamesallen15 I think there may be something else triggering a tooltip somewhere:
- Goto http://localhost:4000/components/datagrid/example-header-icon-with-tooltip.html
- Hover just the empty space inside the "Product Name" cell.  See screenshot 1
- Then hover the icon inside that same cell.  See screenshot 2

| Screen Shot 2023-03-23 at 11 28 32 AM | Screen Shot 2023-03-23 at 11 28 22 AM |

This is maybe related to Tim's comment about the settings, maybe a conflict?

tmcconechy:
yeah thats right @EdwardCoyle i think the `headerTooltip` in the example can go. For purposes of fully testing the use case.

tjamesallen15:
Thanks for both of you guys @tmcconechy  @EdwardCoyle

tmcconechy (approved):
Added a new example with no tooltip as requested. So we have two examples. Looks good now! Thanks!

PR #7357, opened 2023-03-24
7234 - Updated example page and styles

### LINKED ISSUE #7234: Widget Counts: Make an example showing it in a colored banner

yohannahbautista:

Updated example page and styles for instance count.

Closes #7234 

- Go to: http://localhost:4000/components/counts/example-hero.html
- Alert icons should have matching background with hero widget background


tmcconechy (changes requested):
Looks good but maybe we should make the hero have a dynamic height. Lets do it with 32px padding on top and bottom and center the counts. Probably need to remove it from the card since the card has a fixed height.

Basically the hero is too big for the content and the content is not centered, something like:

Screenshot 2023-03-24 at 9 44 56 AM

Also watch if the scrollbar is in the wrong place after the change..

Screenshot 2023-03-24 at 9 42 47 AM

app/views/components/counts/example-hero.html
```diff
@@ -1,8 +1,15 @@
...
+    max-height: none !important;
+    height: fit-content !important;
+  }
+</style>
 <div id="maincontent" class="page-container scrollable" role="main">
   <div class="hero-widget is-personalizable">
     <div class="hero-top">
       <div class="hero-content">
-        <div class="card count-container">
+        <div class="card count-container auto-height">
```
EdwardCoyle, line 12: Maybe move the styles from the `<style>` tag here, and move them to an entry under `.card.auto-height` somewhere in here: https://github.com/infor-design/enterprise/blob/main/src/components/homepage/_homepage.scss#L318

app/views/components/counts/example-hero.html
```diff
@@ -1,8 +1,15 @@
+<style>
+  .hero-top, .hero-widget {
+    min-height: 0 !important;
+    max-height: none !important;
+    height: fit-content !important;
```
EdwardCoyle, line 5: We should try to avoid using `!important` as much as possible, and rely on stronger CSS selectors for override behavior:

```scss
.hero-widget {
  .card.auto-height {
    height: fit-content;
    min-height: 0;
    max-height: none;
  }
}
```

tmcconechy:
@yohannahbautista to be clearer. Lets make sure:

- https://main-enterprise.demo.design.infor.com/components/homepage/example-hero-widget.html still works
- but  for http://localhost:4000/components/counts/example-hero.html it should just be the height of the content. so maybe add a class to control "autoheight".
- the hero should never need a scroll bar In it the scrolling is at the layer above it

tmcconechy (dismissed):
Approving for now but noting that on http://localhost:4000/components/counts/example-hero.html?colors=ffffff the colors are not showing the status.

We might want to address that or need design feedback...

tmcconechy (dismissed):

tmcconechy:
Screenshot 2023-03-31 at 9 40 34 AM

this looks pretty good but i think we need design here (not your fault). also the circles are not 100% centered.

@talkaboutdesign what do you think about this pattern?

talkaboutdesign:
This pattern need a lot of work in the future. The status indicators are def problematic. Nothing we need to worry about immediately. We can revisit in the future. 

Few notes on banned area: 
- It's always a fixed size 
- Content is always in a widget

CleanShot 2023-03-31 at 12 18 24@2x
CleanShot 2023-03-31 at 12 18 16@2x

tmcconechy:
ok @yohannahbautista based on design feedback

1) leave the colors as is
2) for this example lets keep the size fixed and put those counts in a border less card and put it in the center of the hero widget like:

https://user-images.githubusercontent.com/2926534/229175240-691b2d53-3d88-4375-a693-49dac60cca3d.png

So it looks like the actual use case. So centered in a borderless widget with a title like "Counts" and then somewhat resembling that. 

We can come back on the colors. My mistake on the dynamic size part. Sorry and thanks for the clarification @talkaboutdesign

talkaboutdesign:
@tmcconechy maybe would be helpful to get you guys an environment from Portal 2 where you can see the banner area in action so you can match it. What do you think?

tmcconechy:
Not a bad idea although we wont implement everything. Maybe for now just tell us if our size differs here https://main-enterprise.demo.design.infor.com/components/homepage/example-hero-widget.html

Its probably separate from this exact issue

talkaboutdesign:
Banner Height: 390px (they set this as a min-height) but it's never more than that because it can only have a widget)
Margin bottom: 20px (between banner and widgets on page)
Inside banner area you have a widget thats vertically center aligned with a height of 360px (borderless)
(data vis is inside this widget)

@tmcconechy Please note that some of these values we want updated to reflect our widget card enhancements. What I am giving you here is what is currently in Ming.le Homepages. 

CleanShot 2023-03-31 at 14 30 32@2x

tmcconechy:
Ok, could you make a new issue @talkaboutdesign with the "new" ones you want. I think its a seperate task.

talkaboutdesign:
Added this to the card enhancement ticket here: https://github.com/infor-design/enterprise/issues/7379

tmcconechy (approved):

tmcconechy:
@EdwardCoyle does this look ok for you now?

tmcconechy (dismissed):

jbrcna (changes requested):
this is failing in Firefox. also button and caret hover colors are off and the page blinks when you switch/open with classic view using safari browser

ericangeles:
@jbrcna can you screenshot the issue? That would be helpful. Thanks!

tmcconechy:
@yohannahbautista @ericangeles if you just try http://localhost:4000/components/counts/example-hero.html in firefox at different sizes you can see

1. its not centered
2. scroll bars or its bleeding over the circle
3. also if you hover the view button i see what @jbrcna is saying

Alternative is we could show maybe 4 counts for the example as a lot of this is related to it not fitting.
Then fix the centering and the hover.

tmcconechy (dismissed):
Ok looks good enough for me for now. Thanks @yohannahbautista 
cc @jbrcna what do you think?

jbrcna:
widget is now aligned in Firefox. However, the button color is still off
`Light & High Contrast`  using Default color
image

`Dark`  using Alabaster color
title 'Counts Hero Widget Example' is also not readable
image

tmcconechy (approved):
@jbrcna can you have another look. This doesnt have to be perfect as we will come back to it after design feedback (its an odd pattern)

jbrcna (dismissed):
[example-hero.html?theme=new&mode=contrast&colors=ffffff](http://localhost:4000/components/counts/example-hero.html?theme=new&mode=contrast&colors=ffffff) - paging buttons is not visible
image


[example-hero.html?colors=ffffff&theme=classic&mode=dark](http://localhost:4000/components/counts/example-hero.html?colors=ffffff&theme=classic&mode=dark)
image

tmcconechy (approved):

tmcconechy (approved):
looks good to me. what do you think @jbrcna ?

Issue #6921, opened 2022-10-26
DataGrid: Cannot see more than 3 buttons in a grid toolbar


pooBadiger:
**Describe the bug**
In a data grid, not able to see more than 3 buttons in a grid toolbar. 

- Add more than 3 buttons inside a soho-toolbar.

 `<soho-toolbar class="contextual-toolbar is-hidden">
            <soho-toolbar-title>
                <div soho-selection-count>{{ "REGISTRY.ZERO_SELECTED" | transloco }}</div>
            </soho-toolbar-title>
            <soho-toolbar-button-set>
                 <button class="btn">
                    <svg class="icon">
                        <svg soho-icon icon="edit"></svg>
                    </svg>
                </button>..............
                 ....<add more than 3 buttons>
</soho-toolbar-button-set>`

**Expected behavior**
- Grid toolbar should show all the added buttons.

**Version**
"ids-enterprise-ng": "13.4.0",

**Screenshots**
![image](https://user-images.githubusercontent.com/93663725/197978748-5589aec8-becd-4d43-a287-b953a32aef5d.png)

**Platform**
 - Infor Application/Team Name: Infor Process intelligence/ Infor PI
 - Device:  Dell laptop
 - OS Version: Windows 11 Enterprise
 - Browser Name: chrome
 - Browser Version: Version 106.0.5249.119 (Official Build) (64-bit)

tjamesallen15:
Hello @pooBadiger, I'm trying to reproduce your issue but to no avail. Can you check my screenshots below.

This was screenshot from IDS Main
![image](https://user-images.githubusercontent.com/34049950/209752497-13d76244-bf95-455b-bcb3-73c4183a2d2e.png)

This was screenshot from IDS Angular
![image](https://user-images.githubusercontent.com/34049950/209752553-890bc049-48df-4d1f-a5c3-0436be4b69de.png)

How were you able to reproduce the issue? 

cc @tmcconechy @ericangeles

tmcconechy:
I was at the time @tjamesallen15 but maybe we just needed `maxVisibleButtons` set? Can you try to see if it works with a toolbar-flex. I was thinking the problem is a toolbar-flex doesn't work here in contextual toolbar.

So if we make this about sorting toolbar-flex that would be best as the old toolbar has limitations....

@pooBadiger maybe you can try `maxVisibleButtons` as an example

pooBadiger:
**maxVisibleButtons** helped! That works! thanks!

[...]

PR #7364, opened 2023-03-27
6921 - Datagrid Toolbar More Buttons

tjamesallen15:

This pull request will fix datagrid toolbar to be able to show buttons more than two.

Closes https://github.com/infor-design/enterprise/issues/6921

- Go to http://localhost:4000/components/datagrid/test-datagrid-toolbar-visible-buttons.html
- Select a row to show the buttons in toolbar


EdwardCoyle (changes requested):
This is a great start, but I think its missing some things:
- It should be possible for the Toolbar to spill over buttons that don't fit into a "More Actions" menu like the standard one (see [screenshot from the issue](https://user-images.githubusercontent.com/34049950/209752553-890bc049-48df-4d1f-a5c3-0436be4b69de.png)).  The new example is displaying all the buttons as expected, but the collapsed view isn't displaying this menu and the remaining buttons appear out of alignment as a result.  I think it's probably fine for flex toolbar to 
- The tertiary buttons in the contextual toolbar style seem to be missing some states -- see below:

tmcconechy:
Also noticed if you put a `disabled` attribute on one of the buttons. It does not appear disabled.

yohannahbautista (approved):

tmcconechy (commented):
This works pretty good. But one thing that doesnt look great is the first time you open the page and select a row you see the buttons and then afterwards they shift into place.

Anyway to fix that? Perhaps the code added to datagrid.js should be part of the tooltip code? And i would think that if the example uses flex box then it shouldnt shift. And we only need to fix this for flex toolbar not old toolbar.

tjamesallen15:
> This works pretty good. But one thing that doesnt look great is the first time you open the page and select a row you see the buttons and then afterwards they shift into place.
> 
> Anyway to fix that? Perhaps the code added to datagrid.js should be part of the tooltip code? And i would think that if the example uses flex box then it shouldnt shift. And we only need to fix this for flex toolbar not old toolbar.

Hello @tmcconechy , How about this latest push?

tmcconechy (commented):
I think the styling can be slightly improved

- buttons are not centered 100%
- the actions ... button should be closer to the right side (16px on each side (was 20)
Screenshot 2023-04-03 at 9 58 24 AM

tmcconechy (approved):

yohannahbautista (approved):

ericangeles (approved):

Issue #7358, opened 2023-03-24
Bug "field filter in modal design-glitch #7001" still not fixed


rusteninfor:
**Describe the bug**

See https://github.com/infor-design/enterprise/issues/7001 for details. This issue isnt 100% fixed.
The QA page does not display the error, unless you add the classes : modal is-visible to the `<div class="row top-padding">`. 

So instead of
`<div class="row top-padding">`
it should be 
`<div class="row modal is-visible top-padding"> `
to provoke the error

**To Reproduce**
See above

**Additional context**
#7001

tmcconechy:
@tjamesallen15 can you take another look at this and #7001?

tjamesallen15:
> @tjamesallen15 can you take another look at this and #7001?

sure sure

jbrcna:
QA Passed
https://main-enterprise.demo.design.infor.com/components/modal/test-modal-field-filter.html
<img width="493" alt="image" src="https://user-images.githubusercontent.com/56722184/229463120-dc8ce17a-9245-491b-a86c-25a2281ad208.png">

PR #7365, opened 2023-03-27
7358 - field filter alignment

tjamesallen15:

This pull request will fix a bug in field filter where the design is not properly aligned on Modal.

Closes https://github.com/infor-design/enterprise/issues/7358

- Go to http://localhost:4000/components/modal/test-modal-field-filter.html
- Open `Modal`
- Check the alignment of the input boxes.

### DIFF
--- a/app/views/components/modal/test-modal-field-filter.html
+++ b/app/views/components/modal/test-modal-field-filter.html
@@ -1,0 +1,282 @@
+<div class="row top-padding">
+	<div class="twelve columns">
+
+		<button class="btn-secondary" type="button" id="add-context">Show Modal</button>
+
+		<!-- Modal Example -->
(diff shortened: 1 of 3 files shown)


tmcconechy (changes requested):
@tjamesallen15 maybe we should just make an example with this in a modal for testing instead?
Also if i follow the steps i notice i cant click on  with the field filters with this classes present.

tjamesallen15:
> @tjamesallen15 maybe we should just make an example with this in a modal for testing instead? Also if i follow the steps i notice i cant click on with the field filters with this classes present.

Yeah. That's what I saw maybe because of the overlay when you forced add the classes for appropriately for modal. I made an example with a modal just what you asked and it's working fine now.

ericangeles (approved):

PR #7526, opened 2023-05-22
NG#1463 - add defaultTickCount setting to automatically add ticks when there are no dataset values

ericangeles:

This PR introduces a new setting called `defaultTickCount` that automatically adds ticks when there are no dataset values. 

Additionally, it includes a fix for the issue where the tick text was not being displayed when adjusting the bar height to accommodate the legend inside the card.

Closes https://github.com/infor-design/enterprise-ng/issues/1463

- Go to http://localhost:4000/components/bar/example-with-zero-value or https://1463-bar-chart-zero-bar-enterprise.demo.design.infor.com/components/bar/example-with-zero-value
- It should show `5` ticks under x axis
- It should also show the tick text below of it
- Test also this http://localhost:4000/components/bar/example-index
- The tick text should be visible, and the legend has been positioned inside the card

### DIFF
--- a/app/views/components/bar/example-with-zero-value.html
+++ b/app/views/components/bar/example-with-zero-value.html
@@ -26,6 +26,6 @@ $('body').on('initialized', function () {
       name: '',
     }];
 
-  $('#bar-example').chart({type: 'bar', dataset: dataset, animate: false});
+  $('#bar-example').chart({type: 'bar', dataset: dataset, animate: false });
 });
 </script>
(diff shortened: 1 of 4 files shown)


jbrcna (approved):
there is a line below the title  in classic view
image

tmcconechy:
@EdwardCoyle @jbrcna yes i think its ok. Looks wierd in this case but the classic widget has a border and we shouldnt remove it.

jbrcna:
thank you for clarifying

PR #7558, opened 2023-05-30
7274 - Added setting to insert and replace next character if maximum length is already reached

### LINKED ISSUE #7274: Mask: Masked Input Behavior without data loss
The masked input behavior pushes the input off the end instead of stopping after the max length is reached.

Masking an input would work better if there was an option to remove overwriting once the max length is reached.

Say we have number: #,###.## as the pattern and the field has 1,234.56. If the user chooses to type 7 in the middle of 3 and 4, the input should stay in-tact (1,234.56) and not allow any other character to be typed in since the field is already filled up. The only possible way for a user to add a number is when they choose to delete a number existing in the field (If we were to delete 3, the expected value should be 124.56 or when deleting 5, the value should be 1,234.6).

This is especially useful when a user does not want to deal with data loss and wants to modify texts with a given mask. We hope that this improvement would also cater to negative values with pattern: -#,###.##

1. We have tried limiting input to its max length - If we restrict the input field with maxlength attribute, this leaves no room anymore for the allowNegative and allowDecimal options on the mask.
2. We have tried keepCharacterPositions - One suggested way was to keep characters in-tact using keepCharacterPositions. We have recreated this on [stackblitz](https://stackblitz.com/edit/ids-quick-start-1501-bbo9bt?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html) but we don't think its working right. Also, if we were to use keepCharacterPositions, we can not add in between any characters even if the field is not filled in and there is a space holder left if we were to delete text in the middle of a string.

**Proposal**
TJM: Add support for insert key. I.E. when the input key is down on mac or pc the behavior acts like it is selecting the value next in the field and replacing it with the key stroke. cc @EdwardCoyle for feasibility / estimate

yohannahbautista:

Added retainValue setting to retain value if maximum length is already reached

Note: Not sure how to do the insert key proposal with Mac.

Closes #7274 

- Go to: http://localhost:4000/components/input/example-index
- On Number 
   - Type an input until it reaches maximum length
   - Type somewhere in between
   - Should behave as usual (pushing the input to end)
   - Hold Insert key and type
   - Should replace next character instead
- On Number - Insert Value
   - Same as Number but any new input after it's reached max length will replace next character instead of pushing the value


tmcconechy:
@kyfloresca is this the desired behavior?
@yohannahbautista what i was thinking is 

a) if insert key is down it works like the insert key should https://www.computerhope.com/jargon/i/insertke.htm#:~:text=The%20Insert%20key%20toggles%20how,Insert%20key%20on%20Apple%20keyboards%3F
b) then the behavior can be permanently turned on maybe `useInsert` as a setting.

I dont think this is right as now you just cant type

kyfloresca:
agreeing on @tmcconechy's thoughts. the behavior should allow insert key override.

glenlieorillo (dismissed):
QA Approved.


https://github.com/infor-design/enterprise/assets/102151352/17ca563b-d268-4471-871f-fe542f58da32

tmcconechy (changes requested):
Marking as request changes. I think it needs to work via the insert key.

jbrcna (dismissed):

janahintal (dismissed):

tmcconechy (dismissed):
This looks right to me. Nice Job. Please test it @kyfloresca ?

deep7102 (dismissed):
LGTM

ericangeles (dismissed):

ericangeles:
@infor-design/qa FYI

jbrcna (approved):

Issue #7386, opened 2022-09-08
IdsAppMenu: Module Switcher new Component


tmcconechy:
**Is your feature request related to a problem or use case? Please describe.**
Add a new side bar view to the app menu. This will encourage teams to simplify the navigation experience

https://www.figma.com/proto/bL4EBYtWjgfBsvTVI9seXU/branch/eHul2nKNMdypFLSb55ZwR1/Application-Menu?page-id=3185%3A78&node-id=3185-6970&viewport=-726%2C826%2C0.5&scaling=scale-down&starting-point-node-id=3185%3A6954

Status: Currently does need some final design clarification. New name "Module Switcher".

**Describe the solution you'd like**

- [ ] Add an option to show the side bar view
- [ ] When hamburger is clicked it animates to the "old view"
- [ ] Options to only show the new view / only show the old view
- [ ] Ability to codify which icons appear (only top level will), so possibly just a slot with icon buttons

**Describe alternatives you've considered**

**Additional context**

- [ ] Masthead should be one on the right of the two (maybe)
- [ ] Logo should line up with masthead

tmcconechy:
Design Should be more ready Monday - Olga is the contact person. We might skip the module switcher / and collapse and expand button for now (may be top/conversation on going).

ozelenska:
Hi @EdwardCoyle! Exactly as @tmcconechy said, we will be ready provide handoff file on Monday. We still work on Module Switcher and Collapse / Expand, but it shouldn't affect other functionality, so you can start your work. If any questions - feel free to ping me. I am rare github user, so pinging me on Teams might be faster :) Let's collaborate on it closely 🤝

EdwardCoyle:
Working on this in branch [`7386-add-nav-menu`](https://github.com/infor-design/enterprise/compare/7386-add-nav-menu)

PR #7601, opened 2023-06-14
NG1477 - Improve some Enterprise components for better compatibility with Angular Module Nav

EdwardCoyle:
This PR fixes a couple issues in IDS Components related to behavior in the Angular Module Nav component wrapper:
- Fixes a bug where the `filterable` setting was always on
- Adds a new demo page with a disabled `filterable` setting
- Fixes a dropdown bug that was causing Module Nav Role Switcher to render its Dropdown wrapper element twice
- Fixes some styling bugs and consistency issues in demo examples

Related to #7386
Related to infor-design/enterprise-ng#1477

- Open https://ng1477-module-nav-patch-enterprise.demo.design.infor.com/components/module-nav/example-index.html
- Everything should behave as expected (the example has been updated to enable filtering explicitly).  Specifically check filtering, dropdown behavior, and settings menu behavior
- Open https://ng1477-module-nav-patch-enterprise.demo.design.infor.com/components/module-nav/example-not-filterable.html
- See that using the Searchfield on this example does nothing (this is for custom search behaviors)

** EXTRA NOTE:** @tmcconechy this will need patching back into the `v4.84.x` branch, and a patch release is needed.

### DIFF
--- a/app/views/components/module-nav/example-hidden.html
+++ b/app/views/components/module-nav/example-hidden.html
@@ -1,9 +1,3 @@
-<style>
-  .test-spacer {
-    height: 2000px;
-  }
-</style>
-
 <section class="module-nav-container">
   <aside id="nav" class="module-nav">
     <div class="module-nav-bar">
--- a/app/views/components/module-nav/example-not-filterable.html
+++ b/app/views/components/module-nav/example-not-filterable.html
@@ -1,11 +1,5 @@
-<style>
-  .test-spacer {
-    height: 2000px;
-  }
-</style>
-
 <section class="module-nav-container mode-collapsed">
-  <aside id="nav" class="module-nav" data-options="{ 'filterable': true }">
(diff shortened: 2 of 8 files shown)


EdwardCoyle:
I think that's happening by default when the accordion is re-rendered (it's supposed to do that on initial state).  Should we prevent that in this case?

tmcconechy:
@EdwardCoyle i wasn't 100% sure at first i thought it was remembering... wonder tho if it should just be at the same state (if its tough its not a big deal)

EdwardCoyle:
Maybe just track it for now in a follow up issue and see what the behavior should be

tmcconechy:
ok lets see how it goes. We can raise it if it comes up

tmcconechy (approved):

ericangeles (approved):

PR #7610, opened 2023-06-16
Card/Widget fixes in NG

ericangeles:

This PR fixes the visibility of the custom action to work with the Angular wrapper. It reduces specificity in the jQuery component to align with Angular's lifecycle. Essentially, the jQuery instance is now executed after the Angular component, ensuring that the condition is met.

N/A

- Go to NG project and get the jquery fix via copying the dist or just change directly in the node_modules -> ids-enterprise -> dist -> js ->sohoxi.js
- Go to http://localhost:4200/ids-enterprise-ng-demo/cards-workspace-widgets
- Button `more` should be visible under `Bordered with Actions` card

### DIFF
--- a/src/components/cards/cards.js
+++ b/src/components/cards/cards.js
@@ -102,7 +102,7 @@ Cards.prototype = {
     const isSingle = this.settings.selectable === 'single';
     const isBordered = this.settings.bordered === true;
     const isBorderLess = this.settings.bordered === false;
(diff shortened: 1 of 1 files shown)


ericangeles:
Workaround for this issue is to simply add the class `show-buttons` to the card/widget. However, personally, this solution doesn't seem logical. It would be better to address this problem in the core code.

EdwardCoyle (approved):

tmcconechy (approved):

PR #7670, opened 2023-07-05
7383 - Updated popupmenu color styles

### LINKED ISSUE #7383: Popupmenu/CAP: Popupmenu button isn't visible when the menu is open and is in a CAP header in classic theme dark mode
**Describe the bug**
**In classic theme dark mode** - when there's a popupmenu in a contextual action panel header, the popupmenu button has the same color as the header when the menu is open, resulting in it not being visible.

**To Reproduce**
1. Go to https://design.infor.com/code/ids-enterprise/latest/demo/components/contextualactionpanel/test-complex-toolbar.html?colors=50535A&theme=classic&mode=dark
2. Click on "Contextual Action Panel" to open the CAP
3. Click on "..." to open the popupmenu
4. Notice that the color of the popupmenu button is changed and no longer visible

**Expected behavior**
The button should be a different color.

**Version**
- ids-enterprise: 4.80.1

**Platform**
 - Infor Application/Team Name: Homepages

**Additional context**
In New theme, the button color is set to `#606066` and looks like this:
image

yohannahbautista:

Add color style for toolbar flex popupmenu button

Closes #7383 

- Go to: http://localhost:4000/components/contextualactionpanel/test-complex-toolbar-flex.html
- Open Contextual Action Panel
- Open Filter popup menu
- Filter button color should be visible when open
- Check other versions and modes

### DIFF
--- a/src/components/toolbar-flex/_toolbar-flex.scss
+++ b/src/components/toolbar-flex/_toolbar-flex.scss
@@ -309,6 +309,14 @@
     [class^='btn'] {
       margin-right: 0;
 
+      &.btn-menu.is-open {
+        color: $button-color-tertiary-initial-font;
+
+        .icon {
+          color: $button-color-tertiary-initial-font;
+        }
+      }
+
       @media (max-width: $breakpoint-slim) {
         width: auto;
       }
--- a/src/components/toolbar-flex/_toolbar-flex.scss
+++ b/src/components/toolbar-flex/_toolbar-flex.scss
@@ -309,14 +309,6 @@
     [class^='btn'] {
       margin-right: 0;
 
-      &.btn-menu.is-open {
-        color: $button-color-tertiary-initial-font;
-
-        .icon {
-          color: $button-color-tertiary-initial-font;
-        }
-      }
-
       @media (max-width: $breakpoint-slim) {
         width: auto;
       }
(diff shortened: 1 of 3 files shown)


src/components/toolbar-flex/_toolbar-flex.scss
```diff
@@ -309,6 +309,14 @@
     [class^='btn'] {
       margin-right: 0;
 
+      &.btn-menu.is-open {
```
EdwardCoyle, line 312: This looks good and works fine, but my only comment is that I think these rules don't belong in the base toolbar code.  There is another set of Toolbar style rules that live in the Contextual Action Panel code that drive these styles, and should probably land there: https://github.com/infor-design/enterprise/blob/main/src/components/contextualactionpanel/_contextualactionpanel.scss#L151

Otherwise this looks good 👍

tmcconechy (approved):

glenlieorillo (changes requested):
Fix already applied. However, filter button is cut-off in mobile view.

image

jbrcna (approved):

tmcconechy:
I think some of the icons may be cut off depending on the size of the screen. Not a ton we can do about that. So will skip that point

PR #7681, opened 2023-07-06
7681-Overflow issue on Android and Tablet Devices for message section [v 4.84.0 ]

harshpareek06:
Changes: I have added the overflow property under the CSS class modal-body-wrapper for the modal component.

Fix: When adding the property overflow: auto for the modal-body-wrapper class under the modal component it basically provides the scrolling functionality which resolves the overflow issue.

Related Jira: https://jira.infor.com/browse/WFMD-3411
Description: When Changing the ids-enterprise-ng version from 15.2.0 --> 15.4.0 I found that 
it breaks the UI for the Message screen in Self-Service-Portal on mobile and Tablet Devices. On further investigation I found the Overflow property was missing.

image-2023-06-15-16-05-09-863


tmcconechy:
@harshpareek06 this wont work to just merge this into 4.84.0. I think what i would do is make an issue as its not clear to me what the fix does as there is too many changes. Then we have to put it through quick grooming and look at it.

i.e. all fixes should have

a) an issue describing the change
b) be merged on main first not a branch

You can mention you want it patched on the issue and we can do that after it goes in main but this one is adding too much for what your describing to 4.84 (looks like if you made it for main instead it might work?)

harshpareek06:
Hey, Tim I have stated the issue in the detail of the pull request, and
talking about the changes I have added only one change which is in file
_modal. scss, and there also I just added an overflow property.

The client is on Angular version 14 and I am not sure if the ids-enterprise
latest version may be compatible with it.

Thanks,
Harsh pareek

Issue #7530, opened 2023-05-22
Datagrid: defaultFilterOperator is not honored on a column filterable with a lookup


claudenbach:
**Describe the bug**
A column whose filterType is 'lookup' does not honor the defaultFilterOperator.  Works correctly on a column whose filterType is 'text'.

**To Reproduce**

1. Using the attached 'test-filter-with-default-operator.html'
2. For Product Id and Product Name columns, the defaultFilterOperator has been set to 'starts-with'
3. Notice that the displayed filter operator for Product Id column is 'contains' 
4. Open the filter operator for Product Id column, notice no item is checked
5. Notice that the displayed filter operator for Product Name column is 'starts-with'
6. Open the filter operator for Product Name column, notice that 'Starts With' is checked

**Expected behavior**
Expect that the defaultFilterOperator should be honored on a column that is filterable by a lookup.

**Version**

- ids-enterprise: v4.83.0-beta.0

**Additional context**
Landmark is currently on v4.80.x

PR #7710, opened 2023-07-13
7530 - Datagrid Default Operator For Lookup

tjamesallen15:

This pull request will fix a bug in datagrid where default operator for lookup is not rendered properly.

Closes https://github.com/infor-design/enterprise/issues/7530

- Go to http://localhost:4000/components/datagrid/test-filter-with-default-operator.html
- See the Default Operator for the `Product Id`

### DIFF
--- a/app/views/components/datagrid/test-filter-with-default-operator.html
+++ b/app/views/components/datagrid/test-filter-with-default-operator.html
@@ -1,0 +1,85 @@
+<div class="row">
+  <div class="twelve columns">
+    <div id="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+      var grid,
+        columns = [],
+        data = [];
+
+      // Define Some Sample Data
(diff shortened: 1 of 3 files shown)


tmcconechy (dismissed):
@claudenbach how does this look for you?

claudenbach:
Seems correct for a lookup is that is an alpha type, but what about a lookup that is a numeric type?  The operators are different for a numeric vs alpha field type.

tmcconechy:
@claudenbach i dont think we have had this before tho. Its always treated as a string. Although we "could" check it..

claudenbach:
We do have examples were the column with a lookup is a numeric.  In these cases, we would be using the numeric operators (or filter conditions). For example a column containing a manufacturer number, product number, or employee number.

tmcconechy:
and that already works ? ok @tjamesallen15 do you know what to do to fix?

claudenbach:
Yes, that currently works.

claudenbach:
Here's an example...

Git-7530-Example

tjamesallen15:
Try checking my latest push

tmcconechy:
@claudenbach how does it looks now?

claudenbach:
Looks good.  Thanks!

PR #7831, opened 2023-08-17
7819 - Use a different selector in homepage

### LINKED ISSUE #7819: Homepage: Unable to use multiple .hero-widget
**Describe the bug**
In OS Portal, we're planning to use the `.hero-widget` class/component for our Banner widgets, which is currently a custom implementation based on the IDS hero. But we want to use `.hero-widget` to get the new, improved colors for the area.

But this code in `homepage.js` is problematic to us

```
 initHeroWidget() {
    let heroWidget = $('.hero-widget');
    if (heroWidget.length > 1) {
      heroWidget = heroWidget.not(':first').remove();  <---
    }
    this.heroWidget = heroWidget;
  },
```` 
The selector seems too generic. A user has multiple workspaces (homepages), each could contain a banner area ("hero-widget"). This means that if loading a second workspace containing a banner area, it will be removed and not visible.

I can see that one homepage component should only have one related hero area, but not that the entire app is only allowed to have one.

**Expected behavior**
More than one `.hero-widget` element must be allowed in DOM. One per homepage.

**Version**
- ids-enterprise: v4.85.0

tmcconechy:

Changes the selector so that if there is more than one homepage component it will work.

Fixes #7819

- go to http://localhost:4000/components/homepage/example-hero-widget.html
- ensure works fine like before

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,7 @@
 - `[Editor]` Fixed fonts selection selection on editor. ([#7762](https://github.com/infor-design/enterprise-ng/issues/7762))
 - `[FieldFilter]` Fixed Dropdown border not rendered properly. ([#7600](https://github.com/infor-design/enterprise/issues/7600))
 - `[FileuploadAdvanced]` Fixed Close Button not rendered properly. ([#7604](https://github.com/infor-design/enterprise/issues/7604))
+- `[Homepage]` Changed selector so multiple hero banners in a page will work. ([#7819](https://github.com/infor-design/enterprise/issues/7819))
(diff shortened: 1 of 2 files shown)


janahintal (dismissed):
Approving. Thanks!

janahintal:
@tmcconechy, noticed that the '...' button isn't that visible in New theme, light mode. 
image

tmcconechy:
@janahintal not seeing that issue ~ if you need more steps lets make a new issue as its unrelated to this

Screenshot 2023-08-18 at 11 43 01 AM

clayinfor (approved):

ericangeles (approved):

PR #7878, opened 2023-08-31
7680 - Added expanded default for expandable formatter

### LINKED ISSUE #7680: Datagrid: Allow to configure expandable rows to be expanded
Datagrid Expander formatter based on the documentation should allow to be exanded by default:
image
This is not implemented right now, but could be done similarly to how it is done in Group formatter.

Passing expanded to the Expander formatter should allow to make expandable rows expanded by default.

**Additional context**
Infor Application/Team Name: Infor WFM

yohannahbautista:

A row will be expanded if there's an `expanded: true` in the dataset item.

Closes #7680 

- Go to: http://localhost:4000/components/datagrid/example-expandable-row.html
- First row should be expanded

### DIFF
--- a/app/views/components/datagrid/example-expandable-row.html
+++ b/app/views/components/datagrid/example-expandable-row.html
@@ -15,7 +15,7 @@
         data = [];
 
       // Some Sample Data
-      data.push({ id: 1, productId: 2142201, sku: 'SKU #9000001-237', productName: 'Compressor', activity:  'Assemble Paint', quantity: 1, price: 210.99, status: 'OK', orderDate: new Date(2014, 12, 8), action: 'Action'});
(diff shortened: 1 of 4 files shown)


ericangeles (approved):

glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/cc806e1b-e7ca-4507-b2aa-03fcdabe8d14

jbrcna (approved):

tmcconechy (changes requested):
This works great. But maybe we can use `rowExpanded` instead of `expanded` thats what i used in the web component version and that could make it easier if folks upgrade.

Also might make it more `unique`

tmcconechy (approved):
Decided this is fine for now since other apis use `expanded` they dont need to be the same

PR #7958, opened 2023-09-19
7843 - Fixed Widget menu is  correctly aligned in RTL

### LINKED ISSUE #7843: Widget menu is not correctly aligned in RTL
**Describe the bug**
In RTL mode the widget menu dots (...) are to far from the widget border in both classic and New IDS. 
Also the Widget header padding should also be swapped in RTL.

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/homepage/example-index.html?theme=new&mode=light&colors=default&locale=he-IL 
2. Hover the first 1x1  widget on the second row.
3. See error
4. Go to https://main-enterprise.demo.design.infor.com/components/homepage/example-index.html?theme=classic&mode=light&colors=default&locale=he-IL 
5. See error on widgets on second line

**Expected behavior**
The left 15 px should perhaps be removed? And the paddings for the widget header should be swapped in RTL.

**Version**

**Screenshots**
image
image

**Platform**
 - Infor Application/Team Name: OS Portal & Homepages
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 116.0.5845.97 (Official Build) (64-bit)

**Additional context**

adminrupn:
Fixes  #7843
  
  
image

### DIFF
--- a/src/components/cards/_cards.scss
+++ b/src/components/cards/_cards.scss
@@ -725,7 +725,7 @@ html[dir='rtl'] {
     }
 
     &:not(.has-toolbar) > .btn-actions {
-      left: 15px;
+      /*left: 15px; */
       right: auto;
     }
 
--- a/src/components/tabs-module/_tabs-module.scss
+++ b/src/components/tabs-module/_tabs-module.scss
@@ -5,8 +5,12 @@
   @include transition(background-color 500ms cubic-bezier(0.17, 0.04, 0.03, 0.94));
 
   background-color: $module-tabs-bg-color;
-  border-bottom: 1px solid $module-tabs-y-border-color;
+ /* border-bottom: 1px solid $module-tabs-y-border-color; 
   border-top: 1px solid $module-tabs-y-border-color;
+  */
+  border-top: 1px solid #000000;
+  border-bottom: 1px solid #000000;
+
   font-size: 0;
   height: 36px;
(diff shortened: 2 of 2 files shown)


src/components/cards/_cards.scss
```diff
@@ -725,7 +725,7 @@ html[dir='rtl'] {
     }
 
     &:not(.has-toolbar) > .btn-actions {
-      left: 15px;
```
tmcconechy, line 728: @adminrupn thanks for chipping in

a) We started to use `inline-inset-start` and `inline-inset-end` instead of left and right i would suggest those corrections rather then using left and right
b) when removing code just remove it rather than comment out

tmcconechy (changes requested):
A couple suggestions @adminrupn and  thanks for chipping in

a) We started to use `inline-inset-start` and `inline-inset-end` instead of left and right i would suggest those corrections rather then using left and right
b) when removing code just remove it rather than comment out
c) see also lint errors
d) can you add an entry to the CHANGELOG.md

tmcconechy:
@adminrupn can you address the issues on this PR? Or we will close and take care of it later.

tmcconechy:
@adminrupn please feel free to rereaise the PR at a later date with the suggestions. Closing for now as no response.

PR #7990, opened 2023-09-26
7739 - Added options for alternate formatting in weekview

### LINKED ISSUE #7739: WeekView: Time Formatting and Display Order

My problem is a use case for a specific requirement, so to match that i need this enhancement.

[Week view] component has by default:
*Time Format Display*: Currently, the default schedule displays time ranges in different formats depending on the duration of the event. For instance, if an event spans from AM to PM, it shows as "2am-5pm." However, for events within a single time period, it displays as "2-5am" or "2-5pm." so, I want to achieve "2am - 5am" or "2pm - 5pm".

*Time-Sorting Order*: Additionally, we would like to change the sorting order of the schedule entries. The default behavior is to display entries in the order of subject first and then time. However, we need the schedule to display entries in the order of time first and then subject to match the design requirements. By order i mean the order of the inner contents

*Time format change*: Need to change the Time display from 12 hours format to 24hours format

This is what default week view looks
image

This is the requirement which needs to be matched
image

I am seeking a flag that allows me to modify the format of time ranges in the schedule. Currently, the format is displayed as "2-5AM," and I would like the flexibility to change it to "2AM-5AM."

Additionally, I have a requirement to change the order of display data in the schedule. The default order is subject first, followed by time. However, I need the schedule to show the time first and then the subject for better alignment with the design.

need an attribulte to make the week view 24/12 hours format.
provide me a solution/enhancement accordingly.

yohannahbautista:

Added options for alternate formatting in weekview

Closes #7739 

- Go to: http://localhost:4000/components/week-view/example-card.html
- Check last card (Alternate Format)
- Should have the enhancements mentioned in the ticket


tmcconechy (changes requested):
@yohannahbautista the only thing i would like to change is the setting name `militaryTime` as  i dont think thats quite accurate. Can we

a) Make sure the default is 7:00 AM - 7:30 AM now as validated by design (instead of existing 7-7:30 AM)
b) Add a timeFormat option thats more flexible like `HH:mm - HH:mm` as an option instead that uses existing time format options
c) maybe changing that will resolve the tests

ericangeles (approved):
Great work, @yohannahbautista! Since we've added the `timePattern` setting, please also include it in NG. Thanks!

tmcconechy (changes requested):
@yohannahbautista the only issue with this is that the locale doesnt work

http://localhost:4000/components/week-view/example-stacked.html?locale=de-DE (should show 24hr) https://main-enterprise.demo.design.infor.com/components/week-view/example-stacked.html?locale=de-DE

I think all you need to do is make timePattern empty. Then use either the time Pattern if set or the locale's time pattern

ericangeles (approved):

glenlieorillo (approved):
This is now working as expected.

**Light**

image

image


**Dark**

image

image



**High Contrast**

image

image

Issue #8058, opened 2023-10-10
ExpandableRow: Component getting empty after cell change in soho-datagrid


SudarshanDivakarla:
**Is your feature request related to a problem or use case? Please describe.**

While working with soho-datagrid where have a requirement of Expandable Row. On the Expanded section, I have a component for displaying data in it. So, when the row is expanded, i can see the data in the expanded section but when i change a cell of the expanded row, the data is disappearing, and I can only see empty expanded section.

I am using **rowTemplateComponent** in grid options and providing a component (**ExpandableAreaComponent**) which has data in it.

On Cell change updating that row and reassigning the dataset to data grid for updating the summary row which is causing an issue with expandable section.

**Describe the solution you'd like**

**Requirement**:

-> User can expand particular row 
-> expandable section should be expanded and the data in the expanded section should be visible.
-> He can make changes on cell1, then cell 2 and cell3 of that row need to be updated with the values from the backend and simultaneously the summary row also needs to update.

**In this feature Request we need below Methods:**
-> Api to update summary row,
-> Expandable section to be expanded after updateRow() and the component which is given in rowTemplateComponent should be triggered and the data in the component should be visible.

**Please find the below stack blitz example for more information:**
https://stackblitz.com/edit/ids-quick-start-1671-q3e7sw?file=src%2Fapp%2Fdatagrid%2Fdatagrid.component.ts

**Steps to Reproduce:**
-> Expand particular row,
-> You can see some labels available in the expanded section.
-> change Cumulative Period Progress cell of the expanded row.
-> on tab the expanded section is expanded but the data in the expanded section is getting disappear, which shouldn't happen.

**Additional context**
<img width="712" alt="image" src="https://github.com/infor-design/enterprise/assets/94103218/11c554fc-5b21-4944-8b97-a5c39ba31034">

And what application do you work on?  **LN Enterprise Collaboration Portal**

ericangeles:
Sure thing, @tmcconechy

PR #8070, opened 2023-10-12
8058 - Datagrid Expandable Row Updates

tjamesallen15:

This pull request will add ability for expandable and summary rows to be updated after cell update.

Closes https://github.com/infor-design/enterprise/issues/8058

- Go to http://localhost:4000/components/datagrid/test-expandable-row-editable-with-summary.html
- Open an expandable row
- Try to edit quantity or price
- See it will update in the expandable row and also in the summary row


tjamesallen15:
Still not yet done with api methods.

tmcconechy (commented):
@tjamesallen15 i know you need to do a bit more here. But noting that

a) when you expand the first row
b) then click update on row zero
c) you will see the - turn to a  + on the expandable row

src/components/datagrid/datagrid.js
```diff
@@ -11494,6 +11530,24 @@ Datagrid.prototype = {
...
+   * @param {object} col The column data
+   * @param {number} cell The cell index
+   */
+  updateSummaryRow(col, cell) {
+    if (!col.summaryRowFormatter) {
+      return;
+    }
+
+    const totals = this.calculateTotals();
+    const rowNodes = this.rowNodes(this.settings.dataset.length);
```
ericangeles, line 11545: @tjamesallen15, maybe we could reduce unnecessary DOM queries, string concatenation, and simplify it.

Off the top of my mind, it might work (I could be wrong, so give it a try and test it).

```js
const totals = this.calculateTotals();
const cellNode = this.rowNodes(this.settings.dataset.length).find('td').eq(cell);

cellNode.html(`<div class="datagrid-cell-wrapper">${totals[col.id]}</div>`);
```

ericangeles (approved):
Looks great! Thanks, @tjamesallen15 🍌

glenlieorillo (approved):
Approving.

https://github.com/infor-design/enterprise/assets/102151352/64cb1bf2-7a13-4e72-becd-37e3f9387a71

However, the focus state is not centered. Will file another issue instead. Thanks.

image

image

janahintal (approved):
image

PR #8261, opened 2023-12-15
8132 - Added contentWidth column setting to set width in formatters

### LINKED ISSUE #8132: Allow buttons to span full width of column in datagrid
Buttons appear clipped on initial load if the column width is lower than the width of the button. For example..
image

An option to allow button to span full width of column. Like this..
full-width-button-example

**Additional context**
This is for LMCLIENT ticket reported by one of our App teams.

yohannahbautista:

Added contentWidth column setting to set width in formatters. Currently only used in button formatter

Closes #8132 

- Go to: http://localhost:4000/components/datagrid/test-contentwidth
- Adjust column width of button column
- Should stay in 100% and adjust accordingly

### DIFF
--- a/app/views/components/datagrid/test-contentwidth.html
+++ b/app/views/components/datagrid/test-contentwidth.html
@@ -21,7 +21,7 @@
       data.push({ id: 7, productId: 2642206, productName: 'Some Compressor', activity:  'inspect and Repair', quantity: 41, price: '100.99', status: 'OK', orderDate: new Date(2014, 6, 9, 12, 12, 12), action: 'On Hold', ordered: 0});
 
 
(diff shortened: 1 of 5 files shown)


ericangeles (changes requested):
@yohannahbautista, there's a strange behavior when resizing the column. Can we also fix it, so we can ensure it's working perfectly?

tmcconechy:
Also seeing the strange resize maybe can fix it by using fixed width columns in initially?

tmcconechy:
@krishollenbeck will this be a problem? Or else we have to reduce the padding

Screenshot 2023-12-18 at 10 07 04 AM

tmcconechy (approved):

ericangeles (approved):
One way to improve this example is by adding the `btn` class to the buttons using a callback function in the formatter.

But I am okay with this fix.

jbrcna (approved):
image

krishollenbeck:
@tmcconechy Hello, I was on vacation for a couple weeks. Just saw your comment on the padding. I think maybe we can add a class or something in our formatter to override the padding if we need to.

Issue #8275, opened 2023-12-19
Tabs: + button looks off when focused


janahintal:
**Describe the bug**
Tabs: + button looks off when focused

**To Reproduce**

1. Go to https://main-enterprise.demo.design.infor.com/components/tabs/example-add-tab-button.html 
2. Click + button 
3. Notice the focus looks off 

**Expected behavior**
Focus should not look off

**Version**

- ids-enterprise: [e.g. v4.9.0 or v4.10.0]

**Screenshots**
![image](https://github.com/infor-design/enterprise/assets/45757016/a42a451f-778b-435b-aafb-716f00367cd0)

**Platform**
 - Infor Application/Team Name: [e.g. Infor XYZ, Infor ABC]
 - Device: (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name: [e.g. chrome, safari, stock browser]
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**

PR #8317, opened 2024-01-08
8275 - Tab Add Tab Button

tjamesallen15:

This pull request will fix tabs add tab button focus.

Closes https://github.com/infor-design/enterprise/issues/8275

- Go to http://localhost:4000/components/tabs/example-add-tab-button.html
- Click the `+` button
- Focus Border should look good


tmcconechy (changes requested):
@tjamesallen15 I see a couple style issues can you check with figma:

https://www.figma.com/file/yaJ8mJrqRRej8oTsd6iT8P/IDS-(SoHo)-Component-Library-v4.8?type=design&node-id=35787-106917&mode=design&t=PcEjm1vR6tYjroyn-0


1) The icon looks like the wrong X? And/Or wrong size?

Screenshot 2024-01-08 at 10 41 08 AM

2) I also see when i hover the tabs the tab shape looks wrong (too much space on top)
Screenshot 2024-01-08 at 10 40 51 AM

tmcconechy:
Looking better. But i think the + looks like it floats a bit much more in space
Screenshot 2024-01-11 at 1 31 12 PM

Can we do it more like above (in figma), also still looks like the wrong icon? 

Also the X looks like to could be moved more to the right and centered vertically

Screenshot 2024-01-11 at 1 31 19 PM

tjamesallen15:
> Looking better. But i think the + looks like it floats a bit much more in space Screenshot 2024-01-11 at 1 31 12 PM
> 
> Can we do it more like above (in figma), also still looks like the wrong icon?
> 
> Also the X looks like to could be moved more to the right and centered vertically
> 
> Screenshot 2024-01-11 at 1 31 19 PM

Will do @tmcconechy , I laugh about your comment in the add button in space. Good giggles. Hahaha.

tjamesallen15:
@tmcconechy   , i'm done with the first one. However the second request you want, it seems like the same with the figma, Giving screenshots below.

Figma
image

Local
image

tmcconechy:
ok @tjamesallen15 looks pretty good now. except only two things

1. the blue line should be connected to the bottom (its a pixel or two up)

Good:
Screenshot 2024-01-16 at 11 46 55 AM
Wrong:
Screenshot 2024-01-16 at 11 47 02 AM

2. Instead of using the + as a character should use the svg icon then it will look the same - i.e. its not using the icon

Screenshot 2024-01-16 at 11 48 26 AM

tmcconechy (approved):

jbrcna (approved):
image

image

image

image

janahintal (approved):
image
image

PR #8377, opened 2024-01-23
copy deploy script

corpulent:
The point of this change is to keep scripts which are executed by the container process baked into the image itself. Or else we run into issues running jobs against published tags with hardcoded values like that api endpoint to publish the docs.

A few TODOs must happen. Maybe @tmcconechy can help.

- We need a package.json file here https://github.com/infor-design/enterprise/tree/main/scripts/publish-docs to install all the dependencies of the publish-docs.js script. We will npm install from it when publishing the docs, and not from the main package.json.
- We should clean up all references to the old deploy-documentation.js script and remove it as well.


tmcconechy (approved):

tmcconechy:
@corpulent any updates here?

PR #8384, opened 2024-01-24
8374 - Add avatar classes for module nav

### LINKED ISSUE #8374: Module Nav Avatar Support
A clear and concise description of what the problem is. [e.g. {Component Name} has a problem when {...}]

This request is for Avatar in module nav to be more dynamic.
image

There are a few issues we are running into with the current implementation. The current implementation assumes it would be used only for a guest process. Hence it's static nature. However, for our purposes, we have a few uses cases when the avatar would be displayed. 

Internally (outside of mingle), external web app (with registration process), and guest candidate.

If a user has created an account or has accessed the web app outside of the context of mingle. The avatar with actor info is displayed.  For example it would look like this...

image

image
image

Currently we have an algorithm in place which will map the first character of each phrase to a color class. Which is mapped to supported colors in IDS themes. We will currently allow up to two characters.

```
 public getUniqueBackgroundColorClass(textValue: string): string {
    /**
     * Color class names should match the class names within
     * _landmark-classic-common.scss
     * _landmark-classic-common.scss
     */
    const colorClasses = [
      'lm-color-amber-bg', 'lm-color-azure-bg', 'lm-color-amethyst-bg',
      'lm-color-turquoise-bg', 'lm-color-emerald-bg', 'lm-color-ruby-bg',
      'lm-color-slate-bg', 'lm-color-graphite-bg', 'lm-color-amethyst-dark-bg',
      'lm-color-ruby-light-bg', 'lm-color-amber-light-bg', 'lm-color-azure-dark-bg',
      'lm-color-emerald-dark-bg'
    ];
```

We then stick the corresponding class on the element depending on the value, which just adds a background color.

`<span class="actor-initials {{actorInitialsColorClass}}">{{actorInitials}}</span>`

A few options I can think of,

- Just CSS for the container, setting height, font, font size, padding, and so forth. And we would populate the HTML with our class and value.
- If the plan is to use SVG then some way to dynamically create SVG avatars. 
  - A  function which could dynamically create the path values to match corresponding value, "L" for example.
  - Have an option to pass in HEX value for color. We would want to match these to be the same as the above classes.

Today we just put our own avatar in there and sort of match the CSS to this example,
https://main-enterprise.demo.design.infor.com/components/module-nav/example-hide-show-sections.html?colors=fff

**Additional context**
Designs:
https://www.figma.com/file/krSBY5vxKygKrnrr1Aqg13/Module-Nav%3A-Icons?type=design&node-id=1-8&mode=design&t=R3vQEWzZ0cuHiDv9-0

https://www.figma.com/file/yaJ8mJrqRRej8oTsd6iT8P/IDS-(SoHo)-Component-Library-v4.8?type=design&node-id=40669-52676&mode=design&t=w6jdEcaIwT7fQhQ8-4

tmcconechy:

Updates the existing image example to show the avatars used in module nave can be done with code with any initial.

Fixes #8374

- go to http://localhost:4000/components/module-nav/example-hide-show-sections.html
- the guest section initials can be changed inline
- go to http://localhost:4000/components/images/example-avatar
- see new options on the top, can set colors and two sizes and 2 or 3 initials (3 is not capitalized)


tmcconechy:
cc @krishollenbeck

tjamesallen15 (approved):

jbrcna (approved):
image
image

image

ericangeles (approved):

glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/3190c762-42b7-4ee6-9535-fc555ea255fc


https://github.com/infor-design/enterprise/assets/102151352/52a24511-1fb7-456d-83c6-92f1f0e94e86

Issue #8372, opened 2024-01-23
Datagrid: Add automation id to the tooltip rendered on mouseover of column's filter operator


claudenbach:
**Is your feature request related to a problem or use case? Please describe.**
Would like an automation id on the datagrid generated div for a mouse over on a column's filter operator.

**Describe the solution you'd like**
Suggest propagating the datagrid column automation id or the list automation id the div created for the mouseover tooltip of the filter operator.

**Describe alternatives you've considered**
Attempting to the created div element by using the list id.

![ListFilterRowTooltip](https://github.com/infor-design/enterprise/assets/39804038/fd1580f1-adf5-4c67-8395-29f294eb6195)

**Additional context**
![image](https://github.com/infor-design/enterprise/assets/39804038/8a8f1a29-d4bb-4100-821d-6682d1ed6729)

PR #8415, opened 2024-02-05
8372 - Add data automation id in grid tooltip

ericangeles:

This PR adds a data automation ID to the grid tooltip, which is based on it's ID. It is added to both the grid tooltip and the tooltip content.

FYI: This applies not only to the filter button but also to all tooltips in the grid.

Closes https://github.com/infor-design/enterprise/issues/8372

- Go to http://localhost:4000/components/datagrid/test-filter-with-tooltip.html
- Hover on button filter
- Open the dev tools and check the tooltip element
- There should be `data-automation-id` in the grid tooltip
- There should also be a `data-automation-id` in the tooltip content

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v4.93.0
 
+## v4.93.0 Features
+
+- `[Datagrid]` Added data automation id of the column's filter operator. ([#8372](https://github.com/infor-design/enterprise/issues/8372))
+
 ## v4.93.0 Fixes
 
 - `[Datagrid]` Fixed cell editable not getting focused on click. ([8408](https://github.com/infor-design/enterprise/issues/8408))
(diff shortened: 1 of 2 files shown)


tmcconechy (approved):

n-ace-ancog (commented):
data-automation-ids for both grid tooltip and tooltip content are displayed.

**Filter tooltip**
image

**Other tooltip in the datagrid**
image

jbrcna (commented):
image

order date doesn't have a tooltip when hovered.
image

ericangeles:
@jbrcna, that's because there is no option checked/selected in the filter options. Once you select/check from the list, you'll see the tooltip.

jbrcna:
> @jbrcna, that's because there is no option checked/selected in the filter options. Once you select/check from the list, you'll see the tooltip.

thanks for clarifying @ericangeles

jbrcna (approved):

n-ace-ancog (approved):

PR #8522, opened 2024-03-07
8211 - Added row height style classes for widgets

### LINKED ISSUE #8211: Homepage control does not allow 3 or 4 rows
**Describe the bug**
We have received a bug ticket for portal that the Homepages control can only add widgets that are two columns high. There are scenarios when you want to use the Related Information viewer widget to view a full document and you need to read the document and even if you have more space in your insight group there is no way to get a widget span 3 rows and 2 columns and then add more widgets next to it. The hight is really important when reading documents.

Note: It has nothing to do with columns. The issue is the limit of the ROWS.
**To Reproduce**
Check the API for the Homepages control, there is no option to set 3 or 4 rows which is what we need. 

**Expected behavior**
I expect that it is possible to add a widget with height 3 rows and then have the algoritm try and fill all the available space with widgets. If there are empy spaces when widgets does not fit, then that is OK, You would have to build the workspace in a way that fits your most common screen size.

**Version**
- ids-enterprise: 4.88.0

**Screenshots**
image

**Platform**
 - Infor Application/Team Name: OS Portal
 - Device: (if applicable) -
 - OS Version: Windows 11
 - Browser Name: All browsers
 - Browser Version:: '

**Additional context**
This is a blocker for one of our customers but we are also dicussing other options, like adding an icon to the smart panel and have the Information viewer in full screen, in which case we don't need this change. But I still wanted to add the ticket for the feedback, as it could be usable to be able to set more rows for others.

yohannahbautista:

Added row height style classes for widgets

Closes #8211 

- Go to: http://localhost:4000/components/homepage/example-height.html
- Check the heights of the widgets

### DIFF
--- a/app/views/components/homepage/example-height.html
+++ b/app/views/components/homepage/example-height.html
@@ -1,0 +1,97 @@
+<div id="maincontent" class="page-container scrollable"  role="main">
+  <div class="homepage" data-columns="5">
+    <div class="content">
+
+      <div class="widget sextuple-height double-width">
+        <div class="widget-header">
+          <h2 tabindex="0" class="widget-title">Widget 1x1 (Sextuple Height) - A</h2>
+        </div>
+      </div>
+
+      <div class="widget quintuple-height">
+        <div class="widget-header">
+          <h2 tabindex="0" class="widget-title">Widget 1x1 (Quintuple Height) - B</h2>
+        </div>
+      </div>
+
+      <div class="widget quad-height">
+        <div class="widget-header">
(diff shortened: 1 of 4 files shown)


jbrcna (approved):
image
image
image
image

glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/4cb8ae4f-39ca-4246-b454-8780cedc5aa6


https://github.com/infor-design/enterprise/assets/102151352/aad6c529-6c80-4149-83bf-5dcebedc6721

PR #7577, opened 2023-06-06
7486 - Fix on incorrect row updates on adding a new row to the next page

yohannahbautista:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

Include page if active page isn't one when updating rows.

**Related github/jira issue (required)**:

Closes #7486 

**Steps necessary to review your pull request (required)**:

- Pull this branch, build and run the app
- Go to: http://localhost:4000/components/datagrid/test-selected-rows-addnew.html
- In More menu, select "Add Row (Bottom)"
- Edit on any editable cell in the new row
- Should not get data from previous rows
- Enter new edits, should update
- Check console and the "Cell change" logs. Row in event and dataRow in active cell should be correct

**Included in this Pull Request**:
- [ ] A note to the change log.

PR #8605, opened 2024-04-05
1672 - Fix errors editing on second page

tmcconechy:

Fixed a bug editing on the second page in server side paging. This was caused by https://github.com/infor-design/enterprise/pull/7577

Fixes https://github.com/infor-design/enterprise-ng/issues/1672

- go to http://localhost:4000/components/datagrid/test-expandable-paging-server-side.html
- go to page two
- edit the editable cells (productName)
- check console and should be no error
- retest https://github.com/infor-design/enterprise/pull/7577 should still work

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ## v4.95.0 Fixes
 
+- `[Datagrid]` Fixed an error editing on non first page in server side paging datagrid. ([#8537](https://github.com/infor-design/enterprise-ng/issues/1672))
 - `[Forms]` Fixed fileupload layout in compact form. ([#8537](https://github.com/infor-design/enterprise/issues/8537))
 
 ## v4.94.0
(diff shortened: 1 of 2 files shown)


n-ace-ancog (approved):
**Updating rows** no errors encountered
@tmcconechy - are we expecting changed values to persist when going to other pages?

Animation

**Previous issue**
Animation

janahintal (changes requested):
Still having console error 
https://www.loom.com/share/78418ff18ce34bc89cc91b3debbf7b51?sid=217a87b0-5633-46c0-b166-67ec43b76bbf

ericangeles:
@janahintal, the error in your video does not seem to be related to our library. Also, I have tested it and I don't see any errors.

tmcconechy:
@n-ace-ancog no the value does not persist in server side examples. (We dont have a place to save it so when the data reloads per page its replaced) but it is saving it as to what we need, its an example issue.

Since this is approved an the error from @janahintal seems to come from some other web plugin. Will merge and patch

Issue #8719, opened 2024-05-14
Datagrid: Incorrectly firing a cellchange event


claudenbach:
**Describe the bug**
In datagrid, open a cell for editing but tab out of the cell without changing the value.  When the value remains the same it's still firing the cellchange event.  Prior to v4.95.x, the method updateCellNode would call this.coerceValue.  In coreceValue, if the oldVal was a number it would perform a Locale.parseNumber.  Without converting the coerceValue to a number (from a string), this statement always sets isCellChange to true;

isCellChange = coercedVal !== oldVal && !fromApiCall;

**To Reproduce**

1. Using the attached example 'http://localhost:4000/components/datagrid/test-cell-changed.html'
2. Click on Product Id cell to enter edit mode
3. Tab out of the cell without changing the value
4. Notice that the row status indicates the row was changed (also have a console log)

**Expected behavior**
The cellchanged event is only fired when the cell value has changed.

**Version**

- ids-enterprise: v4.95.0 (was working in v4.92.7)

**Additional context**
Landmark will need a patch to v4.95

tmcconechy:
@ericangeles if one of you can fix, i'll patch it.. Just push to main..

claudenbach:
Believe this issue was introduced with https://github.com/infor-design/enterprise/pull/8520

PR #8735, opened 2024-05-20
8719 - Datagrid Cell Changed Trigger

tjamesallen15:

This pull request will fix cell changed trigger event on number and string comparison.

Closes https://github.com/infor-design/enterprise/issues/8719

- Go to http://localhost:4000/components/datagrid/test-cell-changed.html
- Click on Product Id cell to enter edit mode
- Tab out of the cell without changing the value

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -10713,6 +10713,11 @@ Datagrid.prototype = {
       this.getActiveDataset()[dataRowIndex];
 
     let oldValue = this.fieldValue(rowData, col.field);
+
+    if (typeof(oldValue) === "number" && typeof(newValue) === "string") {
+      newValue = oldValue == newValue ? oldValue : newValue;
+    }
+
     if (col.beforeCommitCellEdit && (isFileupload || !isCallback)) {
       const vetoCommit = col.beforeCommitCellEdit({
         cell,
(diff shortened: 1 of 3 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -10713,6 +10713,12 @@ Datagrid.prototype = {
       this.getActiveDataset()[dataRowIndex];
 
     let oldValue = this.fieldValue(rowData, col.field);
+
```
ericangeles, line 10716: Could you try this, @tjamesallen15?

```js
if (typeof oldValue === 'number' && typeof newValue === 'string') {
  const newValueAsNumber = Number(newValue);
  newValue = (oldValue === newValueAsNumber) ? oldValue : newValue;
}
```

We need to avoid potential type coercion issues in your code (`==` vs `===`).

In that code also, we need to check to see if `newValue` can be converted into a number before the comparison.

tmcconechy: Also could try just comparing oldValue to newValue? Or does the type change as well and thats the issue?
Also can we add more editors to the other columns so we can test the event in other cases?

tjamesallen15: > Also could try just comparing oldValue to newValue? Or does the type change as well and thats the issue? Also can we add more editors to the other columns so we can test the event in other cases?

the type change and that's the issue. on other editors it doesn't have the problem. only on input editors

ericangeles (approved):
Looks good! Thanks, @tjamesallen15 🍌

n-ace-ancog (approved):
Working as expected:

https://github.com/infor-design/enterprise/assets/156752175/2f655da0-8793-43e0-b239-32f34ead391e

tmcconechy (approved):

PR #8738, opened 2024-05-21
8485. Adjusted width

### LINKED ISSUE #8485: CSS media style rules can be potentially overridden by other "desktop" styles
**Issue Description**
A bug was recently logged in our software where fields laid out in columns contained within an expandable area are displayed without the margin/padding when the window or screen size is small (< 768 pixels wide).  After some initial investigation it was found the @media style that applies the left margin to columns was overridden by the main desktop style.  This should not be the case and the @media style should always take precedence over any other styles (when applicable, due to small screen low resolution etc.).

**Steps to Reproduce**
To reproduce you can take the latest version of the ng-enterprise project and override (expandablearea.demo.html)
with the contents of [expandablearea.demo.txt](https://github.com/infor-design/enterprise/files/14377095/expandablearea.demo.txt)

If you then run the project and view the expandable area demo with a small screen size you should see the following (note the lack of margins around the fields)

expandable_area_issue

If you check the CSS rules applied in chrome dev tools, you will note the @media style for the left margin of the column is overridden by the .expandable-area rule.  This should note be the case and @media rule should be applied instead.
expandable_area_rule

**Expected behavior**
The columns are displayed with the left-margin correctly applied when viewed in a small window/device.

**Resolution**
The issue is occurring because the @media rules are appearing early in the compiled theme-new-light.css file.  This means any rules that appear afterwards (targeted at desktop) can potentially override these rules.  The @media rules should appear last in the compiled CSS file to ensure this situation cannot occur.  In that respect this issue is potentially more wide ranging than the issue described above.

**Version**
Tested on Enterprise-ng 17.4.3

yohannahbautista:

Looking into it, this was more an issue on responsive width. If you put a similar setup on a page without the module nav (like in IDS stackblitz) or without the module nav expanded by default (like in IDS NG stackblitz), this bug doesn't appear as the expected behavior is that the width is automatically adjusted to avoid that squished appearance. 

In so far, the bug only appeared when a module nav is present and when its expanded. If this isn't the case, comments will be appreciated and preferably a more detailed example page. 

Closes #8485 

- Go to: http://localhost:4000/components/module-nav/test-six-columns.html
- Check the input fields in different screen widths and when the module nav is expanded or not
- Should not squish together

### DIFF
--- a/app/views/components/module-nav/test-six-columns.html
+++ b/app/views/components/module-nav/test-six-columns.html
@@ -1,0 +1,471 @@
+<style>
+  .test-spacer {
+    height: 2000px;
+  }
+</style>
+
+<div class="svg hidden">
+    <div class="svg-icons">
+      <svg xmlns="http://www.w3.org/2000/svg" version="1.1" class="svg-icons">
+        <symbol id="icon-app-ac" viewBox="0 0 32 32">
+          <rect width="32" height="32" rx="8" fill="#4DCC86" />
+          <path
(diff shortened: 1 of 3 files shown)


n-ace-ancog (dismissed):
Working as expected:

https://github.com/infor-design/enterprise/assets/156752175/1df0e923-a18b-465d-93d2-b1e8e4bbbc27

https://github.com/infor-design/enterprise/assets/156752175/8ee8bdfb-be02-44b1-a347-195880071047

tmcconechy (approved):

tmcconechy (approved):

PR #8823, opened 2024-06-14
NG1697 - Fixed example page for disabling dismissible tags

yohannahbautista:

Updated example page. The issue was the check, it would have never returned false. The element passed on the beforeclose event is already the hyperlink and not the list item, so finding the child hyperlink leads to nothing (and therefore leads to undefined which leads to the if statement on the example page never being executed).

> Add ability to disable the x button
> Check the dismissible attribute is working (this one hides the x)

I'm also not sure if disabling /hidingthe close icon when a tag has the "dismissible" class is what it should be for because the "Dismissible" tag has the same class in the example
https://github.com/infor-design/enterprise/blob/main/app/views/components/tabs/test-prevent-dismissible.html#L16
The only tag that has no close icon is the one without the dismissible class. Could toggling the presence of the dismissible class and updating tags be a solution if the close icon existing is an issue?

Closes https://github.com/infor-design/enterprise-ng/issues/1697

- Go to: http://localhost:4000/components/tabs/test-prevent-dismissible.html
- Dismissible tag should close when closing
- Prevent Dismissible tag should not close

### DIFF
--- a/app/views/components/tabs/test-prevent-dismissible.html
+++ b/app/views/components/tabs/test-prevent-dismissible.html
@@ -36,17 +36,16 @@
 
 <script>
   $('body').on('initialized', function() {
-    var tabs = $('#dismissible-tabs');
+    const tabs = $('#dismissible-tabs');
 
-    function announceEvent(e, li) {
-      var type = e.type,
-        a = li.children('a'),
-        href = a.attr('href');
+    function announceEvent(e, a) {
+      const type = e.type;
+      const href = a.attr('href');
 
       $('body').toast({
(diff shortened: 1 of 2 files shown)


tmcconechy:
@yohannahbautista does this have/work in the angular example too?

tmcconechy (approved):

jbrcna (approved):
image

image

glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/751d6121-392c-41c0-b4d2-229e99278f12

n-ace-ancog (approved):
Working on my end.

image

ericangeles (approved):

PR #8871, opened 2024-07-04
8814 - Added searched event

### LINKED ISSUE #8814: DataGrid: add "searched" trigger event
we have this use case that empty message for the data grid changes depends on the following conditions:
- when the data grid has no records
  -  empty message is to instruct the user to create a new record
- when filtering/keyword searching has no records
  -  empty message is to instruct the user to refine the search 

Problem is DataGrid only fires event when filtering of rows coming from Column filters but not when keyword search is triggered.
image

to add a new trigger event "searched", same with "filtered", emit the filter expression
image

**Additional context**
And what application do you work on? Infor M3 CLM

In addition, if in case this gets implemented, implementation of the "searched" event emitter to enterprise-ng is much appreciated.

yohannahbautista:

Added searched event. Will add in NG once this is approved to be sure.

Closes #8814 

- Go to: http://localhost:4000/components/datagrid/example-custom-toolbar-flex.html
- Inspect console
- Type on toolbar search
- Console should print searched and the filtered rows

### DIFF
--- a/app/views/components/datagrid/example-custom-toolbar-flex.html
+++ b/app/views/components/datagrid/example-custom-toolbar-flex.html
@@ -84,8 +84,8 @@
                 idProperty:'officeId',
                 dataset: data,
                 selectable: 'single',
-                toolbar: {keywordFilter: true, results: true}
-            });
+                toolbar: { keywordFilter: true, results: true}
+            }).on('searched', (e, args) => console.log('searched', args));
 
             $('#export').on('click', function () {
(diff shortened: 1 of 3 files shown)


glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/47cf9c87-cd29-4925-8852-1c708325b132

jbrcna (approved):
image

tmcconechy (approved):

PR #8882, opened 2024-07-09
1689 - Allow checkboxes to be disabled for selection

tmcconechy:

Allows isEditable checkboxes to disable rows selections. Also pushed incorrect initial count in some examples

Fixes https://github.com/infor-design/enterprise-ng/issues/1689

- go to http://localhost:4000/components/datagrid/example-disabled-selection-checkbox.html
- clicking the disable checkboxes should not select a row
- clicking the last 2 will select a row
- repeat test with keyboard / space key should only select non editable rows
- note click to select is off so clicking anywhere but the checkbox column will not select the row

- [x] A test for the bug or feature.

### DIFF
--- a/app/views/components/datagrid/example-disabled-selection-checkbox.html
+++ b/app/views/components/datagrid/example-disabled-selection-checkbox.html
@@ -55,6 +55,36 @@
       }
       return isEditable;
     };
+    const checkDisabled = (row, cell, value, item, rowData) => {
+      let isDisabled = true;
+      if (rowData.lineType === 'Amount Only') {
+        if (item.field === 'orderAmount') {
+          isDisabled = false;
+        }
+      } else if (rowData.lineType === 'Rate Only') {
+        if (
+          item.field === 'procurementRate' ||
+          item.field === 'procurementRateUnit'
+        ) {
+          isDisabled = false;
+        }
+      } else if (rowData.lineType === 'Quantity & Rate') {
+        if (
+          item.field === 'procurementRate' ||
+          item.field === 'procurementRateUnit' ||
+          item.field === 'quantity' ||
+          item.field === 'quantityUnit'
+        ) {
+          isDisabled = false;
+        }
+      }
+      if (rowData.status === 'Modified') {
+        if (item.id === 'selectionCheckbox') {
+          isDisabled = false;
+        }
+      }
+      return isDisabled;
+    };
     const columnGroups = [
       {
         colspan: 1,
@@ -105,7 +135,8 @@
         resizable: false,
         formatter: Soho.Formatters.SelectionCheckbox,
         align: 'center',
-        isEditable: checkEditable,
+        // isEditable: checkEditable -> this also works
+        disabled: checkDisabled
       },
       {
         id: 'LineEdit',
--- a/app/views/components/datagrid/example-disabled-selection-checkbox.html
+++ b/app/views/components/datagrid/example-disabled-selection-checkbox.html
@@ -1,0 +1,310 @@
+
+<div class="row">
+  <div class="twelve columns">
+    <div class="contextual-toolbar toolbar is-hidden">
+      <div class="title selection-count">0 Selected</div>
+      <div class="buttonset">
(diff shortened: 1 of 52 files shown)


app/views/components/datagrid/example-disabled-selection-checkbox.html
```diff
@@ -0,0 +1,310 @@
...
+    </div>
+    <div id="datagrid" class="datagrid">
+    </div>
+  </div>
+</div>
+
+<script>
+  $('body').one('initialized', function () {
+    const checkEditable = (row, cell, value, item, rowData) => {
+      let isEditable = false;
```
sami1riaz, line 29: Can we not use the disabled attribute on the column to disable the checkbox? I believe we can use that attribute on the  Soho.Formatters.Button and Soho.Formatters.Checkbox to disable them. But not with Soho.Formatters.SelectionCheckbox

tmcconechy: Do you mean instead of isEditable? https://github.com/infor-design/enterprise/pull/8882/files#diff-871e1d2fe997e8998bc13fc9cab16588e521845c6f79bf10073567f1836e6139R108

Use disabled function as well? I think we can do that.

tmcconechy: Update: I do not see any disabled setting on the column. Only `isReadonly` or `isEditable`

You cant set it right on the checkbox as its dynamic from the column data.

glenlieorillo (changes requested):
Using the spacebar to click the checkbox in the disabled row is still possible.

https://github.com/infor-design/enterprise/assets/102151352/9d46e6d4-fb60-4603-b2c5-bdbfbec5c4f1

jbrcna (changes requested):
image

app/views/components/datagrid/example-disabled-selection-checkbox.html
sami1riaz: Ok, i was under the impression that the disabled setting could be set for selectionCheckboxes as well, the way we do it for buttons.

tmcconechy: Do you have an example of that? Just to make sure im not missing something else?

tmcconechy:
@sami1riaz OK i see what you mean now. Was missing examples for this so i didnt see it.

I updated the selection checkbox to work on `disabled` as well. Note that you will have to reverse your logic (false vs true) and then i used the checkbox style but didnt make the column readonly. I think this might be what you want.

Please try and confirm

sami1riaz:
@tmcconechy, thanks! Yes this is what we need. Looks good.

tmcconechy:
ok @sami1riaz we will patch once QA looks and the tests pass 👍🏻

janahintal (approved):
Now working as expected on my end. Thanks!
Test Data: [screen-capture (1).webm](https://github.com/infor-design/enterprise/assets/45757016/bbb01a4d-7855-41c5-bbd6-9e47a2e0b093)

glenlieorillo (approved):
https://github.com/infor-design/enterprise/assets/102151352/1b91611c-cf5d-4b01-ae27-bf44c6e32113

jbrcna (approved):
great work!
image

sami1riaz:
Hi @tmcconechy, @jbrcna, I just tested the fix and it is not working with the selectAll checkbox or with selecting the parent row in a tree grid:

image

You can see in the screenshot that the disabled row is also being selected. Can this be fixed as well please?

jbrcna:
@sami1riaz what link did you use for testing? 
I tried it on main and it's working. https://main-enterprise.demo.design.infor.com/components/datagrid/example-disabled-selection-checkbox.html

tmcconechy:
@jbrcna we just missed something in the context of the disabled rows. I will fix and flag on a new pull request

jbrcna:
thanks for clarifying @tmcconechy

Issue #8850, opened 2024-06-26
Datagrid: Incorrect updates in commitCellEditUtil


crisjoshua-inf:
**Describe the bug**
When using setActiveCell and click to refocus from one editable cell to another. A unique bug occurs where the value of the initially entered/clicked value is placed in the next focused value. This is only achieved with a specific scenario of code

Things to note: 
datagrid.settings.source always has a value with an placeholder function much like this
() => {
/* placeholder empty */
};.
when enter is done, a different focus is simulated and setActiveCell or element.click is used by the application to focus on the editable cell

**To Reproduce**
1. Use attached html file and replace the existing test-editable-inline-editor.html (https://github.com/user-attachments/files/15966129/test-editable-inline-editor.zip)
2. Press enter on 3 editable field or as show on gif below
![SimulatedEnterKeyIssue](https://github.com/infor-design/enterprise/assets/115533426/e78d0cd5-fd11-4100-8ce6-121869c1fe1a)

**Expected behavior**
Value of the 3rd editable row does not duplicate in the first editable row

**Version**
v4.97.0

**Platform**
 - Infor M3 H5A Application
 - Browser Name: Chrome/Edge
 - Browser Version: Version 126.0.6478.115 (Official Build) (64-bit) / Version 126.0.2592.68 (Official build) (64-bit)

**Additional context**

tmcconechy:
This one is https://github.com/infor-design/enterprise/pull/8838 correct?

crisjoshua-inf:
Yes this is the one.

PR #8838, opened 2024-06-20
Fix incorrect updates in commitCellEditUtil 

crisjoshua-inf:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**

On commitCellEditUtil function:
Added handling when cellNode and this.activeCell mismatches rowIndex and dataRowIndex to avoid incorrect updates to editable cells. Added Difference check for cellNode and active cell index

**Related github/jira issue (required)**:
https://inforwiki.atlassian.net/browse/MUA-14253

Fixes #8850

**Steps necessary to review your pull request (required)**:

test scenarios:
In M3 Angular application:

1. Open MMS300 Application -> Go to Data with valid editable cell. 
2. Edit First Editable cell on first row then edit second row as well on same column.
3. On Second Row, Editable cell, press enter twice very fast.

Actual: Value in second row is being placed in first row editable cell.
DoubleEnterKeyIssue

Things to note in M3 Angular Application:
1. this.settings.source always has a value with an placeholder function much like this 
() => {
          /* placeholder empty */
};.
2. when enter is done, a different focus is received from BE and setActiveCell or element.click is used by the application to focus on the editable cell
3. Will attempt to provide a stackblitz or an example in enterprise demo code in the future for better referrence. Thanks!

**Included in this Pull Request**:
- [ ] An e2e or functional test for the bug or feature.
- [ ] A note to the change log.

PR #8833, opened 2024-06-19
8832 - Fix memory leak in datagrid

henryb0119:
**Explain the _details_ for making this change. What existing problem does the pull request solve?**
On updateRow function:
Existing DOM elements were not cleaned up before replacing with new ones.
Fixes:
- Detach eventListeners.
- Remove all attributes.
- Remove all attached data.

setTimeout functions have closure errors.
- Fix: Now passing the external variables as function parameter.

eventListener functions have closure errors.
- set the parent function of eventListener's function to null during destroy event of the datagrid.

calculateTextWidth is calculating the entire list cells.
- Fix: Just calculate what is visible in the screen.

**Related github/jira issue (required)**:

Closes https://github.com/infor-design/enterprise/issues/8832

**Steps necessary to review your pull request (required)**:
- test scenarios:
In M3 Angular application:
1. Open 2 programs then scroll down 10x
2. Scroll 10x more 
3. Check chrome task manager
    Expected: No memory leak
    Actual: Memory heap usage has reached 3gb
memleak
memleakTM
https://github.com/infor-design/enterprise/assets/20462137/ec159218-9182-4b1c-8694-652f2777daea
original recording: file://PHMANHBANZON01/Users/hbanzon/OneDrive%20-%20Infor/For%20Share/memleak.webm

**Included in this Pull Request**:
- [x] An e2e or functional test for the bug or feature.
- [ ] A note to the change log.

PR #8899, opened 2024-07-15
Chore: Fix incorrect updates in commitCellEditUtil - Additional changes

crisjoshua-inf:
Explain the details for making this change. What existing problem does the pull request solve?

On commitCellEditUtil function:
https://github.com/infor-design/enterprise/pull/8838 combined with https://github.com/infor-design/enterprise/pull/8833 has issues where we encountered errors in this.actualRowIndex(cellNode.parent()) being NaN. This is a proposed PR to mitigate having console log errors.

Related github/jira issue (required):
https://inforwiki.atlassian.net/browse/MUA-14253

Fixes https://github.com/infor-design/enterprise/issues/8850

Steps necessary to review your pull request (required):
test scenarios:
In M3 Angular application:

Original Issue:
Open MMS300 Application -> Go to Data with valid editable cell.
Edit First Editable cell on first row then edit second row as well on same column.
On Second Row, Editable cell, press enter twice very fast.
Actual: Value in second row is being placed in first row editable cell.
DoubleEnterKeyIssue

Issue caused by combination:
image

Things to note in M3 Angular Application:
this.settings.source always has a value with an placeholder function much like this
() => {
/* placeholder empty */
};.
when enter is done, a different focus is received from BE and setActiveCell or element.click is used by the application to focus on the editable cell.

Replication steps is still in: 
http://localhost:4000/components/datagrid/test-editable-inline-editor-quick-focus-change.html.

Currently using v40.9.5.

### DIFF
--- a/src/components/datagrid/datagrid.js
+++ b/src/components/datagrid/datagrid.js
@@ -10925,21 +10925,18 @@ Datagrid.prototype = {
       newValue = xssUtils.escapeHTML(newValue);
     }
 
-    let rowIndex;
+    let rowIndex = this.actualRowIndex(cellNode.parent());
     let dataRowIndex;
     if (this.settings.source !== null && isUseActiveRow) {
-      if (cellNode && this.actualRowIndex(cellNode.parent()) !== this.activeCell.rowIndex) {
-        rowIndex = this.actualRowIndex(cellNode.parent());
+      if (!isNaN(rowIndex) && rowIndex !== this.activeCell.rowIndex) {
         dataRowIndex = this.dataRowIndex(cellNode.parent());
       } else {
         rowIndex = this.activeCell.rowIndex;
         dataRowIndex = this.activeCell.dataRow;
       }
     } else {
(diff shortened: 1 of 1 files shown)


src/components/datagrid/datagrid.js
```diff
@@ -10925,21 +10925,18 @@ Datagrid.prototype = {
       newValue = xssUtils.escapeHTML(newValue);
     }
 
-    let rowIndex;
+    let rowIndex = this.actualRowIndex(cellNode.parent());
```
jdilag90, line 10928: maybe add the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to `cellNode` before calling `.parent()`?
Previously, there was a truthy for `cellNode` in the `if` statement.

```
let rowIndex = this.actualRowIndex(cellNode?.parent());
```

crisjoshua-inf: Hi, Added the optional operator. Thanks

tmcconechy (approved):

PR #8911, opened 2024-07-23
8744 - Refresh pager when applyFilter is called

### LINKED ISSUE #8744: Datagrid: Pager not updating after new data is pushed and filter is applied.
**Describe the bug**

We've got a grid that we push new data to via a timer. On our side we update the data, then we reapply the filter to this grid. However, within your control, you're not correctly changing the filterChanged flag. If I override filterChanged it works fine.

Also the text is not aligned between the input and page selector.

**To Reproduce**
This issue is difficult to reproduce, as you don't have any example pages with this set-up. The screenshots detail this issue well and you can see this issue. It would appear you're looking for a change in the filter when you apply it rather than just applying it as is.

For the alignment issue see below.

**Expected behaviour**
Any time a filter is applied I expect it to still be applied when I add/refresh/remove data from the grid.

**Version**
4.93.0 - I checked out the version and reviewed the change logs this issue has not occurred before.

**Screenshots**

Please see the highlighted area, this showcases the error.
image

The error is caused by filterchange not being set correctly. 
image

**Platform**
 - Infor Application/Team Name: Infor SunSystems
 - Device: (if applicable) N/A
 - OS Version: Window 10
 - Browser Name: Chrome & Edge
 - Browser Version: 124.0.6367.210 (Official Build) (64-bit) & Version 125.0.2535.51 (Official build) (64-bit)

**Additional context**

yohannahbautista:

Refresh pager when applyFilter is called and if pager API exists

Closes #8744 

- Go to: http://localhost:4000/components/datagrid/test-pager-addrow.html
- Click Add Row
- Pager should update

### DIFF
--- a/app/views/components/datagrid/test-pager-addrow.html
+++ b/app/views/components/datagrid/test-pager-addrow.html
@@ -1,0 +1,279 @@
+<div class="row">
+  <div class="twelve columns">
+    <div role="toolbar" class="toolbar">
+      <div class="title">Datagrid #8744</div>
+
+      <div class="buttonset">
+        <button type="button" id="add-row-button" class="btn">
+          <span>Add Row</span>
+        </button>
+        <button type="button" id="applyFilter" class="btn">
+          <span>Reapply Filter</span>
+        </button>
+      </div>
+    </div>
+
+    <div id="datagrid"></div>
+  </div>
+</div>
+
+<script>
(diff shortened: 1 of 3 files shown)


jbrcna (approved):
image

tmcconechy (approved):

Issue #8870, opened 2024-07-04
Datagrid: Add selectOnEdit setting for inline editor


rcoseng:
**Describe the bug**
On 1st click in the inline editor, the text is highlighted (expected)
On succeeding clicks in between characters, the cursor moves and then is highlighted.
When trying to drag to select only a set of characters, the whole text is highlighted on release of mouse.

**To Reproduce**

1. Go to https://latest-enterprise.demo.design.infor.com/components/datagrid/test-editable-with-inline-editor.html
2. Click on inline editor "Repair"
3. Click in between "R" and "e" (cursor should be in between)
4. Click in between "a" and "i"
5. See text highlight behavior
6. Click in between "R" and "e" (cursor should be in between)
7. Drag to select "epai" only
8. Notice the whole text is selected

**Expected behavior**
On clicks after the first, the cursor is moved in between characters. 
When dragging to select a set of characters, the whole text is not highlighted on release.

**Version**

- ids-enterprise: 4.95

**Screenshots**

https://github.com/infor-design/enterprise/assets/120151603/f7b1c4e2-8d5c-43d7-9929-e67ac1204577

**Additional context**
see linked comment for `selectOnEdit` suggestion:
https://github.com/infor-design/enterprise/issues/8763#issuecomment-2195120378

tmcconechy:
Seems to still be a bug on https://main-enterprise.demo.design.infor.com/components/datagrid/test-editable-with-inline-editor.html

PR #8919, opened 2024-07-25
8870 - Datagrid Inline Editor SelectOnEdit

tjamesallen15:

This pull request will fix selecting issue on inline editor by adding `selectOnEdit` setting.

Closes https://github.com/infor-design/enterprise/issues/8870

- Go to http://localhost:4000/components/datagrid/test-editable-with-inline-editor.html
- Click the value inside the inline editor
- It should not be select on click

### DIFF
--- a/app/views/components/datagrid/test-editable-with-inline-editor.html
+++ b/app/views/components/datagrid/test-editable-with-inline-editor.html
@@ -120,6 +120,7 @@
       paging: true,
       pagesize: 5,
       pagesizes: [2, 5, 6],
+      selectOnEdit: false,
       trimSpaces: true,
       columnReorder: true,
       showDirty: true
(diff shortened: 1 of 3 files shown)


jbrcna (dismissed):
image

n-ace-ancog (dismissed):
Working on my end.

https://github.com/user-attachments/assets/4fbb0f25-a444-4fa4-a689-d8f3cf2b4d3f

tmcconechy:
@rcoseng can you test this?

I was thinking that selectOnEdit would focus all the text when you click in. Was this the idea? I think the behavior is correct now but maybe we need this to be the default a selectOnEdit selects all the text.

ericangeles (approved):
Looks good. I think this is a good way to provide options.

Can you add this to NG as well, @tjamesallen15 ?

jbrcna (approved):

tmcconechy (approved):

rcoseng:
> @rcoseng can you test this?
> 
> I was thinking that selectOnEdit would focus all the text when you click in. Was this the idea? I think the behavior is correct now but maybe we need this to be the default a selectOnEdit selects all the text.

@tmcconechy Hi, yes that is the idea. selectOnEdit = true would focus all the text. I have also tested it and it looks good. Thanks!

PR #8971, opened 2024-08-15
8646 - Fixed auto sizing in tree grid

### LINKED ISSUE #8646: Datagrid: Autosizing not working correctly with tree grid
**Describe the bug**
When using column auto-sizing in tree grid there are two issues occurring:
 1. Children content is not taken into consideration when column width is calculated
 2. Audible description of expand/collapse button is used as part of text when calculating the width, making the column wider than it needs to be.

**To Reproduce**

1. Go to 'https://stackblitz.com/edit/ids-enterprise-4940-qethhi?file=index.js'
2. Notice in first datagrid that the second row's (first child of first row) content is clipped.
3. Notice in second datagrid that the first column is a lot wider than it needs to be.

**Expected behavior**
1. Children of tree grid should be taken into consideration when calculating column width.
2. Hidden text of expand/collapse button should not be taken into account when calculating column width.

**Version**

- ids-enterprise: v4.94

**Platform**
 - Infor Application/Team Name: Infor WFM

yohannahbautista:

Included calculating width for children in tree grid

Closes #8646 

- Go to: http://localhost:4000/components/datagrid/test-treegrid-colwidth.html
- Content shouldn't be cut off in first column of first datagrid
- Width shouldn't be very long in the first column of second datagrid

### DIFF
--- a/app/data/tree-tasks.js
+++ b/app/data/tree-tasks.js
@@ -41,6 +41,9 @@ export default function treeTasks(req, res) {
       portable: true,
       children: [
         {
+          id: 16, escalated: 0, depth: 2, taskName: 'Follow up action with HMM USA. Deserunt magnam iste, nor consectetur. In. Explicabo. Veniam commodi, and quae, for officia.', desc: 'Contact sales representative with the updated purchase order.', comments: 'Example Comment', orderDate: new Date(2014, 6, 9), time: '20:10 PM', portable: true
+        },
+        {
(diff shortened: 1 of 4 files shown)


tmcconechy (dismissed):
Looks good @yohannahbautista do you think..

a) we can add some margin/padding between the grid examples?
b) and an example with more combinations of levels (3/4 more) (can reuse other example data or change other examples to defaulting width)?

tmcconechy (dismissed):

jbrcna (approved):
image

glenlieorillo (approved):
image

ericangeles (approved):

PR #9002, opened 2024-09-04
8779 - Added promise option for beforeclose event

### LINKED ISSUE #8779: Support asynchronous operation in modal.beforeClose() callback
The Modal Component's beforeClose() callback only accepts a callback method that returns boolean. That means, no async operations can be performed within the callback. This prevents user from being able to perform async validation inside the .beforeClose() callback, when deciding whether the modal should be closed or not. For example, if we want to submit a form from a modal, and the API call returns error, we keep the modal opened and display an error message.

The Modal component would work better if the .beforeClose() can also accept a callback that returns a Promise<boolean>, so that when .close() is called on the modal, the beforeClose() callback can be executed asynchronously.

Instead of using .beforeClose() to decide whether a modal should be closed, we hardcode the validation process into the onClick method in the button that closes the modal, and only invokes modal.close() if validation passes.

**Additional context**
And what application do you work on?
Infor Workforce Management(WFM)

yohannahbautista:

Added promise option for beforeclose event

Will add to NG on beforeClose once this is approved

Closes #8779 

- Go to: http://localhost:4000/components/modal/test-async.html
- Open modal
- Close modal 
- Should close after 1 second

### DIFF
--- a/app/views/components/modal/test-async.html
+++ b/app/views/components/modal/test-async.html
@@ -37,7 +37,7 @@
 
   const onBeforeShow = (api, response) => {
     timer = setTimeout(function() {
-      response(); 
+      response();
     }, 2000);
   };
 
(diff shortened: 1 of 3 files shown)


glenlieorillo (approved):
https://github.com/user-attachments/assets/f63459e9-863d-40a6-ae2b-6e48b14acea2

**RTL**

https://github.com/user-attachments/assets/3fa88806-552d-47c4-8d41-ada6a039ab02

ericangeles (approved):

tmcconechy (approved):

tmcconechy (approved):

